PR #10, opened 2018-10-14
improved titles

conradsnicta:
set of "improved" titles

### DIFF
--- a/doc/mlsys_paper/paper.tex
+++ b/doc/mlsys_paper/paper.tex
@@ -36,7 +36,16 @@
 \begin{document}
 
 % TODO: title could possibly be improved
-\title{\texttt{ensmallen}: a generic C++ library for fast optimization}
+%\title{\texttt{ensmallen}: a generic C++ library for fast optimization}  %% it's not clear what "optimization" refers to here
+%% other possibilities:
+\title{\texttt{ensmallen}: a flexible C++ library for function optimization in machine learning}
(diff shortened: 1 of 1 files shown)


doc/mlsys_paper/paper.tex
```diff
@@ -36,7 +36,16 @@
 \begin{document}
 
 % TODO: title could possibly be improved
-\title{\texttt{ensmallen}: a generic C++ library for fast optimization}
+%\title{\texttt{ensmallen}: a generic C++ library for fast optimization}  %% it's not clear what "optimization" refers to here
+%% other possibilities:
+\title{\texttt{ensmallen}: a flexible C++ library for function optimization in machine learning}
```
rcurtin: Hmm, what do you think of "ensmallen: an easy and flexible C++ library for function optimization"?

(I think we can drop the 'machine learning' since ensmallen isn't specific to that.)

zoq: Agreed, we shouldn't create the impression this is only useful for machine learning, even if this is a really broad term.

conradsnicta: How about a simplified version: "ensmallen: a flexible C++ library for function optimization"
I'm not keen on having "easy" in the title as that has connotations with "simple" and "unsophisticated". I'd prefer to avoid potential perception traps like that, given that we're dealing with a cutthroat reviewing environment.

rcurtin: Sure, that works for me.

PR #11, opened 2018-10-14
first stab at the abstract

conradsnicta:
starting point for the abstract. needs sharpening and more meat (eg. need to explicitly state the speedups from the experiments).


rcurtin:
Definitely better than what I'd written :)

PR #12, opened 2018-10-14
sharpening and defluffing the introduction

conradsnicta:


rcurtin:
Definitely looks better.  Agreed on all points.

PR #13, opened 2018-10-15
title candidate

conradsnicta:
proposed title: "ensmallen: a flexible C++ library for efficient function optimization"
I've added "efficient" for two reasons:
(a) suggests that it's "better"/"faster" than other approaches
(b) the words "function optmization" are now together on a separate line within the title (much cleaner look)

### DIFF
--- a/doc/mlsys_paper/paper.tex
+++ b/doc/mlsys_paper/paper.tex
@@ -38,7 +38,17 @@
 \begin{document}
 
 % TODO: title could possibly be improved
-\title{\texttt{ensmallen}: a generic C++ library for fast optimization}
+%\title{\texttt{ensmallen}: a generic C++ library for fast optimization}  %% it's not clear what "optimization" refers to here
+%% other possibilities:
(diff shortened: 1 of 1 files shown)


rcurtin:
I like it, I was just thinking that maybe we should fit 'efficient' or 'fast' in there somehow.

rcurtin:
That's true, but at the same time we have the code that people can just simply run in the repository, and the speedups are really quite good.  I've certainly done my fair share of opening this Pandora's box, but at the same time, when we submitted the framework as part of mlpack last year, the reviewers actually rejected it mostly on the grounds that there were no performance comparisons of any sort and therefore we couldn't validate if the library was even useful.

conradsnicta:
Having the performance comparisons is certainly useful.  Perhaps packaging them as "ensmallen is more efficient than xyz", which sounds a little less hazardous than "ensmallen is faster than xyz".  The word "efficient" can encompass more than just speed, so it's constructively ambiguous (ie. deliberate use of ambiguous language on a potentially sensitive issue).

PR #16, opened 2018-10-15
fix the default super-ugly coloured link boxes (citations, urls, figure numbers, footnote numbers, etc)

conradsnicta:
Remove those super-ugly link boxes and replace them with sane looking links.  This noticeably improves the presentation of the paper (ie. less annoying for reviewers).


rcurtin:
Yeah... they are super ugly, definitely.  :+1:

PR #17, opened 2018-10-15
section 2: clarifications, sharpening and defluffing

conradsnicta:
Clarifications, sharpening and defluffing for section 2 (Types of objective functions).
I've also added notes on where further edits are probably required.

### DIFF
--- a/doc/mlsys_paper/paper.tex
+++ b/doc/mlsys_paper/paper.tex
@@ -173,7 +173,7 @@ It is available for download from \url{https://www.ensmallen.org}.
 Armadillo~\cite{sanderson2016armadillo} is used for efficient linear algebra operations.
 
 \vspace*{-0.2em}
-\section{Types of objective functions}
+\section{Types of Objective Functions}
 \vspace*{-0.6em}
 
 \begin{table}[b!]
@@ -261,7 +261,7 @@ that the framework can natively take advantage of sparse gradients; and
 \end{table}
 
 {\tt ensmallen} provides a {\bf set of optimizers} for
-{\bf optimizing user-defined objective functions}.  It's also easy to implement a
+optimizing {\bf user-defined objective functions}.  It is also easy to implement a
 new optimizer in the {\tt ensmallen} framework.  Overall, our goal is to provide
 an easy-to-use library that can solve the problem
 %\vspace*{-0.4em}
@@ -270,18 +270,22 @@ $\operatorname{argmin}_{x} f(x)$
 %\end{equation}
 %\vspace*{-0.4em}
 %\noindent
-for any function $f(x)$ that takes some vector or matrix input $x$.  Ideally,
-the user should only need to specify what $f(x)$ is (or provide an
-implementation of a function that can compute $f(x)$).  But it is impossible to
-implement something both so generic and fast---for instance, gradient-based
-approaches converge far more quickly than gradient-free approaches (in general),
-so we must design an abstraction that is able to simultaneously generalize to
+for any function $f(x)$ that takes a vector or matrix input $x$.  Ideally,
+the user should only need to specify what $f(x)$ is, or provide an
+implementation of a function that can compute $f(x)$.  However, it is very difficult to
+implement something that is both generic and fast---for instance, gradient-based
+approaches typically converge far quicker than gradient-free approaches.
+As such, we must design an abstraction that is able to simultaneously generalize to
 many problem types, as well as take advantage of accelerations and
-optimizations.
+code optimizations.
+%% CS: in the above sentence, it's not clear what is meant by "we must design ..."
+%% CS: should it be "we provide ..." or "ensmallen provides ..." ?
+%% CS: The overall intent of the sentence is not clear -- to my eyes
+%% CS: it doesn't seem to gel well with its surrounding context.
 
 Therefore, users may implement objective functions with specific properties that
 the library can use to accelerate the computation.  These non-exclusive
-properties are shown below:
+properties are listed below:
 
 \vspace*{-0.3em}
 \begin{itemize} \itemsep -1pt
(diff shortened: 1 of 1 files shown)


doc/mlsys_paper/paper.tex
```diff
@@ -270,18 +270,22 @@ \section{Types of objective functions}
...
+implement something that is both generic and fast---for instance, gradient-based
+approaches typically converge far quicker than gradient-free approaches.
+As such, we must design an abstraction that is able to simultaneously generalize to
 many problem types, as well as take advantage of accelerations and
-optimizations.
+code optimizations.
+%% CS: in the above sentence, it's not clear what is meant by "we must design ..."
+%% CS: should it be "we provide ..." or "ensmallen provides ..." ?
+%% CS: The overall intent of the sentence is not clear -- to my eyes
+%% CS: it doesn't seem to gel well with its surrounding context.
```
rcurtin: Right, I was thinking this also.  The idea was to present the ideal---that a user just writes $f(x)$---but then point out this is not realistic, and as a result we have this abstraction where users often need to write $f(x)$ and $f'(x)$.

Why don't we just ditch that and try something like this... instead of "Ideally... ...code optimizations" we could do this:

```
...for any function $f(x)$ that takes a vector or matrix input $x$.
In most cases, $f(x)$ will have special structure; one example might be that $f(x)$ is differentiable.  Therefore, the abstraction we have designed for {\tt ensmallen} takes advantage of this structure.  A user can also, for instance, provide an implementation of $f'(x)$, which in turn allows first-order gradient-based optimizers to be used.  This can allow significant speedups.
```

(Perhaps the last sentence or two could be improved.)

doc/mlsys_paper/paper.tex
```diff
@@ -301,48 +305,50 @@ \section{Types of objective functions}
 \vspace*{-0.3em}
 
 In this framework, if a user wants to optimize a differentiable objective
-function, they need only provide implementations of $f(x)$ and $f'(x)$, and
+function, they only need to provide implementations of $f(x)$ and $f'(x)$, and
```
rcurtin: I did that to reclaim a little space, but if it's not necessary "only need to" reads better.

doc/mlsys_paper/paper.tex
```diff
@@ -301,48 +305,50 @@ \section{Types of objective functions}
...
 work with as many optimizers as possible.
-To do this, {\tt ensmallen} makes heavy use of C++ template metaprogramming.
-When implementing an objective function to be
-optimized, a user can implement only a few methods and we can use
-metaprogramming to check that the given functions match the requirements of the
+
+To accomplish the flexibility, {\tt ensmallen} makes heavy use of
+C++ template metaprogramming.
+When implementing an objective function to be optimized,
+a user can implement only a few methods; metaprogramming is then automatically
```
rcurtin: Do you mind if I change "can implement" to "may implement" here?  This wording could be misread to imply that a user is only allowed to implement a few methods, which sounds restrictive.

doc/mlsys_paper/paper.tex
```diff
@@ -301,48 +305,50 @@ \section{Types of objective functions}
...
 \end{itemize}
 \vspace*{-0.3em}
 
-Or, alternately, they can simply implement a \texttt{EvaluateWithGradient()}
+Alternatively, the user can simply implement a \texttt{EvaluateWithGradient()}
 function that computes both $f(x)$ and $f'(x)$ simultaneously, which is useful
 in cases where both the objective and gradient depend on similar computations.
 
-The required API for separable differentiable objective functions (i.e. those
+The required API for separable differentiable objective functions (ie., those
```
rcurtin: Sorry for the pedantry :)  I think this should be "i.e.": https://english.stackexchange.com/questions/3911/correct-spelling-italicization-of-e-g-i-e

doc/mlsys_paper/paper.tex
conradsnicta: Looks good.  I recommend changing
"... takes advantage of this structure. A user can also, for instance, provide an
implementation of $f'(x)$"
to:
"... can optionally take advantage of this structure. For example, in addition to $f(x)$, the user can provide an implementation of $f'(x)$"

The "can optionally" is important, as we want to be absolutely clear that the user is not forced to define $f'(x)$ in all cases.

Another recommended change to emphasise positive aspects.  Change:
"This can allow significant speedups."
to"
"This can often lead to significant speedups."

doc/mlsys_paper/paper.tex
conradsnicta: "they need only" sounds weird - like something out of Shakespeare :)

doc/mlsys_paper/paper.tex
conradsnicta: ok with me.  how about:
"a user may only need to implement a few methods"
this sounds more positive in the sense that it implies that not a lot of work is needed

doc/mlsys_paper/paper.tex
conradsnicta: i'm fine with either.  the only issue is the extra spacing that latex inserts after a full stop -- this is why i used a comma here.
how about using the ~ character to prevent the extra space. for example, instead of "i.e. those", use "i.e.~those".

doc/mlsys_paper/paper.tex
```diff
@@ -301,48 +305,50 @@ \section{Types of objective functions}
...
 that would use an optimizer like SGD) is very similar, except that
-\texttt{Evaluate()} and \texttt{Gradient()} (or \texttt{EvaluateWithGradient()})
-should operate only on small mini-batches.  The same pattern applies for other
+\texttt{Evaluate()}, \texttt{Gradient()} and \texttt{EvaluateWithGradient()}
```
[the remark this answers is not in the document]
conradsnicta: The restrictions apply to all three functions (ie. they all operate on mini batches).  Having an "or" confuses things.  By having an "or" it could be read that either Evaluate() and Gradient() have the restrictions, or that EvaluateWithGradient() has the restrictions while the other two functions don't. Whether the user employs Evaluate() in conjunction with Gradient(), or only employs EvaluateWithGradient(), it's up to them.  We're simply listing the restrictions that apply to the functions.

PR #18, opened 2018-10-15
section 3: minor rewrites and suggestions for further improvements

conradsnicta:
Minor rewrites and suggestions for further improvements in section 3 (Example: Learning Linear Regression Models).

See the embedded notes/comments for more details.

### DIFF
--- a/doc/mlsys_paper/paper.tex
+++ b/doc/mlsys_paper/paper.tex
@@ -351,25 +351,27 @@ function itself must be implemented and then any optimizer may be used.
 \vspace*{-0.5em}
 
 As an example of usage, consider the linear regression objective
-function\footnote{Just for simplicity, we ignore the bias term.  It can be
-rederived by taking $x^*_i = (x_i, 1)$.}.  Given some dataset $X \in
+function\footnote{For simplicity, we ignore the bias term.  It can be
+rederived by taking $x^*_i = (x_i, 1)$.}.  Given a dataset $X \in
 \mathcal{R}^{n \times d}$ and associated responses $y \in \mathcal{R}^n$, the
 model of linear regression is to assume that $y_i = x_i \theta$ for each
-point and response $(x_i, y_i)$.  To fit this model $\theta \in \mathcal{R}^d$ best
+point and response $(x_i, y_i)$.  To fit this model $\theta \in \mathcal{R}^d$
 to the data, we must find
 
+%% CS: i've added \nolimits to save a bit of space
 \vspace*{-1.1em}
 \begin{equation}
-\operatorname{argmin}_\theta \sum_{i = 1}^n (y_i - x_i \theta)^2 =
+% f(\theta) =   %% CS: for clarity, perhaps we should have f(\theta) = ...
+\operatorname{argmin}_\theta \sum\nolimits_{i = 1}^n (y_i - x_i \theta)^2 =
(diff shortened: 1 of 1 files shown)


doc/mlsys_paper/paper.tex
```diff
@@ -398,24 +408,26 @@ \section{Example: Learning Linear Regression Models}
     };
 \end{minted}
 \vspace*{-0.4em}
+%% CS: the above code is currently shown across two pages - not a clean look.
+%% CS: how about we place it in a figure ?
```
rcurtin: A figure's fine with me.  I liked it inline if it was on one page since it keeps the linear flow of reading, but if it splits over a page, you're right, it's a little awkward.  (Also, the other code is inline and probably should stay that way.  Instead do you think we should aim to just get all the code for `LinearRegressionFunction` on one page and leave it inline?)

doc/mlsys_paper/paper.tex
```diff
@@ -378,18 +380,26 @@ \section{Example: Learning Linear Regression Models}
...
 
 % Indented a little to make it clear it's different from the text...
+
+%% CS: "accu((y - X * theta) % (y - X * theta))" is redoing computations
+%% CS: ie. efficiency problem
+%% CS: how about:
+%% CS:   arma::mat tmp = y - X * theta;
+%% CS:   return arma::accu(tmp % tmp);  // accumulate element-wise product
+%% CS: in the above code, the comment hints at what the % operator does,
+%% CS: as otherwise it's potentially confusing to someone not familiar with Armadillo
```
rcurtin: Agree, the only problem would be that it makes the code a bit longer.  How about:

```
return std::pow(arma::norm(y - X * theta), 2.0);
```

(I didn't like the extra usage of `std::pow()` there though originally.)

doc/mlsys_paper/paper.tex
conradsnicta: I'd strongly prefer to have large code listings (ie. more than, say, 4 or 5 lines) to be figures. It looks more professional and cleaner.  It's also makes it super clear what we're referring to by using "the code shown in Fig. X", rather than relying on the reader to accept that the "funny looking" stuff below is actually a continuation of the preceding text.

There's an effect known as [Banner Blindness](https://en.wikipedia.org/wiki/Banner_blindness), where people skip over stuff that seems out of place.  This is exactly what I'm experiencing when seeing this big code chunk -- my eyes want to skip over it and get back to reading the "real" text.  If I'm experiencing this, there's a good chance a reviewer may as well.

doc/mlsys_paper/paper.tex
conradsnicta: ok with me.

doc/mlsys_paper/paper.tex
rcurtin: I'll handle this one tomorrow morning unless you want to handle it while I'm asleep. :)

conradsnicta: I'll do a bunch of PRs, probably spaced over 10 hours or so.

rcurtin: Sounds good. I am headed to bed now, so I definitely won't be able to reapproach this within 10 hours. One PR is fine too, up to you.

conradsnicta: 1 large-ish PR done. sent email with more info and suggestions for improving the flow

PR #20, opened 2018-10-15
section 5: notes on why this section should be removed

conradsnicta:
This sections is very weak and not convincing; it basically states "look, we can do inlines", but provides no real evidence and/or citations where the supposed efficiency gains are demonstrated.

I suggest removing this section entirely, as it causes more harm than good (from a reviewer's point of view).

In practice, effect of C++ function inlining is only notably pronounced when dealing with data comprised of lots of small vectors.

For this section to be fleshed out, it would require reasoning based on citations covering compiler optimisation, effect of data size, experiments showing the effect of various vector sizes (eg. with and without inlining). There is no room in this paper to do this properly, and doing a poor job will reduce likelihood of acceptance.

### DIFF
--- a/doc/mlsys_paper/paper.tex
+++ b/doc/mlsys_paper/paper.tex
@@ -453,10 +453,27 @@ In addition, many optimization algorithms are currently in development and will
 be released in the future.
 
 \vspace*{-0.3em}
-\section{Templates for speed}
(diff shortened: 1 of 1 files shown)


rcurtin:
I removed the section so I'll close this one without merging.

PR #24, opened 2018-10-16
many improvements

conradsnicta:


rcurtin:
Looks great to me. :beers:

conradsnicta:
Another revision or two and we should be almost there.

PR #30, opened 2018-10-18
explicitly check for suitable armadillo version in case ensmallen is used without the cmake based installation

conradsnicta:


rcurtin:
:+1:

PR #34, opened 2018-10-19
bug fixes: use correct include header guard; use ENS_ prefix for macros

conradsnicta:
bug fixes:
* use correct include header guard
* use ENS_ prefix for macros to avoid potential clashes with mlpack and other software which may use same macro names
* use undef before define as good hygiene


rcurtin:
Thanks---I'd forgotten to correctly adapt this after porting it.

PR #35, opened 2018-10-19
bump to version 1.10; proposed name

conradsnicta:
"Corporate Catabolism"
https://en.wiktionary.org/wiki/catabolism


rcurtin:
I learned a new word. :)

PR #41, opened 2018-10-27
travis ci config

zoq:
Travis config to automatically run the test suite.

### DIFF
--- a/.travis.yaml
+++ b/.travis.yaml
@@ -1,0 +1,24 @@
+sudo: required
+dist: trusty
+language: cpp
+
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install -y --allow-unauthenticated libopenblas-dev liblapack-dev g++
+  - curl https://ftp.fau.de/macports/distfiles/armadillo/armadillo-6.500.5.tar.gz | tar xvz && cd armadillo*
+  - cmake . && make && sudo make install && cd ..
+
+install:
+  - mkdir build && cd build && cmake .. && make -j2
+
+script:
+  - CTEST_OUTPUT_ON_FAILURE=1 travis_wait 30 ctest -j2
+
+notifications:
+  email:
+    - mlpack-git@lists.mlpack.org
+  irc:
+    channels:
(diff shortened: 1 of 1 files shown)


.travis.yaml
```diff
@@ -0,0 +1,24 @@
+sudo: required
+dist: trusty
+language: cpp
+
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install -y --allow-unauthenticated libopenblas-dev liblapack-dev g++
+  - curl https://ftp.fau.de/macports/distfiles/armadillo/armadillo-6.500.5.tar.gz | tar xvz && cd armadillo*
```
conradsnicta: I recommend using the latest available version of Armadillo (currently 9.100.6).
http://sourceforge.net/projects/arma/files/armadillo-9.100.6.tar.xz
Version 6.500 is truly ancient at this point.  Later versions are considerably faster and contain many bug fixes.

@zoq @rcurtin

zoq: I see your point, the only reason for the downgrade is that this is the minimal supported version at least at this point; and I think it's easier to catch issues (like new functions) if we test against the minimal version.

rcurtin: We could run a matrix build, one with the newest and one with the oldest, if we wanted to.  The build time should not be anywhere near as long as mlpack on Travis.  Looks like that MacPorts site keeps up-to-date versions available.  Here's some ugly shell script to get the most recent version available there under some assumptions:

```
curl https://ftp.fau.de/macports/distfiles/armadillo/ -- | grep '.tar.xz' | sed 's/^.*<a href="\(armadillo-[0-9]*.[0-9]*.[0-9]*.tar.xz\)".*$/\1/' | tail -1
```

But in any case I do think it's important to use the minimum supported Armadillo version here, since one of the main uses of Travis is for pull requests; with the oldest supported version we can quickly know if a contributor's written code that needs a newer version.

zoq: The build should run in parallel anyway, so timing wise I don't see an issue here. The current version runs against the latest version and 6.500.5.

PR #42, opened 2018-10-28
SWATS optimizer

zoq:
Implementation of "Improving Generalization Performance by Switching from Adam to SGD" by Nitish Shirish Keskar and Richard Socher.

### DIFF
--- a/include/ensmallen_bits/swats/swats.hpp
+++ b/include/ensmallen_bits/swats/swats.hpp
@@ -73,7 +73,8 @@ class SWATS
    * @param beta1 Exponential decay rate for the first moment estimates.
    * @param beta2 Exponential decay rate for the weighted infinity norm
             estimates.
-   * @param eps Value used to initialise the mean squared gradient parameter.
+   * @param epsilon Value used to initialise the mean squared gradient
+   *        parameter.
    * @param maxIterations Maximum number of iterations allowed (0 means no
    *        limit).
    * @param tolerance Maximum absolute tolerance to terminate algorithm.
@@ -84,7 +85,7 @@ class SWATS
         const size_t batchSize = 32,
         const double beta1 = 0.9,
         const double beta2 = 0.999,
-        const double eps = 1e-8,
+        const double epsilon = 1e-8,
         const size_t maxIterations = 100000,
         const double tolerance = 1e-5,
         const bool shuffle = true);
--- a/include/ensmallen_bits/swats/swats.hpp
+++ b/include/ensmallen_bits/swats/swats.hpp
@@ -1,0 +1,158 @@
+/**
+ * @file swats.hpp
+ * @author Marcus Edel
+ *
+ * SWATS is a simple strategy which switches from Adam to SGD when a triggering
+ * condition is satisfied.
+ *
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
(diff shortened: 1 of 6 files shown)


include/ensmallen_bits/swats/swats_update.hpp
```diff
@@ -0,0 +1,176 @@
...
+  SWATSUpdate(const double epsilon = 1e-8,
+              const double beta1 = 0.9,
+              const double beta2 = 0.999) :
+    epsilon(epsilon),
+    beta1(beta1),
+    beta2(beta2),
+    iteration(0),
+    phaseSGD(false),
+    sgdRate(0),
+    sgdLamda(0)
```
rcurtin: I think this should be `sgdLambda`.

include/ensmallen_bits/swats/swats_update.hpp
```diff
@@ -0,0 +1,176 @@
...
+   *
+   * @param rows Number of rows in the gradient matrix.
+   * @param cols Number of columns in the gradient matrix.
+   */
+  void Initialize(const size_t rows, const size_t cols)
+  {
+    m = arma::zeros<arma::mat>(rows, cols);
+    v = arma::zeros<arma::mat>(rows, cols);
+
+    sgdV = arma::zeros<arma::mat>(rows, cols);
```
rcurtin: Since `sgdV` is only used during the SGD portion of optimization, it seems like it could be possible to reuse `v` here.  I don't have a strong opinion either way though.

tests/swats_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+TEST_CASE("SWATSSGDFunction","[SWATSTest]")
+{
+  SGDTestFunction f;
+  SWATS optimizer(1e-3, 1, 0.9, 0.999, 1e-6, 400000, 1e-9, true);
+
+  arma::mat coordinates = f.GetInitialPoint();
+  optimizer.Optimize(f, coordinates);
+
+  REQUIRE(coordinates[0] == Approx(0.0).margin(0.1));
+  REQUIRE(coordinates[1] == Approx(0.0).margin(0.1));
```
rcurtin: Looks like this failed... maybe the test needs to be run multiple times and have at least one success?

zoq:
> I have to wonder, would it be possible with an improved optimizer like this to reduce the step size earlier (i.e., perhaps at epoch 100 instead of epoch 150) and still achieve the same generalization results? I suppose an experiment like that might be a little hard to put in a paper, but still, to me, that would be the most interesting outcome of this line of work.


Agreed, the basic idea is really simple and I think it would be quite interesting to test some of the other optimizers and see if the parameters could be reduced.

zoq:
This should be ready as well.

include/ensmallen_bits/swats/swats.hpp
```diff
@@ -0,0 +1,159 @@
...
+ *                 arma::mat& gradient,
+ *                 const size_t batchSize);
+ *
+ * NumFunctions() should return the number of functions (\f$n\f$), and in the
+ * other two functions, the parameter i refers to which individual function (or
+ * gradient) is being evaluated.  So, for the case of a data-dependent function,
+ * such as NCA, NumFunctions() should return the number of points in the
+ * dataset, and Evaluate(coordinates, 0) will evaluate the objective function on
+ * the first point in the dataset (presumably, the dataset is held internally in
+ * the DecomposableFunctionType).
```
rcurtin: We can replace this with something like 

```
 * SWATS can optimize differentiable separable functions.  For more details, see
 * the documentation on function types include with this distribution or on the
 * ensmallen website.
```

PR #43, opened 2018-10-28
WNGrad optimizer

zoq:
Implementation of "WNGrad: Learn the Learning Rate in Gradient Descent" by Xiaoxia Wu et al.


include/ensmallen_bits/wn_grad/wn_grad.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * the learning rate can be initialized at a high value, and is subsequently
+ * decreased according to gradient observations.
+ *
+ * For more information, see the following.
+ *
+ * @code
+ * @article{Wu2018,
+ *   author  = {{Wu}, X. and {Ward}, R. and {Bottou}, L.},
+ *   title   = {WNGrad: Learn the Learning Rate in Gradient Descent},
+ *   journal = {ArXiv e-prints},
```
rcurtin: Looks to me like this is NIPS formatting---do you know if the paper got accepted?  Might be nicer to include the NIPS version.  (Although, I suppose the bibliographic reference is not ready yet since the conference has not happened.)

include/ensmallen_bits/wn_grad/wn_grad.hpp
```diff
@@ -0,0 +1,133 @@
...
+   *
+   * @param stepSize Step size for each iteration.
+   * @param batchSize Number of points to process in a single step.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *        limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param shuffle If true, the function order is shuffled; otherwise, each
+   *        function is visited in linear order.
+   */
+  WNGrad(const double stepSize = 0.56,
```
rcurtin: Should we go with the paper's value of 0.562?  (I know it does not make much difference empirically, but someone looking through the code (like me) might get a bit confused otherwise.)

include/ensmallen_bits/wn_grad/wn_grad.hpp
zoq: Right, we should update the reference once there is something to update.

include/ensmallen_bits/wn_grad/wn_grad.hpp
zoq: Good point.

include/ensmallen_bits/wn_grad/wn_grad.hpp
conradsnicta: NIPS seems to be rebranding to NeurIPS: https://neurips.cc/
Didn't realise the previous name was actually a problem until the rebranding.

PR #45, opened 2018-10-31
EVE optimizer

zoq:
Implementation of "Improving Stochastic Gradient Descent with Feedback" by Jayanth Koushik and Hiroaki Hayashi.


include/ensmallen_bits/eve/eve.hpp
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * NumFunctions() should return the number of functions (\f$n\f$), and in the
+ * other two functions, the parameter i refers to which individual function (or
+ * gradient) is being evaluated.  So, for the case of a data-dependent function,
+ * such as NCA, NumFunctions() should return the number of points in the
+ * dataset, and Evaluate(coordinates, 0) will evaluate the objective function on
+ * the first point in the dataset (presumably, the dataset is held internally in
+ * the DecomposableFunctionType).
+ */
+class EVE
```
rcurtin: Do you think we should just call it "Eve" not "EVE"?  THe paper doesn't suggest the name as an acronym, just a reference to 'Adam and Eve' I guess.

include/ensmallen_bits/eve/eve_impl.hpp
```diff
@@ -0,0 +1,171 @@
...
+    epsilon(epsilon),
+    clip(clip),
+    maxIterations(maxIterations),
+    tolerance(tolerance),
+    shuffle(shuffle)
+{ /* Nothing to do. */ }
+
+//! Optimize the function (minimize).
+template<typename DecomposableFunctionType>
+double EVE::Optimize(
```
rcurtin: Ahh, too bad we could not reuse SGD here.  I guess this is because the Eve update rule requires the function value itself.

include/ensmallen_bits/eve/eve.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+  //! Get the tolerance for termination.
+  double Tolerance() const { return tolerance; }
+  //! Modify the tolerance for termination.
+  double& Tolerance() { return tolerance; }
+
+  //! Get whether or not the individual functions are shuffled.
+  bool Shuffle() const { return shuffle; }
+  //! Modify whether or not the individual functions are shuffled.
+  bool& Shuffle() { return shuffle; }
```
rcurtin: We should also make sure the `beta1`, `beta2`, `beta3`, `clip`, and `epsilon` parameters are accessible.

include/ensmallen_bits/eve/eve_impl.hpp
```diff
@@ -0,0 +1,171 @@
...
+    {
+      const double d = std::abs(objective - lastObjective) /
+          std::min(objective, lastObjective);
+
+      dt *= beta3;
+      dt += (1 - beta3) * std::min(std::max(d, 1.0 / clip), clip);
+    }
+    else
+    {
+      lastObjective = objective;
```
rcurtin: Shouldn't this happen every iteration, not just when `i == 0`?

zoq: You are right, nice catch!

zoq: You are right, nice catch!

include/ensmallen_bits/eve/eve.hpp
```diff
@@ -0,0 +1,192 @@
...
+ * @code
+ * @article{Koushik2016,
+ *   author  = {Jayanth Koushik and Hiroaki Hayashi},
+ *   title   = {Improving Stochastic Gradient Descent with Feedback},
+ *   journal = {CoRR},
+ *   year    = {2016},
+ *   url     = {http://arxiv.org/abs/1611.01505}
+ * }
+ *
+ * For EVE to work, a DecomposableFunctionType template parameter is required.
```
rcurtin: I think we can replace this with something like the other optimizers now that just points to the online documentation.

include/ensmallen_bits/eve/eve_impl.hpp
```diff
@@ -0,0 +1,169 @@
...
+    v *= beta2;
+    v += (1 - beta2) * (gradient % gradient);
+
+    const double biasCorrection1 = 1.0 - std::pow(beta1, (double) (i + 1));
+    const double biasCorrection2 = 1.0 - std::pow(beta2, (double) (i + 1));
+
+    if (i > 0)
+    {
+      const double d = std::abs(objective - lastObjective) /
+          std::min(objective, lastObjective);
```
rcurtin: I wonder if this is the source of the nan in the tests.  Maybe it needs some correction in the denominator?

zoq: Right, ``std::min(objective, lastObjective)`` could be zero.

PR #46, opened 2018-11-06
Padam optimizer

zoq:
Implementation of "Closing the Generalization Gap of Adaptive Gradient Methods in Training Deep Neural Networks" by Jinghui Chen and Quanquan Gu.

### DIFF
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -1,0 +1,1792 @@
+## AdaDelta
+
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+AdaDelta is an extension of [AdaGrad](#adagrad) that adapts learning rates
+based on a moving window of gradient updates, instead of accumulating all past
+gradients. Instead of accumulating all past squared gradients, the sum of
+gradients is recursively defined as a decaying average of all past squared
+gradients.
+
+#### Constructors
+
+ * `AdaDelta()`
+ * `AdaDelta(`_`stepSize`_`)`
+ * `AdaDelta(`_`stepSize, batchSize`_`)`
+ * `AdaDelta(`_`stepSize, batchSize, rho, epsilon, maxIterations, tolerance, shuffle`_`)`
+
+#### Attributes
+
+| **type** | **name** | **description** | **default** |
+|----------|----------|-----------------|-------------|
+| `double` | **`stepSize`** | Step size for each iteration. | `1.0` |
+| `size_t` | **`batchSize`**| Number of points to process in one step. | `32` |
+| `double` | **`rho`** | Smoothing constant. Corresponding to fraction of gradient to keep at each time step. | `0.95` |
+| `double` | **`epsilon`** | Value used to initialise the mean squared gradient parameter. | `1e-6` |
+| `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+| `double` | **`tolerance`** | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+| `bool` | **`shuffle`** | If true, the function order is shuffled; otherwise, each function is visited in linear order. | `true` |
+
+Attributes of the optimizer may also be changed via the member methods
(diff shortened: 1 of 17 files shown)


include/ensmallen_bits/adam/padam_update.hpp
```diff
@@ -0,0 +1,153 @@
...
+ * @endcode
+ */
+class PadamUpdate
+{
+ public:
+  /**
+   * Construct the Padam update policy with the given parameters.
+   *
+   * @param epsilon The epsilon value used to initialise the squared gradient
+   *        parameter.
```
rcurtin: Would it be more accurate to say something like "Epsilon is the minimum allowed gradient"?  My understanding of epsilon is that it's necessary to prevent dividing by very small values in the case where v_t is very small.  (Perhaps that might be a documentation update to fix in the AdamUpdate and other similar optimizers---not sure?)

include/ensmallen_bits/adam/padam_update.hpp
```diff
@@ -0,0 +1,153 @@
...
+    v += (1 - beta2) * (gradient % gradient);
+
+    const double biasCorrection1 = 1.0 - std::pow(beta1, iteration);
+    const double biasCorrection2 = 1.0 - std::pow(beta2, iteration);
+
+    // Element wise maximum of past and present squared gradients.
+    vImproved = arma::max(vImproved, v);
+
+    iterate -= (stepSize * std::sqrt(biasCorrection2) / biasCorrection1) *
+        m / arma::pow(arma::sqrt(vImproved) + epsilon, partial * 2);
```
rcurtin: It's possible I have misunderstood something here, but my reading of Algorithm 1 in the paper suggests that this could be `m / arma::pow(vImproved + epsilon, partial)`.  But I think this depends on how the `epsilon` correction should be applied.  Personally I don't think it makes particularly much difference whether it is applied to the square root of `vImproved` or not (the suggestion I wrote would be a little quicker in practice).  But maybe there is some reason I don't know of to write it the way you did?

(In the end I'm indifferent on how it gets implemented, mostly I am just curious here.)

include/ensmallen_bits/adam/padam_update.hpp
```diff
@@ -0,0 +1,153 @@
...
+   */
+  void Update(arma::mat& iterate,
+              const double stepSize,
+              const arma::mat& gradient)
+  {
+    // Increment the iteration counter variable.
+    ++iteration;
+
+    // And update the iterate.
+    m *= beta1;
```
rcurtin: Do we also need to update `beta1` as `beta1 *= lambda` like the theory for Theorem 4.2 requires for convergence?  I guess that Algorithm 1 does not include it, so perhaps it is not necessary.  It also seems like they did not decay `beta1` in the experiments.

include/ensmallen_bits/adam/adam.hpp
```diff
@@ -186,6 +187,8 @@ using NadaMax = AdamType<NadaMaxUpdate>;
 
 using OptimisticAdam = AdamType<OptimisticAdamUpdate>;
 
+using Padam = AdamType<PadamUpdate>;
```
rcurtin: I guess that the user does not have any easy way through the constructor to set `partial`; do you think it is worth writing an extra constructor or something like this?

include/ensmallen_bits/adam/padam_update.hpp
rcurtin: Sounds good to me.  It can always be changed later if someone wants.

rcurtin:
Also, we should probably notify the authors of the paper that we have their optimizer implemented, just so that they know.  (I guess we should probably do that in general!)

doc/optimizers.md
```diff
@@ -195,6 +195,53 @@ optimizer.Optimize(f, coordinates);
  * [Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980) (see section 7)
  * [Differentiable separable functions](#differentiable-separable-functions)
 
+## Padam
```
rcurtin: If you don't mind I think we should move this to alphabetical ordering like the other optimizers.

PR #48, opened 2018-11-08
FTML optimizer

zoq:
Implementation of "Follow the Moving Leader in Deep Learning" by Shuai Zheng and James T. Kwok.

### DIFF
--- a/include/ensmallen_bits/ftml/ftml.hpp
+++ b/include/ensmallen_bits/ftml/ftml.hpp
@@ -74,7 +74,7 @@ class FTML
    * @param beta1 Exponential decay rate for the first moment estimates.
    * @param beta2 Exponential decay rate for the weighted infinity norm
             estimates.
-   * @param eps Value used to initialise the mean squared gradient parameter.
+   * @param epsilon Epsilon is the minimum allowed gradient.
    * @param maxIterations Maximum number of iterations allowed (0 means no
    *        limit).
    * @param tolerance Maximum absolute tolerance to terminate algorithm.
@@ -85,7 +85,7 @@ class FTML
        const size_t batchSize = 32,
        const double beta1 = 0.9,
        const double beta2 = 0.999,
-       const double eps = 1e-8,
+       const double epsilon = 1e-8,
        const size_t maxIterations = 100000,
        const double tolerance = 1e-5,
        const bool shuffle = true);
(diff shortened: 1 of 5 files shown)


include/ensmallen_bits/ftml/ftml.hpp
```diff
@@ -0,0 +1,159 @@
...
+ *   booktitle = {Proceedings of the 34th International Conference on Machine
+ *                Learning},
+ *   pages     = {4110--4119},
+ *   series    = {Proceedings of Machine Learning Research},
+ *   publisher = {PMLR},
+ * }
+ * @endcode
+ *
+ * For FTML to work, a DecomposableFunctionType template parameter is
+ * required. This class must implement the following function:
```
rcurtin: We have this documentation across a lot of optimizers, but the addition of `function_types.md` probably means we can reduce and centralize a lot of it.  Do you think that we could replace this with something like... `FTML requires a separable differentiable function to optimize (see <url>).`  The only problem with that is that it's not clear what URL to use there.  I suppose ideally we'd like to point a user at the website documentation, but, the URL there could change.  So I am not sure what the best choice is.

include/ensmallen_bits/ftml/ftml.hpp
```diff
@@ -0,0 +1,159 @@
...
+   *        limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param shuffle If true, the function order is shuffled; otherwise, each
+   *        function is visited in linear order.
+   */
+  FTML(const double stepSize = 0.001,
+       const size_t batchSize = 32,
+       const double beta1 = 0.9,
+       const double beta2 = 0.999,
+       const double eps = 1e-8,
```
rcurtin: Like for Padam, do you think we should adjust this documentation to point out that it's to avoid divisions by zero?  (Or something like that.)

Also for consistency here I guess we should call it `eps` since there is a method `Epsilon()`.

include/ensmallen_bits/ftml/ftml.hpp
rcurtin: Actually for this one I wonder if it would be best to open another issue instead of handling it in this PR.

zoq:
I'll update ``HISTORY.txt``, once the other PR is merged.

PR #50, opened 2018-11-16
Fix and include GridSearch by default

rcurtin:
I noticed that there were a few bugs in the implementation and that it wasn't included by default.

I wrote and added a test for it also.


rcurtin:
I'll wait until the build is successful, then attempt to use my new script to do an automatic release of ensmallen.

PR #56, opened 2018-11-29
Fix typo in differentiable function docs

neomantra:
Was reading through the docs and noticed that `Evaluate` should probably return a `double`.


rcurtin:
Thanks for pointing this out!  Actually it was just handled in #55 though. :+1:

PR #61, opened 2018-12-05
Clean up old documentation

rcurtin:
This removes all the extra documentation that describes `FunctionType` for each optimizer.  Instead now I changed it to reference the documentation in `function_types.md`, which is more likely to be up to date.

Also I moved a few files relating to the FrankWolfe optimizer into the `fw/` directory.


rcurtin:
I'm not in a hurry, so I think it's fine to wait until #61 is ready until releasing 1.12.0 with these changes.

PR #63, opened 2018-12-29
NIST non-linear least squares problems

zoq:
The National Institute of Standards and Technology non-linear least squares problems. The long term goal is to have an optimizer that is able to solve at least 90% of the problems.


zoq:
For more information see: http://www.itl.nist.gov/div898/strd/nls/nls_info.shtml, which is currently unreachable.

PR #65, opened 2019-01-03
Add newline to end of files to quiet line ending warning.

coatless:
This is a minor PR that adds new lines to the end of two header files that lack them. These header files were being picked up during a nightly `R CMD check` run on the _R_ package (note: new patch deployed for 1.12 and, so, the link may no longer be relevant.)

c.f. https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-clang/RcppEnsmallen-00check.html 

```bash
checking line endings in C/C++/Fortran sources/headers ... NOTE
Found the following sources/headers not terminated with a newline:
....
Some compilers warn on such files.
```


coatless:
Thanks folks!

@rcurtin mind sharing the release script?

rcurtin:
Sure, actually I put it in the ensmallen.org repository: 
https://github.com/mlpack/ensmallen.org/blob/master/scripts/release-ensmallen.sh

Since the mlpack release process is so tedious, I worked to make the ensmallen process the opposite of that.  I'd like to move the mlpack process to something that looks more like this... but there is a long way to go before that. :)

coatless:
@rcurtin thank you! That's a very useful script. 👍

PR #66, opened 2019-01-06
Added Augmented Lagrangian options

lukasmack0:
Added optionality to set the L-BFGS iterations, the penalty update factor and the sigma update factor in the augmented Lagrangian class. Kept defaults the same as the originally hardcoded values.

### DIFF
--- a/include/ensmallen_bits/aug_lagrangian/aug_lagrangian.hpp
+++ b/include/ensmallen_bits/aug_lagrangian/aug_lagrangian.hpp
@@ -31,23 +31,14 @@ namespace ens {
  * ensmallen website.
  */
 
+// TODO expose LBFGS params
+// TODO finalise docs update
+
 class AugLagrangian
 {
  public:
   /**
    * Initialize the Augmented Lagrangian with the default L-BFGS optimizer.
-   */
-  AugLagrangian();
-
-  /**
-   * Optimize the function.  The value '1' is used for the initial value of each
-   * Lagrange multiplier.  To set the Lagrange multipliers yourself, use the
-   * other overload of Optimize().
-   *
-   * @tparam LagrangianFunctionType Function which can be optimized by this
-   *     class.
-   * @param function The function to optimize.
-   * @param coordinates Output matrix to store the optimized coordinates in.
    * @param penaltyThresholdFactor When the penalty threshold is updated set
    *    the penalty threshold to the penalty multplied by this factor. The
    *    default value of 0.25 is is taken from Burer and Monteiro (2002).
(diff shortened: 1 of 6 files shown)


include/ensmallen_bits/aug_lagrangian/aug_lagrangian.hpp
```diff
@@ -49,13 +48,24 @@ class AugLagrangian
...
    *     Lagrangian algorithm.  0 indicates no maximum.
+   * @param internalMaxIterations Maximum number of iterations of L-BFGS
+   *    iterations internal to the Augmented Lagrangian algorithm.
+   *    0 indicates no maximum.
    */
   template<typename LagrangianFunctionType>
   bool Optimize(LagrangianFunctionType& function,
                 arma::mat& coordinates,
-                const size_t maxIterations = 1000);
+                const double penaltyThresholdFactor = 0.25,
```
zoq: Nice catch, I think everything should go into the class constructor, to be consistent with the other optimizers. Let me know what you think.

lukasmack0: Yeah makes sense, I'll move the algorithm arguments across (i.e. maxIterations, defaultMaxIterations, penaltyThresholdFactor and sigmaUpdateFactor).

Happy with the initSigma and initLambda being in the function call still rather than the constructor?

zoq: We can move the parameter as well, and provide a function for each to get/modify the value.

include/ensmallen_bits/aug_lagrangian/aug_lagrangian.hpp
```diff
@@ -49,13 +48,24 @@ class AugLagrangian
...
+   * @param penaltyThresholdFactor When the penalty threshold is updated set
+   *    the penalty threshold to the penalty multplied by this factor. The
+   *    default value of 0.25 is is taken from Burer and Monteiro (2002).
+   * @param sigmaUpdateFactor When sigma is updated  multiply sigma by this
+   *    value. The default value of 10 is taken from Burer and Monteiro (2002).
    * @param maxIterations Maximum number of iterations of the Augmented
    *     Lagrangian algorithm.  0 indicates no maximum.
+   * @param internalMaxIterations Maximum number of iterations of L-BFGS
+   *    iterations internal to the Augmented Lagrangian algorithm.
+   *    0 indicates no maximum.
```
rcurtin: We can actually expose all of the members of the internal L-BFGS optimizer by having a parameter `lbfgs` of type `const L_BFGS&`.  Then, a user can set the maximum iterations of the L-BFGS optimizer in two ways:

```
L_BFGS l(..., 1000 /* max iterations */, ...);
AugLagrangian aug(..., l);
```

Or they can do it like this:

```
AugLagrangian aug(...);
aug.LBFGS().MaxIterations() = 1000;
```

A nice benefit to that is they can also access all the other internal members of the L-BFGS optimizer.

include/ensmallen_bits/aug_lagrangian/aug_lagrangian.hpp
```diff
@@ -49,13 +48,24 @@ class AugLagrangian
    *     class.
    * @param function The function to optimize.
    * @param coordinates Output matrix to store the optimized coordinates in.
+   * @param penaltyThresholdFactor When the penalty threshold is updated set
+   *    the penalty threshold to the penalty multplied by this factor. The
+   *    default value of 0.25 is is taken from Burer and Monteiro (2002).
+   * @param sigmaUpdateFactor When sigma is updated  multiply sigma by this
+   *    value. The default value of 10 is taken from Burer and Monteiro (2002).
```
rcurtin: It's really nice to provide customizable values for these.  My experiences working with the AugLagrangian optimizer when solving LRSDPs indicated that it can be really sensitive to these parameters, but for whatever reason I did not make them customizable when I wrote it.  So this is a nice improvement. :+1:

include/ensmallen_bits/aug_lagrangian/aug_lagrangian.hpp
lukasmack0: I'll remove my exposing of the LBFGS max iterations.

That first constructor which takes an LBFGS object doesn't currently exist though, shall I add it?

rcurtin: Right, exactly, we'd need to add that as a constructor argument.  We can have a default too, so you can just modify the existing constructor and add `const L_BFGS& lbfgs = L_BFGS()`.  :+1:

rcurtin:
Looks good to me, but I think that some of the things that used `AugLagrangian` need to be updated since the build is failing: https://travis-ci.org/mlpack/ensmallen/jobs/476513780

If you can fix that please then I think it is ready for merge. :+1:

lukasmack0:
Looks like all builds are passing now :+1:

PR #79, opened 2019-02-06
Implement weight decay in SGD and Adam (WIP)

chigur:
Hi,

This pull request implements the optimizers in the paper [Decoupled Weight Decay Regularization](https://arxiv.org/pdf/1711.05101.pdf). It's still a work in progress.

Right now SGDW is implemented in the most straightforward way, but I am not sure if that's the best way. Perhaps, we should decouple weight decay from specific update policies and add it to `SGD<>::Optimize` with a variable that controls if weight decay step happens or not. That way weight decay can be added to any update policy.

---

Tasks to complete
- [ ] Finalize the design
- [ ] Complete the SGDW
- [ ] Add tests for SGDW
- [ ] Complete AdamW
- [ ] Add tests for AdamW
- [ ] Update the documentation for both optimizers

### DIFF
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -65,7 +65,7 @@ parameters.
  - `AdaGrad(`_`stepSize, batchSize`_`)`
  - `AdaGrad(`_`stepSize, batchSize, epsilon, maxIterations, tolerance, shuffle`_`)`
  - `AdaGrad(`_`stepSize, batchSize, epsilon, maxIterations, tolerance, shuffle, resetPolicy`_`)`
- 
+
 #### Attributes
 
 | **type** | **name** | **description** | **default** |
@@ -132,7 +132,7 @@ with _`UpdateRule`_` = AdamUpdate`.
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 The attributes of the optimizer may also be modified via the member methods
-`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
+`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`, `DecayPolicy()`
 `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
 
 #### Examples
@@ -151,6 +151,60 @@ optimizer.Optimize(f, coordinates);
  * [SGD](#standard-sgd)
  * [Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980)
  * [Differentiable separable functions](#differentiable-separable-functions)
+
+ ## AdamR
+
+ *An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+ AdamR is a modified version of Adam with warm restarts. This allows the optimiser to have a automated decreasing
+ step size without modification. AdamR uses the Cyclical Decay decay policy.
+
+ #### Constructors
+
+  * `AdamR()`
+  * `AdamR(`_`stepSizeMax, stepSizeMax, batchSize`_`)`
+  * `AdamR(`_`stepSizeMax, stepSizeMax, batchSize, beta1, beta2, eps, maxIterations, tolerance, shuffle`_`)`
+  * `AdamR(`_`stepSizeMax, stepSizeMax, batchSize, beta1, beta2, eps, maxIterations, tolerance, shuffle, resetPolicy`_`)`
+
+ Note that the `AdamR` class is based on the `AdamRType<`_`UpdateRule`_`>` class
+ with _`UpdateRule`_` = AdamUpdate`.
+
+ #### Attributes
+
+ | **type** | **name** | **description** | **default** |
+ |----------|----------|-----------------|-------------|
+ | `double` | **`stepSizeMax`** | Intitial step size for each iteration. | `0.02` |
+ | `double` | **`stepSizeMin`** | Final step size for each iteration. | `0.01` |
+ | `size_t` | **`batchSize`** | Number of points to process in a single step. | `32` |
+ | `double` | **`beta1`** | Exponential decay rate for the first moment estimates. | `0.9` |
+ | `double` | **`beta2`** | Exponential decay rate for the weighted infinity norm estimates. | `0.999` |
+ | `double` | **`eps`** | Value used to initialize the mean squared gradient parameter. | `1e-8` |
+ | `size_t` | **`max_iterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+ | `double` | **`tolerance`** | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+ | `bool` | **`shuffle`** | If true, the function order is shuffled; otherwise, each function is visited in linear order. | `true` |
+ | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
+
+ The attributes of the optimizer may also be modified via the member methods
+ `StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`, `DecayPolicy()`
+ `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
+
+ #### Examples
+
+ ```c++
+ RosenbrockFunction f;
+ arma::mat coordinates = f.GetInitialPoint();
+
+ AdamR optimizer(0.01, 0.005, 32, 0.9, 0.999, 1e-8, 100000, 1e-5, true);
+ optimizer.Optimize(f, coordinates);
+ ```
+
+ #### See also:
+
+  * [Decoupled Weight Decay Regularization](https://arxiv.org/abs/1711.05101)
+  * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
+  * [SGD](#standard-sgd)
+  * [Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980)
+  * [Differentiable separable functions](#differentiable-separable-functions)
 
 ## AdaMax
 
@@ -183,7 +237,7 @@ with _`UpdateRule`_` = AdaMaxUpdate`.
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 The attributes of the optimizer may also be modified via the member methods
-`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
+`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`, `DecayPolicy()`
 `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
 
 #### Examples
@@ -234,7 +288,7 @@ with _`UpdateRule`_` = AMSGradUpdate`.
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 The attributes of the optimizer may also be modified via the member methods
-`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
+`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`, `DecayPolicy()`
 `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
 
 #### Examples
@@ -890,7 +944,7 @@ proximalOptimizer.Optimize(f, coordinates);
 *An optimizer for [differentiable functions](#differentiable-functions)*
 
 L-BFGS is an optimization algorithm in the family of quasi-Newton methods that approximates the Broydenâ€“Fletcherâ€“Goldfarbâ€“Shanno (BFGS) algorithm using a limited amount of computer memory.  
-  
+
 #### Constructors
 
  * `L_BFGS()`
@@ -1061,7 +1115,7 @@ with _`UpdateRule`_` = NadamUpdate`.
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 The attributes of the optimizer may also be modified via the member methods
-`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
+`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`, `DecayPolicy()`
 `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
 
 #### Examples
@@ -1113,7 +1167,7 @@ with _`UpdateRule`_` = NadaMaxUpdate`.
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 The attributes of the optimizer may also be modified via the member methods
-`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
+`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`, `DecayPolicy()`
 `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
 
 #### Examples
@@ -1552,7 +1606,9 @@ For convenience the following typedefs have been defined:
 Attributes of the optimizer may also be changed via the member methods
 `StepSize()`, `BatchSize()`, `MaxIterations()`, `InnerIterations()`,
 `Tolerance()`, `Shuffle()`, and `UpdatePolicy()`.
-
+The attributes of the optimizer may also be modified via the member methods
+`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
+`Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
 Note that the default value for `updatePolicy` is the default constructor for
 the `UpdatePolicyType`.
 
@@ -1697,14 +1753,14 @@ cyclicscd.Optimize(f, coordinates);
 
 *An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
 
-SGDR is based on Mini-batch Stochastic Gradient Descent class and simulates a new warm-started run/restart once a number of epochs are performed. 
+SGDR is based on Mini-batch Stochastic Gradient Descent class and simulates a new warm-started run/restart once a number of epochs are performed.
 
 #### Constructors
 
  * `SGDR<`_`UpdatePolicyType`_`>()`
- * `SGDR<`_`UpdatePolicyType`_`>(`_`epochRestart, multFactor, batchSize, stepSize`_`)`
- * `SGDR<`_`UpdatePolicyType`_`>(`_`epochRestart, multFactor, batchSize, stepSize, maxIterations, tolerance, shuffle, updatePolicy`_`)`
- * `SGDR<`_`UpdatePolicyType`_`>(`_`epochRestart, multFactor, batchSize, stepSize, maxIterations, tolerance, shuffle, updatePolicy`_`, resetPolicy)`
+ * `SGDR<`_`UpdatePolicyType`_`>(`_`epochRestart, multFactor, batchSize, stepSizeMax, stepSizeMin`_`)`
+ * `SGDR<`_`UpdatePolicyType`_`>(`_`epochRestart, multFactor, batchSize, stepSizeMax, stepSizeMin, maxIterations, tolerance, shuffle, updatePolicy`_`)`
+ * `SGDR<`_`UpdatePolicyType`_`>(`_`epochRestart, multFactor, batchSize, stepSizeMax, stepSizeMin, maxIterations, tolerance, shuffle, updatePolicy`_`, resetPolicy)`
 
 The _`UpdatePolicyType`_ template parameter controls the update policy used
 during the iterative update process.  The `MomentumUpdate` class is available
@@ -1722,7 +1778,8 @@ so the shorter type `SGDR<>` can be used instead of the equivalent
 | `size_t` | **`epochRestart`** | Initial epoch where decay is applied. | `50` |
 | `double` | **`multFactor`** | Batch size multiplication factor. | `2.0` |
 | `size_t` | **`batchSize`** | Size of each mini-batch. | `1000` |
-| `double` | **`stepSize`** | Step size for each iteration. | `0.01` |
+| `double` | **`stepSizeMax`** | Intitial step size for each iteration. | `0.01` |
+| `double` | **`stepSizeMin`** | Final step size for each iteration. | `0.005` |
 | `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
 | `double` | **`tolerance`** | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
 | `bool` | **`shuffle`** | If true, the mini-batch order is shuffled; otherwise, each mini-batch is visited in linear order. | `true` |
@@ -1743,7 +1800,7 @@ the `UpdatePolicyType`.
 RosenbrockFunction f;
 arma::mat coordinates = f.GetInitialPoint();
 
-SGDR<> optimizer(50, 2.0, 1, 0.01, 10000, 1e-3);
+SGDR<> optimizer(50, 2.0, 1, 0.01, 0.005, 10000, 1e-3);
 optimizer.Optimize(f, coordinates);
 ```
 
@@ -1752,6 +1809,120 @@ optimizer.Optimize(f, coordinates);
  * [SGDR: Stochastic Gradient Descent with Warm Restarts](https://arxiv.org/abs/1608.03983)
  * [Stochastic gradient descent in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
  * [Differentiable separable functions](#differentiable-separable-functions)
+
+ ## Decoupled Weight Decay Momentum SGD (SGDW)
+
+ *An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+ Stochastic Gradient Descent is a technique for minimizing a function which
+ can be expressed as a sum of other functions. This is modified version of SGD which decouples the
+ weight decay.
+
+ #### Constructors
+
+  * `SGDW()`
+  * `SGDW(`_`stepSize, batchSize`_`)`
+  * `SGDW(`_`stepSize, batchSize, maxIterations, tolerance, shuffle`_`)`
+  * `SGDW(`_`stepSize, batchSize, maxIterations, tolerance, shuffle, momentumPolicy`_`)`
+
+ Note that `MomentumSGD` is based on the templated type
+ `SGD<`_`UpdatePolicyType, DecayPolicyType`_`>` with _`UpdatePolicyType`_` =
+ DecoupledWeightDecayMomentumUpdate` and _`DecayPolicyType`_` = NoDecay`.
+
+ #### Attributes
+
+ | **type** | **name** | **description** | **default** |
+ |----------|----------|-----------------|-------------|
+ | `double` | **`stepSize`** | Step size for each iteration. | `0.01` |
+ | `size_t` | **`batchSize`** | Batch size to use for each step. | `32` |
+ | `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+ | `double` | **`tolerance`** | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+ | `bool` | **`shuffle`** | If true, the function order is shuffled; otherwise, each function is visited in linear order. | `true` |
+ | `DecoupledWeightDecayMomentumUpdate` | **`updatePolicy`** | An instantiated `MomentumUpdate`. | `DecoupledWeightDecayMomentumUpdate()` |
+
+ Attributes of the optimizer may also be modified via the member methods
+ `StepSize()`, `BatchSize()`, `MaxIterations()`, `Tolerance()`, `Shuffle()`, and
+ `UpdatePolicy()`.
+
+ Note that the `DecoupledWeightDecayMomentumUpdate` class has the constructor
+ `DecoupledWeightDecayMomentumUpdate(`_`momentum, weightDecay`_`)` with a
+ default value of `0.5` for the momentum and `0.0005` for the weight decay.
+
+ #### Examples
+
+ ```c++
+ RosenbrockFunction f;
+ arma::mat coordinates = f.GetInitialPoint();
+
+ SGDW optimizer(0.01, 32, 100000, 1e-5, true, DecoupledWeightDecayMomentumUpdate(0.5, 0.0003));
+ optimizer.Optimize(f, coordinates);
+ ```
+
+ #### See also:
+
+  * [Decoupled Weight Decay Regularization](https://arxiv.org/abs/1711.05101)
+  * [Standard SGD](#standard-sgd)
+  * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
+  * [Differentiable separable functions](#differentiable-separable-functions)
+
+ ## Decoupled Weight Decay Stochastic Gradient Descent with Restarts (SGDWR)
+
+ *An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+ SGDWR is a modified version of SGDR with the weight decay decoupled.
+
+ #### Constructors
+
+  * `SGDWR<`_`UpdatePolicyType`_`>()`
+  * `SGDWR<`_`UpdatePolicyType`_`>(`_`epochRestart, multFactor, batchSize, stepSizeMax, stepSizeMin`_`)`
+  * `SGDWR<`_`UpdatePolicyType`_`>(`_`epochRestart, multFactor, batchSize, stepSizeMax, stepSizeMin, maxIterations, tolerance, shuffle, updatePolicy`_`)`
+  * `SGDWR<`_`UpdatePolicyType`_`>(`_`epochRestart, multFactor, batchSize, stepSizeMax, stepSizeMin, maxIterations, tolerance, shuffle, updatePolicy`_`, resetPolicy)`
+
+ The _`UpdatePolicyType`_ template parameter controls the update policy used
+ during the iterative update process.
+
+  #### Attributes
+
+ | **type** | **name** | **description** | **default** |
+ |----------|----------|-----------------|-------------|
+ | `size_t` | **`epochRestart`** | Initial epoch where decay is applied. | `50` |
+ | `double` | **`multFactor`** | Batch size multiplication factor. | `2.0` |
+ | `size_t` | **`batchSize`** | Size of each mini-batch. | `1000` |
+ | `double` | **`stepSizeMax`** | Intitial step size for each iteration. | `0.01` |
+ | `double` | **`stepSizeMin`** | Final step size for each iteration. | `0.005` |
+ | `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+ | `double` | **`tolerance`** | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+ | `bool` | **`shuffle`** | If true, the mini-batch order is shuffled; otherwise, each mini-batch is visited in linear order. | `true` |
+ | `UpdatePolicyType` | **`updatePolicy`** | Instantiated update policy used to adjust the given parameters. | `UpdatePolicyType()` |
+ | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
+
+ Attributes of the optimizer can also be modified via the member methods
+ `EpochRestart()`, `MultFactor()`, `BatchSize()`, `StepSize()`,
+ `MaxIterations()`, `Tolerance()`, `Shuffle()`, `UpdatePolicy()`, and
+ `ResetPolicy()`.
+
+ Note that the default value for `updatePolicy` is the default constructor for
+ the `UpdatePolicyType`.
+
+ Note that the `SGDWR` class is based on the `SGDR<`_`UpdateRule`_`>` class
+ with _`UpdateRule`_` = DecoupledWeightDecayMomentumUpdate`.
+
+ #### Examples:
+
+ ```c++
+ RosenbrockFunction f;
+ arma::mat coordinates = f.GetInitialPoint();
+
+ SGDWR optimizer(50, 2, 32, 0.002, 0.001, 0, 1e-9, true, DecoupledWeightDecayMomentumUpdate(0.5, 0.0003));
+ optimizer.Optimize(f, coordinates);
+ ```
+
+ #### See also:
+
+  * [Decoupled Weight Decay Regularization](https://arxiv.org/abs/1711.05101)
+  * [SGDR: Stochastic Gradient Descent with Warm Restarts](https://arxiv.org/abs/1608.03983)
+  * [Stochastic gradient descent in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
+  * [Differentiable separable functions](#differentiable-separable-functions)
 
 ## Snapshot Stochastic Gradient Descent with Restarts (SnapshotSGDR)
 
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -157,7 +157,7 @@ optimizer.Optimize(f, coordinates);
 *An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
 
 AdamR is a modified version of Adam with warm restarts. This allows the optimiser to have a automated decreasing
-step size without modification. AdamR uses the Cyclical Decay decay policy.
+step size without modification. AdamR uses Cyclical Decay as its default decay policy.
 
 #### Constructors
 
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -132,7 +132,7 @@ with _`UpdateRule`_` = AdamUpdate`.
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 The attributes of the optimizer may also be modified via the member methods
-`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`,
+`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
 `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
 
 #### Examples
@@ -187,7 +187,7 @@ with _`UpdateRule`_` = AdamUpdate`.
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 The attributes of the optimizer may also be modified via the member methods
- `StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`, `DecayPolicy()`
+ `StepSizeMax()`, `StepSizeMin()`, `EpochRestart()`, `MultFactor()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`, `DecayPolicy()`
  `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
 
 #### Examples
@@ -206,14 +206,14 @@ The attributes of the optimizer may also be modified via the member methods
   * [Decoupled Weight Decay Regularization](https://arxiv.org/abs/1711.05101)
   * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
   * [SGD](#standard-sgd)
-  * [Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980)
+  * [Adam](#adam)
   * [Differentiable separable functions](#differentiable-separable-functions)
 
 ## AdamW
 
 *An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
 
- AdamW is a modified version of Adam with decoupled weight decay.
+ AdamW is a modified version of Adam with decoupled weight decay, which tends to perform better than adam with standard L2 regularization.
 
 #### Constructors
 
@@ -239,7 +239,7 @@ The attributes of the optimizer may also be modified via the member methods
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 The attributes of the optimizer may also be modified via the member methods
-`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
+`StepSize()`, `BatchSize()`, `WeightDecay()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
 `Tolerance()`, `Shuffle()`, `WeightDecay()` and `ResetPolicy()`.
 
 #### Examples
@@ -262,10 +262,9 @@ The attributes of the optimizer may also be modified via the member methods
 
 ## AdamWR
 
- *An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
 
- AdamWR is a modified version of AdamW with warm restarts and is combination of AdamW and AdamR
-
+AdamWR is a modified version of AdamW with warm restarts and is combination of AdamW and AdamR, performs almost as well as SGD on large problems.
 
 #### Constructors
 
@@ -294,7 +293,7 @@ The attributes of the optimizer may also be modified via the member methods
   | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
   The attributes of the optimizer may also be modified via the member methods
-   `StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `WeightDecay()`,
+   `StepSizeMax()`, `StepSizeMin()`, `EpochRestart`, `MultFactor`, `WeightDecay()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
    `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
 
  #### Examples
@@ -315,6 +314,8 @@ The attributes of the optimizer may also be modified via the member methods
  * [SGD](#standard-sgd)
  * [Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980)
  * [Differentiable separable functions](#differentiable-separable-functions)
+ * [AdamW](#adamw)
+ * [AdamR](#adamr)
 
 ## AdaMax
 
@@ -1935,7 +1936,7 @@ so the shorter type `SGDR<>` can be used instead of the equivalent
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 Attributes of the optimizer can also be modified via the member methods
-`EpochRestart()`, `MultFactor()`, `BatchSize()`, `StepSize()`,
+`EpochRestart()`, `MultFactor()`, `BatchSize()`, `StepSizeMax()`, `StepSizeMin()`,
 `MaxIterations()`, `Tolerance()`, `Shuffle()`, `UpdatePolicy()`, and
 `ResetPolicy()`.
 
@@ -2036,6 +2037,7 @@ during the iterative update process.
 | `size_t` | **`epochRestart`** | Initial epoch where decay is applied. | `50` |
 | `double` | **`multFactor`** | Batch size multiplication factor. | `2.0` |
 | `size_t` | **`batchSize`** | Size of each mini-batch. | `1000` |
+| `double` | **`weightDecay`** | Decay rate for the parameters. | `0.0005` |
 | `double` | **`stepSizeMax`** | Intitial step size for each iteration. | `0.01` |
 | `double` | **`stepSizeMin`** | Final step size for each iteration. | `0.005` |
 | `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
@@ -2045,7 +2047,7 @@ during the iterative update process.
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 Attributes of the optimizer can also be modified via the member methods
-`EpochRestart()`, `MultFactor()`, `BatchSize()`, `StepSize()`,
+`EpochRestart()`, `MultFactor()`, `BatchSize()`, `StepSizeMax()`, `StepSizeMin()`,
 `MaxIterations()`, `Tolerance()`, `Shuffle()`, `UpdatePolicy()`, and
 `ResetPolicy()`.
 
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -162,9 +162,9 @@ step size without modification. AdamR uses the Cyclical Decay decay policy.
 #### Constructors
 
  * `AdamR()`
- * `AdamR(`_`stepSizeMax, stepSizeMax, batchSize`_`)`
- * `AdamR(`_`stepSizeMax, stepSizeMax, batchSize, beta1, beta2, eps, maxIterations, tolerance, shuffle`_`)`
- * `AdamR(`_`stepSizeMax, stepSizeMax, batchSize, beta1, beta2, eps, maxIterations, tolerance, shuffle, resetPolicy`_`)`
+ * `AdamR(`_`stepSizeMax, stepSizeMin, epochRestart, multFactor, batchSize`_`)`
+ * `AdamR(`_`stepSizeMax, stepSizeMin, epochRestart, multFactor, batchSize, beta1, beta2, eps, maxIterations, tolerance, shuffle`_`)`
+ * `AdamR(`_`stepSizeMax, stepSizeMin, epochRestart, multFactor, batchSize, beta1, beta2, eps, maxIterations, tolerance, shuffle, resetPolicy`_`)`
 
 Note that the `AdamR` class is based on the `AdamRType<`_`UpdateRule`_`>` class
 with _`UpdateRule`_` = AdamUpdate`.
@@ -175,6 +175,8 @@ with _`UpdateRule`_` = AdamUpdate`.
 |----------|----------|-----------------|-------------|
 | `double` | **`stepSizeMax`** | Intitial step size for each iteration. | `0.02` |
 | `double` | **`stepSizeMin`** | Final step size for each iteration. | `0.01` |
+| `size_t` | **`epochRestart`** | Size of initial iteration. | `50` |
+| `double` | **`multFactor`** | Rate at which iteration size changes. | `2.0` |
 | `size_t` | **`batchSize`** | Number of points to process in a single step. | `32` |
 | `double` | **`beta1`** | Exponential decay rate for the first moment estimates. | `0.9` |
 | `double` | **`beta2`** | Exponential decay rate for the weighted infinity norm estimates. | `0.999` |
@@ -194,12 +196,13 @@ The attributes of the optimizer may also be modified via the member methods
  RosenbrockFunction f;
  arma::mat coordinates = f.GetInitialPoint();
 
- AdamR optimizer(0.01, 0.005, 32, 0.9, 0.999, 1e-8, 100000, 1e-5, true);
+ AdamR optimizer(0.01, 0.005, 100, 1.5, 32, 0.9, 0.999, 1e-8, 100000, 1e-5, true);
  optimizer.Optimize(f, coordinates);
  ```
 
 #### See also:
 
+  * [SGDR: Stochastic Gradient Descent with Warm Restarts](https://arxiv.org/abs/1608.03983)
   * [Decoupled Weight Decay Regularization](https://arxiv.org/abs/1711.05101)
   * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
   * [SGD](#standard-sgd)
@@ -216,11 +219,9 @@ The attributes of the optimizer may also be modified via the member methods
 
  * `AdamW()`
  * `AdamW(`_`stepSize, batchSize`_`)`
- * `AdamW(`_`stepSize, batchSize, beta1, beta2, eps, maxIterations, tolerance, shuffle`_`)`
- * `AdamW(`_`stepSize, batchSize, beta1, beta2, eps, maxIterations, tolerance, shuffle, resetPolicy`_`)`
+ * `AdamW(`_`stepSize, batchSize, weightDecay, beta1, beta2, eps, maxIterations, tolerance, shuffle`_`)`
+ * `AdamW(`_`stepSize, batchSize, weightDecay, beta1, beta2, eps, maxIterations, tolerance, shuffle, resetPolicy`_`)`
 
- Note that the `AdamW` class is based on the `AdamType<`_`UpdateRule`_`>` class
- with _`UpdateRule`_` = AdamWUpdate`.
 
 #### Attributes
 
@@ -228,6 +229,7 @@ The attributes of the optimizer may also be modified via the member methods
 |----------|----------|-----------------|-------------|
 | `double` | **`stepSize`** | Step size for each iteration. | `0.001` |
 | `size_t` | **`batchSize`** | Number of points to process in a single step. | `32` |
+| `double` | **`weightDecay`** | Decay rate for the parameters. | `0.0005` |
 | `double` | **`beta1`** | Exponential decay rate for the first moment estimates. | `0.9` |
 | `double` | **`beta2`** | Exponential decay rate for the weighted infinity norm estimates. | `0.999` |
 | `double` | **`eps`** | Value used to initialize the mean squared gradient parameter. | `1e-8` |
@@ -237,8 +239,8 @@ The attributes of the optimizer may also be modified via the member methods
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 The attributes of the optimizer may also be modified via the member methods
-`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`, `DecayPolicy()`
-`Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
+`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
+`Tolerance()`, `Shuffle()`, `WeightDecay()` and `ResetPolicy()`.
 
 #### Examples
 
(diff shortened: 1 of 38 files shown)


zoq:
It looks like it would be possible to decouple the method from specific policies, and implement it as a decay strategy, if we use e fixed decay schedule, which we do anyway right now.

chigur:
@zoq Are you suggesting a `DecayPolicy` like `NoDecay`? If yes, I don't think that will work. In fact my earlier suggestion of modifying `SGD<>::Optimize` won't work either since parameters are modified by the `Update` method of  `UpdatePolicy`. Thus, any weight decay code has to execute inside that method.

I think the decoupling may not be as straightforward as I had thought.

niteya-shah:
@chigur hey I was planning to work on this too , so do u mind if we collaborate ?

chigur:
@niteya-shah Sure, we can collaborate.
Correct me if I am wrong, but the WR versions only differ in the way learning rate is decayed, right? If yes, then there is already infrastructure for providing a learning rate decay policy.

niteya-shah:
SGDR is implmented and it already uses a templated update policy so using corrent method will work quite well , but I didnt see any reference of AdamR so we will have to create it . SGDR is implemented as a different policy altogether so I think we should follow the current style and define AdamR with templated coupling update strategy

chigur:
Ash. I see now.

niteya-shah:
can you add me as a collaborator?

zoq:
@chigur  I was thinking to use the existing adam update policy as the update step (red, 1), and implement the second step (2, blue) as decay which does, ``iterate -= m * g;``, maybe I missed something?

adamw

zoq:
@niteya-shah In case you are interested to work on another Adam flavour, https://arxiv.org/pdf/1810.06801v3.pdf might be interesting as well.

chigur:
@zoq No, you're right. I was a bit confused there.

chigur:
@zoq Assuming `g` is the gradient, shouldn't it be `iterate -= m * iterate`? Because `Update` method changes `iterate` inplace, we might have to keep a copy of the paramaters.

niteya-shah:
@zoq Thanks for the answering my query on IRC , but my doubt is slightly different 
image
from the deeplearning book of MIT (pg 293)
image
from Sebastian Ruder
These are two different update policies for SGD with momentum . The momentum component is added to the gradient in the Ruder Paper while subtracted in the Book . Are these both valid ? Our Decoupled Weights Seems to follow the Sebastian Ruder Paper so I am confused . I hope my query is a bit more verbose now.

Also thanks for the adam Paper , I went through it and it looks really cool and would love to work to on that as well ( I have a break coming ) 
@chigur I think the decay policy is correct as @zoq mentioned , L2 regularisation should happen during Gradient computation and I think that should be in the function itself , whereas the Decoupled Weight Decay is something that the decay policy should handle.

zoq:
> @zoq Assuming `g` is the gradient, shouldn't it be `iterate -= m * iterate`? Because `Update` method changes `iterate` inplace, we might have to keep a copy of the paramaters.

You are right, sorry for the confusion.

include/ensmallen_bits/sgd/update_policies/weight_decay_momentum_update.hpp
```diff
@@ -82,8 +85,8 @@ class WeightDecayMomentumUpdate
               const double stepSize,
               const arma::mat& gradient)
   {
-    velocity = momentum * velocity - stepSize * gradient;
-    iterate += velocity - stepSize * weight_decay * iterate;
+    velocity = momentum * velocity + stepSize * gradient;
+    iterate += - velocity - stepSize * weightDecay * iterate;
```
chigur: You may have noticed it, but we can change this to `iterate -=`

include/ensmallen_bits/sgd/update_policies/weight_decay_momentum_update.hpp
```diff
@@ -15,15 +15,18 @@
 namespace ens {
 
 /**
- * Momentum update for Stochastic Gradient Descent with weight decay.
+ * De-Coupled Weight Decay Policy for SGD with Momentum (SGDR).
```
chigur: Couple to nitpicks:

1. Doesn't SGDR suggest SGD with restarts? If this is not a mistake, ignore me.
2. Does it need to be named de-coupled weight decay policy? The way it's implemented it's just a momentum update which also has weight decay. This could be a potential source of confusion later. Just a thought.

niteya-shah: 1. Yeah SGDR is without decay 
2. Weight Decay is kind of confusing for SGD as both L2 and Weight Decay are equivalent as there is only 1 training value per step , the change will make sense for batches . It doesnt matter if you implment this in Gradient Calculation or during Update but maybe if this paper is built upon in the future then I think this format will atleast provide more backbone as the paper states that this method of Updating can be used for all optimisers , they tried it for SGD and Adam.  
I hope you understand what I am trying to say.

niteya-shah:
@chigur I have finished the SGDWR part , go through it once . Till then ill think on how to implment AdamWR

niteya-shah:
@chigur The AdamW and AdamWR is implmented but I am not getting good results can you check the code ?

niteya-shah:
@zoq @chigur The Implmentations are both ready for review can you please check them .
Some Notes
1) The Cyclical Decay had bugs so I fixed them , It should now perform as the paper intended
2) The Paper used a parameter of min step Size which was missing in the Implmentation , and I had to overload the constructors, please tell me how I can improve this , the logic and implmentation is sound but I dont know if this is the correct design choice.
3) The new methods are AdamW, AdamWR , SGDW , SGDWR
4) The changes to parameter location is to allow for simpler code , I did this is on personnal choice , tell me if you dont like it I will change it `back```

include/ensmallen_bits/adam/adamw_update.hpp
```diff
@@ -0,0 +1,149 @@
...
+class AdamWUpdate
+{
+ public:
+  /**
+   * Construct the Adam update policy with the given parameters.
+   *
+   * @param epsilon The epsilon value used to initialise the squared gradient
+   *        parameter.
+   * @param beta1 The smoothing parameter.
+   * @param beta2 The second moment coefficient.
```
zoq: The ``weightDecay`` parameter is missing.

include/ensmallen_bits/adam/adamw_update.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * It should be noted that the term, m / (arma::sqrt(v) + eps), in the
+     * following expression is an approximation of the following actual term;
+     * m / (arma::sqrt(v) + (arma::sqrt(biasCorrection2) * eps).
+     */
+    iterate -= ((stepSize * std::sqrt(biasCorrection2) / biasCorrection1) *
+        m / (arma::sqrt(v) + epsilon)) + weightDecay * iterate;
+
+    //std::cout<<stepSize<<std::endl;
```
zoq: Looks like we can remove the output here.

include/ensmallen_bits/adam/adamw_update.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+    const double biasCorrection1 = 1.0 - std::pow(beta1, iteration);
+    const double biasCorrection2 = 1.0 - std::pow(beta2, iteration);
+
+    /**
+     * It should be noted that the term, m / (arma::sqrt(v) + eps), in the
+     * following expression is an approximation of the following actual term;
+     * m / (arma::sqrt(v) + (arma::sqrt(biasCorrection2) * eps).
+     */
+    iterate -= ((stepSize * std::sqrt(biasCorrection2) / biasCorrection1) *
```
zoq: Just an idea to reduce code dublication; ``AdamWUpdate`` could hold an instance of ``AdamUpdate`` and in the update step, we something like:

```c++
adamUpdate.Update(iterate, stepSize, gradient);
iterate -= weightDecay * iterate;
```

Let me know what you think

niteya-shah: Makes a lot of sense , simpler to understand and less code . 
Have made the changes to use this method.

zoq:
@niteya-shah For future commits it would be useful to be more descriptive; so that if someone reads the commit message it's somewhat possible to grasp what has changed.

niteya-shah:
@zoq will do . 
About the review , are there any more changes or should I start working on the tests ?

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * AdamR can optimize differentiable separable functions.  For more details, see
+ * the documentation on function types included with this distribution or on the
+ * ensmallen website.
+*/
+template<typename UpdateRule = AdamUpdate>
+class AdamRType
+{
+ public:
+  //Overloaded Constructor to Allow for min stepSize
```
zoq: Do you mind to comment on each parameter?

include/ensmallen_bits/adam/adamw_update.hpp
```diff
@@ -0,0 +1,114 @@
...
+  //! Modify the second moment coefficient.
+  double& Beta2() { return optimiser.Beta2(); }
+
+ private:
+  // The number of iterations.
+  double iteration;
+  // The weight decay rate.
+  double weightDecay;
+  //The AdamUpdate optimser
+  AdamUpdate optimiser;
```
zoq: To be consistent with the rest of the codebase, do you think, we should go with optimizer or update?

zoq:
Let's add the tests 👍

include/ensmallen_bits/adamr/adamr_impl.hpp
```diff
@@ -0,0 +1,103 @@
...
+{
+  // If a user changed the step size he hasn't update the step size of the
+  // cyclical decay instantiation, so we have to do it here.
+  if (optimizer.StepSize() != optimizer.DecayPolicy().StepSize())
+  {
+    optimizer.DecayPolicy().StepSize() = optimizer.StepSize();
+  }
+
+  optimizer.DecayPolicy().EpochBatches() = function.NumFunctions() /
+      double(optimizer.BatchSize());
```
niteya-shah: @zoq Can you please tell me what this  line does ?

zoq: It describes the decay value per number of function evaluations per batch. Here is an example, let say we like to train a neural network with 10 samples, and a batch size of 2, in this case ``function.NumFunctions()`` returns 10 and ``optimizer.BatchSize()`` returns 2.

niteya-shah: In Cyclical Decay , isnt the decay value set by function itself ? Why are we manually setting it ?

niteya-shah:
@zoq The tests are done, can you review please.
Also the Test is failing because of spsa randomly.

tests/adam_test.cpp
```diff
@@ -418,3 +418,51 @@ TEST_CASE("PadamLogisticRegressionTest", "[AdamTest]")
...
+
+TEST_CASE("SimpleAdamWTestFunction", "[AdamTest]")
+{
+  // Sometimes this test can fail randomly, so we allow it to run up to three
+  // times.
+  bool success = false;
+  for (size_t trial = 0; trial < 3; ++trial)
+  {
+    SGDTestFunction f;
+    AdamW optimizer(1e-2, 1, 0.9, 0.99, 0.25, 1e-8);
```
zoq: Do you have some insight how often the test fails? Maybe we should tweak the optimizer parameter?

tests/adam_test.cpp
```diff
@@ -418,3 +418,51 @@ TEST_CASE("PadamLogisticRegressionTest", "[AdamTest]")
       coordinates);
   REQUIRE(testAcc == Approx(100.0).epsilon(0.006)); // 0.6% error tolerance.
 }
+
+TEST_CASE("SimpleAdamWTestFunction", "[AdamTest]")
```
zoq: Should we add this one to the ``adamr_test.cpp`` file?

tests/adam_test.cpp
niteya-shah: Sorry about that , it seems that it used to failed because I parameterised it wrong. I will fix it.

tests/adam_test.cpp
niteya-shah: Adamr is for Test groups with Warm restarts , whereas AdamW is a Adam Update . I dont think we should add this to adamr.

niteya-shah:
Also I have a question , is there anyway we can make mlpacks constructors like Tensorflow ? Different Updates have different number of variable parameters and it is becoming hard to properly call them . So if we can define the Update Class before and then create a constructor , we much more modularity and functionality.

rcurtin:
> Also I have a question , is there anyway we can make mlpacks constructors like Tensorflow ?

Given that TensorFlow is an entirely different framework and language, probably not.

But if you take a look at the documentation you can easily modify each individual parameter like this:

```
Adam a;
a.StepSize() = 0.1;
a.BatchSize() = 128;
...
```

Now I did notice that `Adam` doesn't reflect the `UpdatePolicy()` and `DecayPolicy()` methods from `SGD`.  That could be improved, at least.

niteya-shah:
What I meant was something like
```
AdamWUpdate upd(1e-8, 0.9, 0.999, 0.0001); // Our AdamW Update Policy
CyclicalDecay decay(50, 2, 0.02, 0.01); // Our Decay Policy
Adam a(batchsize, maxIteration, upd, decay); // Our Optimiser ```

We could always add additional methods but something like this seems possible , although it will take time to implement for all Optimisers , but it should help with future work . 
Just an Idea

File: tests/adam_test.cpp (hunk 10 above)
Comment (zoq) [reply tests/adam_test.cpp]: Thanks for the clarification, makes sense.

**Comment by rcurtin** [comment]
Yeah, so if you are using `SGD` directly you can already do this:

```
// I'm "cheating" a little bit here, since this isn't the same as the Adam class.
SGD<AdamWUpdate, CyclicalDecay> s;
s.UpdatePolicy() = AdamWUpdate(1e-8, 0.9, 0.999, 0.0001);
s.DecayPolicy() = CyclicalDecay(50, 2, 0.02, 0.01);
```

The real problem like I said is that we're not reflecting `sgd.updatePolicy` and `sgd.decayPolicy` as `UpdatePolicy()` and `DecayPolicy()` in the `Adam` class.  Do you want to handle that here?  Alternately, I can open an issue for it (and handle it when I have a chance).
```

zoq:
```c++
SGD<AdamWUpdate, CyclicalDecay> a(...., upd, decay);
```
or

```c++
SGD<AdamWUpdate, CyclicalDecay> a(...., AdamWUpdate(...), CyclicalDecay(....));
```

works as well.

niteya-shah:
@zoq yeah something like that , feels a lot cleaner to me . 
@rcurtin I think you should open another issue , because if we want to maintain consistency then something like this will have to implemented for all optimisers and I don't think I can do that alone .

zoq:
What we mentioned above already works. What @rcurtin mend with the issue, is you can't do something like:

```c++
Adam<AdamWUpdate> a;
a.UpdatePolicy() = AdamWUpdate(1e-8, 0.9, 0.999, 0.0001);
```

niteya-shah:
Yeah , I just noticed that. Thanks . 
Ill add the UpdatePolicy and DecayPolicy methods

rcurtin:
@niteya-shah Thanks! :+1:  I think it should probably be only `Adam` that needs these changes, but I haven't checked with all the other optimizers.

niteya-shah:
@zoq I added the methods , can you please review?

include/ensmallen_bits/adam/adam.hpp
```diff
@@ -63,7 +64,7 @@ namespace ens {
  *
  * @tparam UpdateRule Adam optimizer update rule to be used.
  */
-template<typename UpdateRule = AdamUpdate>
+template<typename UpdateRule = AdamUpdate,typename DecayPolicyType = NoDecay>
```
zoq: Missing space after, ``,``, also can you add a comment for the ``DecayPolicyType`` template parameter.

include/ensmallen_bits/adam/adam_impl.hpp
```diff
@@ -37,7 +37,7 @@ AdamType<UpdateRule>::AdamType(
               tolerance,
               shuffle,
               UpdateRule(epsilon, beta1, beta2),
-              NoDecay(),
+              DecayPolicyType(),
```
zoq: Perhaps, we should add the option to pass an instantiation of the update and decay policy, see:

https://github.com/mlpack/ensmallen/blob/a83ff61cbd42d5ea9b86ebdbedff53d73740d33c/include/ensmallen_bits/sgd/sgd.hpp#L96-L97

include/ensmallen_bits/adam/adamw_update.hpp
```diff
@@ -0,0 +1,114 @@
...
+{
+ public:
+  /**
+   * Construct the Adam update policy with the given parameters.
+   *
+   * @param epsilon The epsilon value used to initialise the squared gradient
+   *        parameter.
+   * @param beta1 The smoothing parameter.
+   * @param beta2 The second moment coefficient.
+   * @param weightDecay the rate at which the Update regularises the iterant
```
zoq: The comment should start with a capital letter, and is missing a ``.``  at the end.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,165 @@
...
+ *   url     = {https://arxiv.org/pdf/1711.05101.pdf}
+ *   year    = {2019}
+ * }
+ * @endcode
+ *
+ * AdamR can optimize differentiable separable functions.  For more details, see
+ * the documentation on function types included with this distribution or on the
+ * ensmallen website.
+*/
+template<typename UpdateRule = AdamUpdate,typename DecayPolicyType = CyclicalDecay>
```
zoq: Can you comment on the template parameter?

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,165 @@
...
+   */
+  AdamRType(const double stepSize = 0.001,
+            const size_t epochRestart = 50,
+            const double multFactor = 2.0,
+            const size_t batchSize = 32,
+            const double beta1 = 0.9,
+            const double beta2 = 0.999,
+            const double eps = 1e-8,
+            const size_t maxIterations = 100000,
+            const double tolerance = 1e-5,
```
zoq: See comment above about passing an update or decay policy.

include/ensmallen_bits/sgd/update_policies/weight_decay_momentum_update.hpp
```diff
@@ -0,0 +1,101 @@
...
+  double momentum;
+  // The velocity matrix
+  arma::mat velocity;
+  // The weight decay (lambda) parameter
+  double weightDecay;
+};
+
+} // namespace ens
+
+#endif //ENSMALLEN_WEIGHT_DECAY_MOMENTUM_UPDATE_HPP
```
zoq: Missing space right after ``//``.

niteya-shah:
@zoq I have made the changes .

include/ensmallen_bits/adam/adam.hpp
```diff
@@ -99,6 +103,15 @@ class AdamType
            const bool shuffle = true,
            const bool resetPolicy = true);
 
+  AdamType(const UpdateRule& updateRule,
```
zoq: Instead of introducing a new constructor, we could also put ``updateRule`` and ``decayPolicy`` at the end of the first one and use the default value, let me know what you think.

niteya-shah: I dont think that will work , if you look at how the constructor works,
[https://github.com/mlpack/ensmallen/blob/a8f3022afba52bfc22d6423feba0492b29a198d5/include/ensmallen_bits/adam/adam_impl.hpp#L23-L42](url)
We explicitly pass parameters( epsilon , beta1 and beta2 ) into the Update Rule, if we where to replace it , it would pass those values to the Update , which is wrong because we are passing our own Update and Decay Policies, and if we put the policies in the beginning , then we to pass them to access other paramters and the default parameter values are still passed .
If there is something that I have missed then please correct me .

zoq: You are right, so perhaps it's not the best idea to provide a constructor to set the update rule, just to keep the interface simple. I could also use ``optimizer.UpdatePolicy() = policy;`` to get the same functionality. So I would remove the extra constructor, if you are fine with that, I think I suggested this in the first place.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,180 @@
...
+            const bool shuffle = true,
+            const bool resetPolicy = true);
+
+
+  AdamRType(const UpdateRule& updateRule,
```
zoq: Can we remove this one as well?

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,170 @@
...
+#define ENSMALLEN_ADAM_ADAMR_HPP
+
+#include <ensmallen_bits/adam/adam_update.hpp>
+#include <ensmallen_bits/sgdr/cyclical_decay.hpp>
+#include <ensmallen_bits/sgd/sgd.hpp>
+#include <ensmallen_bits/adam/adamw_update.hpp>
+
+namespace ens {
+/**
+ * This class is based on the Adam Update where the Optimiser
```
rcurtin: Minor comment---this should read `optimizer` not `Optimiser` to match the rest of the library.  And do you mean that this class is based on `AdamUpdate`?  Because that doesn't make sense.  I think you mean it is based on the `Adam` class.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,170 @@
...
+ * simulates a new warm-started run/restart once a number of epochs are
+ * performed.
+ *
+ * @code
+ * @article{
+ *   title   = {Decoupled Weight Decay Regularization},
+ *   author  = {{Ilya}, L. and {Frank}, H.},
```
rcurtin: `Ilya` and `Frank` are the first names; shouldn't this be `{Loschilov, I. and Hutter, F.}`?

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,170 @@
...
+   * modified to store the finishing point of the algorithm, and the final
+   * objective value is returned.
+   *
+   * @tparam DecomposableFunctionType Type of the function to optimize.
+   * @param function Function to optimize.
+   * @param iterate Starting point (will be modified).
+   * @return Objective value of the final point.
+   */
+  template<typename DecomposableFunctionType>
+  double Optimize(DecomposableFunctionType& function, arma::mat& iterate);
```
rcurtin: There should be an empty line below this `Optimize()` declaration and the next method.

include/ensmallen_bits/sgdr/cyclical_decay.hpp
```diff
@@ -131,6 +133,8 @@ class CyclicalDecay
 
   //! Locally-stored epoch.
   size_t epoch;
+  //Locally-stored min step size_t
```
rcurtin: Can you fix the style here please?  `//! Locally-stored minimum step size.`  The comment doesn't make very much sense as it is.

include/ensmallen_bits/sgdr/sgdr.hpp
```diff
@@ -61,7 +62,7 @@ class SGDR
    *
    * @param epochRestart Initial epoch where decay is applied.
    * @param batchSize Size of each mini-batch.
-   * @param stepSize Step size for each iteration.
+   * @param stepSizeMax Initial Step size for each iteration.
```
rcurtin: `Initial Step` -> `Initial step` (this may apply in other places).

include/ensmallen_bits/sgdr/sgdr.hpp
```diff
@@ -71,11 +72,12 @@ class SGDR
    *        parameters.
    * @param resetPolicy If true, parameters are reset before every Optimize
    *        call; otherwise, their values are retained.
+   * @param stepSizeMin Final step size before each restart
    */
-  SGDR(const size_t epochRestart = 50,
+  SGDR(const double stepSizeMax = 0.01,
+       const size_t epochRestart = 50,
```
rcurtin: Why has `stepSizeMax` suddenly become the first parameter?  And where is `stepSizeMin`?

include/ensmallen_bits/sgdr/sgdr_impl.hpp
```diff
@@ -19,10 +19,10 @@ namespace ens {
 
 template<typename UpdatePolicyType>
 SGDR<UpdatePolicyType>::SGDR(
+    const double stepSize,
```
rcurtin: Isn't this `stepSizeMax`?

include/ensmallen_bits/adam/adamw_update.hpp
```diff
@@ -0,0 +1,114 @@
...
+  double Beta2() const { return update.Beta2(); }
+  //! Modify the second moment coefficient.
+  double& Beta2() { return update.Beta2(); }
+
+ private:
+  // The number of iterations.
+  double iteration;
+  // The weight decay rate.
+  double weightDecay;
+  //The AdamUpdate optimser
```
rcurtin: Please correct this comment.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,170 @@
...
+ *   year    = {2019}
+ * }
+ * @endcode
+ *
+ * AdamR can optimize differentiable separable functions.  For more details, see
+ * the documentation on function types included with this distribution or on the
+ * ensmallen website.
+ * @tparam UpdateRule Adam optimizer update rule to be used.
+ * @tparam DecayPolicy The StepSize decay policy to be used by default
+ * CyclicalDecay.
```
rcurtin: Please change this to something more clear like `@tparam DecayPolicy The step size decay policy to be used; by default, CyclicalDecay is used.`.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,170 @@
...
+   * maximum number of iterations refers to the maximum number of points that
+   * are processed (i.e., one iteration equals one point; one iteration does not
+   * equal one pass over the dataset).
+   *
+   * @param stepSize Maximum and initial step size for each batch of warm
+                     restart.
+   * @param epochRestart Restart Rate for Warm Restarts
+   * @param multFactor Multiplier for epochRestart
+   * @param stepSizeMin Minimum and final step size for each batch of warm
+                        restart (Use DecayPolicy() method with CyclicalDecay).
```
rcurtin: `stepSizeMin` doesn't appear in the list below.  In addition, I would suggest that perhaps we put this right next to the `stepSize` parameter.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,170 @@
...
+  UpdateRule& UpdatePolicy() { return optimizer.UpdatePolicy(); }
+
+  //! Get the step size decay policy.
+  const DecayPolicyType& DecayPolicy() const { return optimizer.DecayPolicy(); }
+  //! Modify the step size decay policy.
+  DecayPolicyType& DecayPolicy() { return optimizer.DecayPolicy(); }
+
+ private:
+  size_t batchSize;
+  // The SGDR object with AdamR policy.
```
rcurtin: But this is not an SGDR object, it is an SGD object.  Should it be an SGDR object?  Then you should use an SGDR object directly.  Or clarify the comment.

include/ensmallen_bits/adamr/adamr_impl.hpp
```diff
@@ -0,0 +1,77 @@
...
+              UpdateRule(epsilon, beta1, beta2),
+              DecayPolicyType(
+                  epochRestart,
+                  multFactor,
+                  stepSize),
+              resetPolicy)
+{ /* Nothing to do. */ }
+
+// Overloaded Constructor to allow the user to pass their own Update
+// and DecayPolicy.
```
rcurtin: What does this comment refer to?

include/ensmallen_bits/sgd/update_policies/weight_decay_momentum_update.hpp
```diff
@@ -0,0 +1,101 @@
...
+ * \f]
+ *
+ * @code
+ * @article{
+ *   title   = {Decoupled Weight Decay Regularization},
+ *   author  = {{Ilya}, L. and {Frank}, H.},
+ *   journal = {ArXiv e-prints},
+ *   url     = {https://arxiv.org/pdf/1711.05101.pdf}
+ *   year    = {2019}
+ * }
```
rcurtin: Please fix this, and please add a comment above the `@code` block pointing out that this is the reference to the paper where this strategy is defined.

include/ensmallen_bits/sgd/update_policies/weight_decay_momentum_update.hpp
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * De-Coupled Weight Decay Policy for SGD with Momentum (SGDW).
+ *
+ * This Implments Decoupled Weight Decay Policy  in which the weight decay is
+ * decoupled from the optimization steps w.r.t. to the loss function.
+ * For SGD variants, this simplifies hyperparameter search since it decouples
+ * the settings of weight decay and learning rate.
+ *
+ *
+ * The Update Policy for SGDR
```
rcurtin: This is not a complete sentence.  `The update policy for SGDR is below:` would be better.

include/ensmallen_bits/adam/adam.hpp
```diff
@@ -62,8 +63,9 @@ namespace ens {
  * types included with this distribution or on the ensmallen website.
  *
  * @tparam UpdateRule Adam optimizer update rule to be used.
+ * @tparam DecayPolicy The StepSize decay policy to be used.
  */
-template<typename UpdateRule = AdamUpdate>
+template<typename UpdateRule = AdamUpdate, typename DecayPolicyType = NoDecay>
```
rcurtin: Why are we adding `DecayPolicyType` if it isn't ever set to anything other than `NoDecay`?  I can find no uses of that in the codebase.

include/ensmallen_bits/adam/adam.hpp
```diff
@@ -88,6 +90,8 @@ class AdamType
    *        function is visited in linear order.
    * @param resetPolicy If true, parameters are reset before every Optimize
    *        call; otherwise, their values are retained.
+   * @param updateRule Update Policy to be used for Adam.
+   * @param decayPolicy Decay Policy to be used for Adam.
```
rcurtin: Why have you added documentation but the method itself is unchanged?

include/ensmallen_bits/adam/adamw_update.hpp
```diff
@@ -0,0 +1,114 @@
...
+ * For more information, see the following.
+ *
+ * @code
+ * @article{Kingma2014,
+ *   author  = {Diederik P. Kingma and Jimmy Ba},
+ *   title   = {Adam: {A} Method for Stochastic Optimization},
+ *   journal = {CoRR},
+ *   year    = {2014},
+ *   url     = {http://arxiv.org/abs/1412.6980}
+ * }
```
rcurtin: This comment has no relevance to the code you are adding here.

include/ensmallen_bits/adam/adamw_update.hpp
```diff
@@ -0,0 +1,114 @@
...
+ * @endcode
+ */
+class AdamWUpdate
+{
+ public:
+  /**
+   * Construct the Adam update policy with the given parameters.
```
rcurtin: This is not the Adam update policy.

include/ensmallen_bits/adam/adamw_update.hpp
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * @param epsilon The epsilon value used to initialise the squared gradient
+   *        parameter.
+   * @param beta1 The smoothing parameter.
+   * @param beta2 The second moment coefficient.
+   * @param weightDecay The rate at which the Update regularises the iterant.
```
rcurtin: Why is `Update` capitalized?  And what is the `iterant`?

include/ensmallen_bits/adam/adam_impl.hpp
rcurtin: Since all the existing `UpdateRule`s take `epsilon`, `beta1`, and `beta2`, I think it's fine to leave it as is.  But the new `AdamW` update also has the `weightDecay` parameter.  Therefore I would actually suggest a new `AdamW` class that wraps `Adam<AdamWUpdate>` and allows a parameter for `weightDecay` in the constructor.

And if `DecayPolicyType` is not necessary at all (and I don't think it is) then we don't need to worry about that part.

include/ensmallen_bits/sgd/update_policies/weight_decay_momentum_update.hpp
```diff
@@ -0,0 +1,101 @@
...
+ */
+#ifndef ENSMALLEN_WEIGHT_DECAY_MOMENTUM_UPDATE_HPP
+#define ENSMALLEN_WEIGHT_DECAY_MOMENTUM_UPDATE_HPP
+
+namespace ens {
+
+/**
+ * De-Coupled Weight Decay Policy for SGD with Momentum (SGDW).
+ *
+ * This Implments Decoupled Weight Decay Policy  in which the weight decay is
```
rcurtin: Can you fix this comment please?  It's not clear why some of these words are capitalized.

include/ensmallen_bits/sgd/update_policies/weight_decay_momentum_update.hpp
```diff
@@ -0,0 +1,101 @@
...
+   *
+   * @param rows Number of rows in the gradient matrix.
+   * @param cols Number of columns in the gradient matrix.
+   */
+  void Initialize(const size_t rows, const size_t cols)
+  {
+    velocity = arma::zeros<arma::mat>(rows, cols);
+  }
+
+
```
rcurtin: There's only a need for one line of space here.

include/ensmallen_bits/sgd/update_policies/weight_decay_momentum_update.hpp
```diff
@@ -0,0 +1,101 @@
+/**
+ * @file weight_decay_momentum_update.hpp
```
rcurtin: Should this be named `decoupled_weight_decay_momentum_update.hpp`?

include/ensmallen_bits/sgdr/cyclical_decay.hpp
```diff
@@ -96,7 +87,18 @@ class CyclicalDecay
...
+    // Time to adjust the step size.
+    if (epoch < nextRestart)
+    {
+      // n_t = n_min^i + 0.5(n_max^i - n_min^i)(1 + cos(T_cur/T_i * pi)).
+      stepSize = stepSizeMin + 0.5 * constStepSize *
+                 (1 + cos((( double ) batchRestart / epochRestart)
+                 * arma::datum::pi));
+      // Keep track of the number of batches since the last restart.
+      batchRestart++;
+    }
```
rcurtin: This is a significant change to the functionality of `CyclicalDecay`.  If we are fixing a bug that is okay, but can you provide some justification please on what was wrong before, and what fixes it?  (justification in this comment is fine)

Also, the style is off here.  It should be `cos(((double) batchRestart / epochRestart)`.

tests/sgdr_test.cpp
```diff
@@ -39,12 +39,12 @@ TEST_CASE("SGDRCyclicalResetTest","[SGDRTest]")
       arma::Col<size_t> nextRestart(1000 / 10 /  mult);
       nextRestart(0) = restart;
       for (size_t j = 1; j < nextRestart.n_elem; ++j)
-        nextRestart(j) = nextRestart(j - 1) * mult;
+        nextRestart(j) = nextRestart(j - 1) + restart * std::pow(mult, j);
```
rcurtin: A justification here would also be nice.  Sorry for being picky but I am a little paranoid of changing tests that seemed to work before.

zoq:
Looks like I missed a bunch of important issues, thanks for the review.

include/ensmallen_bits/adam/adam.hpp
niteya-shah: So that if someone wants to change the decay policy , then its possible. A very small use case is 
```
  ens::AdamType<ens::AdamWUpdate> optim;
  optim.DecayPolicy() = ens::CyclicalDecay(50,2,0.03,0.01);
  optim.Optimize(f, coordinates);
```
this is will fail because the decay policy of optim is set to default NoDecay. The way I intend to fix this is by

```
  ens::AdamType<ens::AdamWUpdate, ens::CyclicalDecay> optim;
  optim.DecayPolicy() = ens::CyclicalDecay(50,2,0.03,0.01);
  optim.Optimize(f, coordinates);
```
This is how I added the methods to change Update and decay Policies , if there is a better method then please correct me.

include/ensmallen_bits/sgdr/cyclical_decay.hpp
niteya-shah: image
This is the update strategy as proposed in the paper .
1) The stepSizeMin parameter was added because , while in the original implmentation we assumed it to be 0 , the implmentation didnt really perform well as the learning rate slowed down too much near the ends of each epochBatch.
2) The step size change time , `epoch >= epochRestart` was changed because this is incorrect , we change every time before nextRestart , orginally this method changed it  for values of multiples of mult factor times eoochRestart , which isnt what the paper intended , we should do it until we reach next Restart .
From example of T0 = 10 and mult factor = 2, we changed for values less than 10, 20 ,40 ,80 and so on and not 10 ,30 ,70 etc.
The changes in the tests reflect this .

tests/sgdr_test.cpp
niteya-shah: This test is changed to match the fixes to the method , for example , as above , if T0 is 10 and mult is 2 then , next restart will will take values 10, 20, 40, 80 , not at all what the paper intented , the intented result should be 10, 30, 70, 150 .
My method , on the other hand calculates this , it takes the orginal result values, multiplies it by the power of mult , and then adds it . 
I hope this clarifies my changes .

niteya-shah:
@rcurtin about making an extra class wrapper for AdamW, while I dont have problem in making one , I feel that it wont really fix the problem of new optimisers having many more parameters , making more classes doesnt seem like the correct method when our existing architecture could indeed support the needs.  What I have been thinking of instead of each optimiser having to provide inputs to the default constructor , like 
```   Adam optimizer(1e-3, 1, 0.9, 0.999, 1e-8, 500000, 1e-9, true); ```

we change it to something like this which becomes our default constructor

```   Adam optimizer(1e-3, 1, AdamUpdate(1e-8, 0.9, 0.999), NoDecay(), 500000, 1e-9, true); ```

this way our code is much for functional and we can easily work with future optimisers without having to write a new class for every new optimiser. 
Please let me know what you think.

**Comment by niteya-shah** [comment]
@zoq can you review this , my exams are coming and I wont be able to make changes during that time. 
Also can you please tell me what to do about AdamW , should I make a new class or change the constructor.

File: include/ensmallen_bits/adam/adam.hpp (hunk 33 above)
Comment (rcurtin) [reply include/ensmallen_bits/adam/adam.hpp]: I see, thanks for the clarification.  But note that `AdamType<>` is actually just a wrapper around `SGD<>` that exposes some of the functions of the update policy and other bits specific to Adam.  So if you wanted to do

```
ens::AdamType<ens::AdamWUpdate, ens::CyclicalDecay>
```

then it would actually be better to do

```
ens::SGD<ens::AdamWUpdate, ens::CyclicalDecay>
```

instead.  Now, if you are saying that `ens::AdamType<ens::AdamWUpdate, ens::CyclicalDecay>` is a useful optimizer that we should expect people to regularly use, then we can make a typedef for it.  But the documentation as it currently stands mostly expects people to drop in types that represent optimizers and not play with template types.  You can see that the documentation at http://ensmallen.org/docs.html#standard-sgd says

```
Note that StandardSGD is based on the templated type SGD<UpdatePolicyType, DecayPolicyType> with UpdatePolicyType = VanillaUpdate and DecayPolicyType = NoDecay.
```

which is meant to make people aware that it is built on a template type, but it's also meant to keep the documentation as simple as possible.  "Advanced users" who want to play with different template types can take a look through the source and it *should* be straightforward to figure out how to use with different template types.

Anyway, this is probably more of an answer than you asked for, but I was hoping to also clarify some of the reasoning behind my comment. :)

File: include/ensmallen_bits/sgdr/cyclical_decay.hpp (hunk 41 above)
Comment (rcurtin) [reply include/ensmallen_bits/sgdr/cyclical_decay.hpp]: Thanks for the clarification.  I understand now and agree that you've fixed a bug. :+1:

File: tests/sgdr_test.cpp (hunk 42 above)
Comment (rcurtin) [reply tests/sgdr_test.cpp]: Thanks for the clarification.  The changes make more sense now.  But I disagree with the restart calculations.  The paper reads this:

```
In order to improve anytime performance, we suggest an option to start
with an initially small T_i and increase it by a factor of T_mult at every restart.
```

To me, that means that the number of epochs before restart should be 10, 20, 40, 80 when T_0 is 10 and mult is 2, not 10, 30, 70, 150.

Also, it seems to me like there has not been any change in `CyclicalDecay` to produce the results 10, 30, 70, 150; it multiplies by `multFactor` every time a restart is done.  So honestly I am a bit confused about why this test passes, and I wonder if something is wrong with the test.

**Comment by niteya-shah** [comment]
@rcurtin let me provide some more clarifications 
1) The restart calculation should be 10 , 30 , 70 , 150 and so on because this value is compared to with the total number of epochs that have occured , so if we consider for the values above , then the restart from 10 to 30 will get 20 epochs , similarly for 30 to 70 , 40 epochs .
2) For the changes to Cyclical decay , epochRestart, which holds the current number of epochs for this restart, is mutltiplied by mult factor and then is added to nextRestart. I hope this clarifies my changes.

I added the Decay Policy template to allow for simpler future updates , for other decay policies , but I now realise that its just better to build it on top of the SGD class. I will remove the extra template parameter.
 
Also for what should I do about the class wrapper for AdamW ?

**Comment by rcurtin** [comment]
@niteya-shah sorry for the slow response.  I re-read the tests and I agree, they are correct.  Thanks for the clarifications.

> Also for what should I do about the class wrapper for AdamW ?

I already left a comment about this:
https://github.com/mlpack/ensmallen/pull/79/files#r264453653
Let me know if I can clarify it.

**Comment by niteya-shah** [comment]
@rcurtin can you review this and the Quasi Hyperbolic momentum PR , I have made the changes and added the documentation.

File: include/ensmallen_bits/adamr/adamr.hpp [hunk 43]
```diff
@@ -0,0 +1,169 @@
...
+#define ENSMALLEN_ADAM_ADAMR_HPP
+
+#include <ensmallen_bits/adam/adam_update.hpp>
+#include <ensmallen_bits/sgdr/cyclical_decay.hpp>
+#include <ensmallen_bits/sgd/sgd.hpp>
+
+namespace ens {
+/**
+ * This class is based on the Adam class where the optimizer
+ * simulates a new warm-started run/restart once a number of epochs are
```
zoq: Can you clarify this point, not sure I get the point from the comment.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,169 @@
...
+ * @endcode
+ *
+ * AdamR can optimize differentiable separable functions.  For more details, see
+ * the documentation on function types included with this distribution or on the
+ * ensmallen website.
+ * @tparam UpdateRule Adam optimizer update rule to be used.
+ * @tparam DecayPolicy The step size decay policy to be used; by default,
+ * CyclicalDecay is used.
+*/
+template<typename UpdateRule = AdamUpdate, typename DecayPolicyType = CyclicalDecay>
```
zoq: Please keep the 80 charter line limit in mind.

zoq: Can you comment on each template parameter.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Construct the AdamR optimizer with the given function and parameters. The
+   * defaults here are not necessarily good for the given problem, so it is
+   * suggested that the values used be tailored to the task at hand.  The
+   * maximum number of iterations refers to the maximum number of points that
+   * are processed (i.e., one iteration equals one point; one iteration does not
+   * equal one pass over the dataset).
+   *
+   * @param stepSizeMax Maximum and initial step size for each batch of warm
+     restart.
```
zoq: Can you match the style with the one below.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,169 @@
...
+   * @param stepSizeMin Minimum and final step size for each batch of warm
+     restart.
+   * @param epochRestart Restart Rate for Warm Restarts
+   * @param multFactor Multiplier for epochRestart
+   * @param batchSize Number of points to process in a single step.
+   * @param beta1 Exponential decay rate for the first moment estimates.
+   * @param beta2 Exponential decay rate for the weighted infinity norm
+                  estimates.
+   * @param eps Value used to initialise the mean squared gradient parameter.
```
zoq: Do you mind to use epsilon here, just to improve the readability.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,169 @@
...
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *                      limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param shuffle If true, the function order is shuffled; otherwise, each
+   *        function is visited in linear order.
+   * @param resetPolicy If true, parameters are reset before every Optimize
+   *        call; otherwise, their values are retained.
+   * @param updateRule Update Policy to be used for AdamR.
```
zoq: There is no ``updateRule`` or ``decayPolicy`` parameter.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+  //! Get the step size decay policy.
+  const DecayPolicyType& DecayPolicy() const { return optimizer.DecayPolicy(); }
+  //! Modify the step size decay policy.
+  DecayPolicyType& DecayPolicy() { return optimizer.DecayPolicy(); }
+
+ private:
+  size_t batchSize;
+  // The SGD object with AdamR update policy and decay policy.
+  SGD<UpdateRule, DecayPolicyType> optimizer;
```
zoq: The comment is misleading, since there is no AdamR update policy.

include/ensmallen_bits/adamw/adamw.hpp
```diff
@@ -0,0 +1,156 @@
...
+ * @article{
+ *   title   = {Decoupled Weight Decay Regularization},
+ *   author  = {Loschilov, I. and Hutter, F.},
+ *   journal = {ArXiv e-prints},
+ *   url     = {https://arxiv.org/pdf/1711.05101.pdf}
+ *   year    = {2019}
+ * }
+ * @endcode
+ *
+ */
```
zoq: Can you add a reference to the function type documentation here as well?

include/ensmallen_bits/adamw/adamw.hpp
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * @article{
+ *   title   = {Decoupled Weight Decay Regularization},
+ *   author  = {Loschilov, I. and Hutter, F.},
+ *   journal = {ArXiv e-prints},
+ *   url     = {https://arxiv.org/pdf/1711.05101.pdf}
+ *   year    = {2019}
+ * }
+ * @endcode
+ *
```
zoq: We can remove the extra line here.

include/ensmallen_bits/adamw/adamw.hpp
```diff
@@ -0,0 +1,156 @@
...
+   * points that are processed (i.e., one iteration equals one point; one
+   * iteration does not equal one pass over the dataset).
+   *
+   * @param stepSize Step size for each iteration.
+   * @param batchSize Number of points to process in a single step.
+   * @param weightDecay The rate at which the update regularises the iterate.
+   * @param beta1 Exponential decay rate for the first moment estimates.
+   * @param beta2 Exponential decay rate for the weighted infinity norm
+            estimates.
+   * @param eps Value used to initialise the mean squared gradient parameter.
```
zoq: See comment above.

include/ensmallen_bits/adamw/adamw_impl.hpp
```diff
@@ -0,0 +1,43 @@
+/**
+ * @file adamw_impl.hpp
+ * @author Niteya Shah
+ *
+ * Implmentation of AdamW Update Policy
```
zoq: Isn't that the implementation of the ``AdamW`` optimizer?

include/ensmallen_bits/adamw/adamw_update.hpp
```diff
@@ -0,0 +1,126 @@
...
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_ADAM_ADAMW_UPDATE_HPP
+#define ENSMALLEN_ADAM_ADAMW_UPDATE_HPP
+
+namespace ens {
+
+  /**
+   * De-Coupled Weight Decay Policy for Adam (AdamW).
```
zoq: This one is off by one space.

include/ensmallen_bits/adamw/adamw_update.hpp
```diff
@@ -0,0 +1,126 @@
...
+  double& WeightDecay() { return weightDecay; }
+
+ private:
+  // The number of iterations.
+  double iteration;
+  // The weight decay rate.
+  double weightDecay;
+  // The AdamWUpdate optimser.
+  AdamUpdate update;
+
```
zoq: We can remove the extra line here.

include/ensmallen_bits/adamwr/adamwr.hpp
```diff
@@ -0,0 +1,155 @@
+/**
+ * @file adamwr.hpp
+ * @author Niteya Shah
+ * Declaration of the AdamWR optimizer.
```
zoq: There should be a new line after the author tag.

include/ensmallen_bits/adamwr/adamwr.hpp
```diff
@@ -0,0 +1,155 @@
...
+#ifndef ENSMALLEN_ADAM_ADAMWR_HPP
+#define ENSMALLEN_ADAM_ADAMWR_HPP
+
+#include <ensmallen_bits/adamw/adamw_update.hpp>
+#include <ensmallen_bits/sgdr/cyclical_decay.hpp>
+#include <ensmallen_bits/sgd/sgd.hpp>
+
+namespace ens {
+/*
+ *
```
zoq: This one is missing a description of the ``AdamWR`` method.

include/ensmallen_bits/adamwr/adamwr.hpp
```diff
@@ -0,0 +1,155 @@
...
+                  estimates.
+   * @param eps Value used to initialise the mean squared gradient parameter.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *                      limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param shuffle If true, the function order is shuffled; otherwise, each
+   *        function is visited in linear order.
+   * @param resetPolicy If true, parameters are reset before every Optimize
+   *        call; otherwise, their values are retained.
+   * @param updateRule Update Policy to be used for AdamR.
```
zoq: There is no ``updateRule`` or ``decayPolicy`` parameter.

include/ensmallen_bits/adamwr/adamwr.hpp
```diff
@@ -0,0 +1,155 @@
...
+   * @param decayPolicy Decay Policy to be used for AdamR.
+   */
+  AdamWR(const double stepSizeMax = 0.002,
```
zoq: Looks like we can just use the ``AdamRType`` class here instead of adding ``AdamWR``  and add another alias?

include/ensmallen_bits/sgd/update_policies/decoupled_weight_decay_momentum_update.hpp
```diff
@@ -0,0 +1,102 @@
+/**
+ * @file weight_decay_momentum_update.hpp
+ * @author Ajinkya Tejankar
+ * @author Niteya Shah
+ * Momentum update with weight decay for Stochastic Gradient Descent
```
zoq: Can you add a new line after the author tag.

include/ensmallen_bits/sgd/update_policies/decoupled_weight_decay_momentum_update.hpp
```diff
@@ -0,0 +1,102 @@
...
+*/
+class DecoupledWeightDecayMomentumUpdate
+{
+ public:
+  /**
+   * Construct the momentum update policy with weight decay. Initialize the
+   * momentum and weight decay to given values.
+   *
+   * @param momentum The momentum parameter
+   * @param weight_decay The weight decay parameter
```
zoq: Please use camel casing for all names.

include/ensmallen_bits/adamwr/adamwr.hpp
niteya-shah: The problem here is the same as with AdamW . AdamWR has an extra parameter of weight Decay and I dont want to overload the AdamR constructor . 
Please correct me if I misunderstood your doubt.

include/ensmallen_bits/adamr/adamr.hpp
niteya-shah: Do want me to comment on the AdamUpdate and Cyclical Decay , because the tparam is there.

niteya-shah:
@zoq I have made the changes , can review this ? thanks .

include/ensmallen_bits/adamwr/adamwr.hpp
zoq: Ahh, you are right I missed that.

include/ensmallen_bits/adam/adam.hpp
```diff
@@ -89,6 +89,7 @@ class AdamType
    * @param resetPolicy If true, parameters are reset before every Optimize
    *        call; otherwise, their values are retained.
    */
+
```
zoq: We can remove the extra line here.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,171 @@
...
+#include <ensmallen_bits/adam/adam_update.hpp>
+#include <ensmallen_bits/sgdr/cyclical_decay.hpp>
+#include <ensmallen_bits/sgd/sgd.hpp>
+
+namespace ens {
+/**
+ * This class is based on the Adam class where the optimizer
+ * simulates a new warm-started restart once the set number of epochs are
+ * performed. It modifies the learning rate at via the decay policy and once the
+ * set epoch number is reached , it resets the learning rate
```
zoq: We can remove the extra space before the comma.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,171 @@
...
+ *   journal = {ArXiv e-prints},
+ *   url     = {https://arxiv.org/pdf/1711.05101.pdf}
+ *   year    = {2019}
+ * }
+ * @endcode
+ *
+ * AdamR can optimize differentiable separable functions.  For more details, see
+ * the documentation on function types included with this distribution or on the
+ * ensmallen website.
+ * @tparam UpdateRule Adam optimizer update rule to be used.
```
zoq: Can you add an empty line right before the template parameter description.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,171 @@
...
+ * @tparam DecayPolicy The step size decay policy to be used; by default,
+ * CyclicalDecay is used.
```
zoq: The indentation should be 4 spaces here.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Construct the AdamR optimizer with the given function and parameters. The
+   * defaults here are not necessarily good for the given problem, so it is
+   * suggested that the values used be tailored to the task at hand.  The
+   * maximum number of iterations refers to the maximum number of points that
+   * are processed (i.e., one iteration equals one point; one iteration does not
+   * equal one pass over the dataset).
+   *
+   * @param stepSizeMax Maximum and initial step size for each batch of warm
+   * restart.
```
zoq: See comment above.

include/ensmallen_bits/adamr/adamr_impl.hpp
```diff
@@ -0,0 +1,76 @@
+/**
+ * @file adamr_impl.hpp
+ * @author Niteya Shah
+ * Implementation of the AdamR optimizer.
```
zoq: The should be an empty line right after the ``@author`` tag.

include/ensmallen_bits/adamw/adamw.hpp
```diff
@@ -0,0 +1,159 @@
+/**
+ * @file adamw.hpp
+ * @author Niteya Shah
+ *
```
zoq: The description is missing.

include/ensmallen_bits/adamw/adamw.hpp
```diff
@@ -0,0 +1,159 @@
...
+   * @param epsilon Value used to initialise the mean squared gradient parameter.
+   * @param maxIterations Maximum number of iterations allowed (0 means no
+   *        limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param shuffle If true, the function order is shuffled; otherwise, each
+   *        function is visited in linear order.
+   * @param resetPolicy If true, parameters are reset before every Optimize
+   *        call; otherwise, their values are retained.
+   */
+
```
zoq: We can remove the extra line here.

include/ensmallen_bits/adamwr/adamwr_impl.hpp
```diff
@@ -0,0 +1,75 @@
+/**
+ * @file adamr_impl.hpp
+ * @author Niteya Shah
+ * Implementation of the AdamR optimizer.
```
zoq: Can you add an empty line right after the ``@author`` tag.

include/ensmallen_bits/sgd/update_policies/decoupled_weight_decay_momentum_update.hpp
```diff
@@ -0,0 +1,103 @@
...
+ * @code
+ * @article{
+ *   title   = {Decoupled Weight Decay Regularization},
+ *   author  = {Loschilov, I. and Hutter, F.},
+ *   journal = {ArXiv e-prints},
+ *   url     = {https://arxiv.org/pdf/1711.05101.pdf}
+ *   year    = {2019}
+ * }
+ * @endcode
+ *
```
zoq: We can remove the extra line here.

niteya-shah:
@zoq I have made the changes

doc/optimizers.md
```diff
@@ -152,6 +152,170 @@ optimizer.Optimize(f, coordinates);
  * [Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980)
  * [Differentiable separable functions](#differentiable-separable-functions)
 
+## AdamR
+
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+AdamR is a modified version of Adam with warm restarts. This allows the optimiser to have a automated decreasing
+step size without modification. AdamR uses the Cyclical Decay decay policy.
```
zoq: Should we remove the second decay?

niteya-shah: I think we should replace it with "AdamR uses Cyclical Decay as its default decay policy."

zoq: Sounds good.

doc/optimizers.md
```diff
@@ -132,7 +132,7 @@ with _`UpdateRule`_` = AdamUpdate`.
 | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
 The attributes of the optimizer may also be modified via the member methods
-`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
+`StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`,
```
rcurtin: To be honest `UpdatePolicy()` is not a method that typical users playing with `Adam` should ever need to use, so I would say that we should remove this from the documentation.  Advanced users can take a look inside the code to see what it does more closely and it's well-documented so they shouldn't have a problem there.

doc/optimizers.md
```diff
@@ -152,6 +152,170 @@ optimizer.Optimize(f, coordinates);
...
+| `double` | **`beta2`** | Exponential decay rate for the weighted infinity norm estimates. | `0.999` |
+| `double` | **`eps`** | Value used to initialize the mean squared gradient parameter. | `1e-8` |
+| `size_t` | **`max_iterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+| `double` | **`tolerance`** | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+| `bool` | **`shuffle`** | If true, the function order is shuffled; otherwise, each function is visited in linear order. | `true` |
+| `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
+
+The attributes of the optimizer may also be modified via the member methods
+ `StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `UpdatePolicy()`, `DecayPolicy()`
+ `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
```
rcurtin: I know this is pedantic but I have always been careful to make sure the ordering of options here matches the ordering in the table above.  Can you make the same change here please?

doc/optimizers.md
```diff
@@ -152,6 +152,170 @@ optimizer.Optimize(f, coordinates);
...
+ optimizer.Optimize(f, coordinates);
+ ```
+
+#### See also:
+
+  * [SGDR: Stochastic Gradient Descent with Warm Restarts](https://arxiv.org/abs/1608.03983)
+  * [Decoupled Weight Decay Regularization](https://arxiv.org/abs/1711.05101)
+  * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
+  * [SGD](#standard-sgd)
+  * [Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980)
```
rcurtin: I'd say it's better to link to the Adam optimizer we have implemented, with `[Adam](#adam)`.

doc/optimizers.md
```diff
@@ -152,6 +152,170 @@ optimizer.Optimize(f, coordinates);
...
+  * [Differentiable separable functions](#differentiable-separable-functions)
+
+## AdamW
+
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+ AdamW is a modified version of Adam with decoupled weight decay.
```
rcurtin: It would be nice to add a sentence stating, e.g., why this might be a good technique to use or a little bit more about it.  This could help users decide which optimizer to use.  Perhaps even just an explanation of decoupled weight decay would be fine.

doc/optimizers.md
```diff
@@ -152,6 +152,170 @@ optimizer.Optimize(f, coordinates);
...
+ * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
+ * [SGD](#standard-sgd)
+ * [Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980)
+ * [Differentiable separable functions](#differentiable-separable-functions)
+
+## AdamWR
+
+ *An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+ AdamWR is a modified version of AdamW with warm restarts and is combination of AdamW and AdamR
```
rcurtin: There are a couple spaces on the start of the line here (and other places); they aren't necessary and we can remove them.  A little bit more description here could be helpful also.

doc/optimizers.md
```diff
@@ -152,6 +152,170 @@ optimizer.Optimize(f, coordinates);
...
+
+
```
rcurtin: The extra line here isn't needed.

doc/optimizers.md
```diff
@@ -152,6 +152,170 @@ optimizer.Optimize(f, coordinates);
...
+  | `double` | **`beta2`** | Exponential decay rate for the weighted infinity norm estimates. | `0.999` |
+  | `double` | **`eps`** | Value used to initialize the mean squared gradient parameter. | `1e-8` |
+  | `size_t` | **`max_iterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+  | `double` | **`tolerance`** | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+  | `bool` | **`shuffle`** | If true, the function order is shuffled; otherwise, each function is visited in linear order. | `true` |
+  | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
+
+  The attributes of the optimizer may also be modified via the member methods
+   `StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`, `WeightDecay()`,
+   `Tolerance()`, `Shuffle()`, and `ResetPolicy()`.
```
rcurtin: The options here are also out of order and some appear to be missing.

doc/optimizers.md
```diff
@@ -152,6 +152,170 @@ optimizer.Optimize(f, coordinates);
...
+  optimizer.Optimize(f, coordinates);
+  ```
+
+#### See also:
+
+ * [SGDR: Stochastic Gradient Descent with Warm Restarts](https://arxiv.org/abs/1608.03983)
+ * [Decoupled Weight Decay Regularization](https://arxiv.org/abs/1711.05101)
+ * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
+ * [SGD](#standard-sgd)
+ * [Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980)
```
rcurtin: Probably we should link to the `AdamW` and `AdamR` sections too.

include/ensmallen_bits/adamw/adamw_update.hpp
rcurtin: Now it's off by two spaces.

include/ensmallen_bits/sgd/update_policies/decoupled_weight_decay_momentum_update.hpp
```diff
@@ -0,0 +1,102 @@
...
+ */
+#ifndef ENSMALLEN_WEIGHT_DECAY_MOMENTUM_UPDATE_HPP
+#define ENSMALLEN_WEIGHT_DECAY_MOMENTUM_UPDATE_HPP
+
+namespace ens {
+
+/**
+ * De-Coupled Weight Decay Policy for SGD with Momentum (SGDW).
+ *
+ * This Implments the decoupled weight decay policy in which the weight decay is
```
rcurtin: `Implments` -> `implements`

include/ensmallen_bits/sgdr/sgdr.hpp
```diff
@@ -71,11 +72,13 @@ class SGDR
...
    * @param resetPolicy If true, parameters are reset before every Optimize
    *        call; otherwise, their values are retained.
+   * @param stepSizeMin Final step size before each restart
    */
   SGDR(const size_t epochRestart = 50,
        const double multFactor = 2.0,
        const size_t batchSize = 1000,
-       const double stepSize = 0.01,
+       const double stepSizeMax = 0.01,
+       const double stepSizeMin = 0.005,
```
rcurtin: So, unfortunately, this is a reverse-compatibility-breaking change.  We should add a deprecated constructor (don't worry about marking it as deprecated, we can do that in #113) that matches the old format if needed.

```
SGDR(const size_t epochRestart, const double multFactor, const size_t batchSize, const double stepSize,
    const size_t maxIterations, const double tolerance = 1e-5, const bool shuffle = true, 
    const UpdatePolicyType& updatePolicy = UpdatePolicyType(), const bool resetPolicy = true);
```

(We can't use default constructors before the `maxIterations` bit otherwise there will be ambiguity in the constructors.)

In this deprecated constructor we could just take `stepSizeMin = 0.5 * stepSizeMax` or something.

tests/adam_test.cpp
```diff
@@ -418,3 +418,44 @@ TEST_CASE("PadamLogisticRegressionTest", "[AdamTest]")
...
+{
+  // Sometimes this test can fail randomly, so we allow it to run up to three
+  // times.
```
rcurtin: This comment doesn't seem accurate.  We don't seem to be running this test three times.

tests/adamr_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+using namespace ens;
+using namespace ens::test;
+
+/*
+ * The test to check for cyclic consistency has been perfomed in SGDR test
+ *
+ * Run AdamR on logistic regression and make sure the results are acceptable.
+ */
+
```
rcurtin: This is an extra empty line, we can remove it.

tests/sgdr_test.cpp
rcurtin: We should address this comment before merge.  I still don't have a good understanding of what is going on here---please correct me if I've misunderstood anything.

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,173 @@
...
+   * @param iterate Starting point (will be modified).
+   * @return Objective value of the final point.
+   */
+  template<typename DecomposableFunctionType>
+  double Optimize(DecomposableFunctionType& function, arma::mat& iterate);
+
+  //! Get the step size.
+  double StepSize() const { return optimizer.StepSize(); }
+  //! Modify the step size.
+  double& StepSize() { return optimizer.StepSize(); }
```
rcurtin: (This applies to more optimizers than just this one.)

So, I see that we are allowing users to pass `stepSizeMax` and `stepSizeMin` into the constructor, but we only allow access later by `StepSize()`.  We should instead provide `StepSizeMax()` and `StepSizeMin()` functions---perhaps `StepSizeMin()` returns `optimizer.DecayPolicy().StepSizeMin()` or something?

include/ensmallen_bits/adamr/adamr.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+ private:
+  size_t batchSize;
+  //! The SGD object with update policy and decay policy.
+  SGD<UpdateRule, DecayPolicyType> optimizer;
+};
+
+using AdamR = AdamRType<AdamUpdate>;
+
+}
```
rcurtin: Also pedantic, but if this is closing a namespace we should add a comment `// namespace ens`.

niteya-shah:
@rcurtin I have made the changes but can you please check the deprecated constructor, I have just added a new constructor that followed the old format, and the conversion is stepSizeMin = stepSizeMax * 0.1 , this seemed good enough to me. with regards to the doubts you have with the cyclic decay, can you please tell me what they are because then maybe I may be able to give a better explanation.

Issue #80, opened 2019-02-09
CMAES sometimes gives runtime errors


favre49:
If you run CMAES with different seeds every time with the code:
`#include "ensmallen.hpp"
#include <armadillo>

int main()
{
	ens::test::RosenbrockFunction f;
	arma::mat coordinates = f.GetInitialPoint();
	arma::arma_rng::set_seed_random();
	ens::CMAES<> optimizer(0, -10, 10, 32, 200, -1);
	double value = optimizer.Optimize(f, coordinates);
	coordinates.print();
	std::cout << value;
}
`
and then executing, you sometimes get the error:

error: chol(): decomposition failed
terminate called after throwing an instance of 'std::runtime_error'
  what():  chol(): decomposition failed
Aborted (core dumped)

signifying that the cholesky decomposition in line 131 of cmaes_impl.cpp has failed.

rcurtin:
@favre49 thanks for the report.  I think this can happen when the observed covariance matrix isn't positive definite (which can happen).  There are many possible solutions, but one might be to detect when the Cholesky decomposition fails and then add some small value to the diagonal (or alternately perform an eigendecomposition, set any negative eigenvalues to 0, and reconstruct).  However, I have no idea how this will affect the convergence of the algorithm itself, so I think there are lots of possibilities for what we could do here.  If you have any better suggestions feel free to implement them.

PR #83, opened 2019-02-16
Simple fix to issue #80

favre49:
This is a simple fix to issue #80 . This is done by adding small values to the diagonal of the covariance matrix whenever the Cholesky decomposition fails to make the matrix positive definite. According to my testing, this seems to have fixed the problem.


doc/optimizers.md
```diff
@@ -506,6 +506,47 @@ optimizer.Optimize(f, coordinates);
...
+DE optimizer(200, 1000, 0.6, 0.8);
+optimizer.Optimize(f, coordinates);
+```
+
+#### See also:
+
+ * [Differential Evolution - A simple and efficient adaptive scheme for global optimization over continuous spaces](http://www1.icsi.berkeley.edu/~storn/TR-95-012.pdf)
+ * [Differential Evolution in Wikipedia](https://en.wikipedia.org/wiki/Differential_Evolution)
+ * [Arbitrary functions](#arbitrary-functions)
+
```
rcurtin: Oops, I think these changes should not be a part of this PR. :)

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -128,7 +128,20 @@ double CMAES<SelectionPolicyType>::Optimize(
...
+      covLower = arma::chol(C.slice(idx0), "lower");
+    }
+    catch (const std::runtime_error& err)
+    {
+      // If the covariance matrix isn't positive definite, add a small amount
+      // to the diagonal and try again
+      arma::mat I(iterate.n_elem, iterate.n_elem, arma::fill::eye);
+      C.slice(idx0) += I*(1e-16);
+      covLower = arma::chol(C.slice(idx0), "lower");
+    }
```
rcurtin: Actually you can simplify this a bit...

```
C.slice(idx0).diag() += 1e-16;
```

Also, if `1e-16` is generally sufficient I'm okay with that, but it's also possible to loop iteratively until the Cholesky decomposition succeeds.  You could use the version `chol(C.slice(idx0), covLower, "lower")` which returns a `bool`, and loop until it's true.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -281,4 +294,4 @@ double CMAES<SelectionPolicyType>::Optimize(
 
 } // namespace ens
 
-#endif
+#endif
```
rcurtin: No need to remove the newline from the end of the file, please. :+1:

include/ensmallen_bits/cmaes/cmaes_impl.hpp
rcurtin: Here is a simplification. :)

```
while (!arma::chol(covLower, C.slice(idx0), "lower"))
  C.slice(idx0).diag() += 1e-16;
```

favre49: Yup, far simpler, thank you for the suggestion

rcurtin:
@favre49 thanks! :+1:

PR #85, opened 2019-02-22
Introduce particle swarm optimization method

adl1995:
This implementation was previously added in mlpack and has now been refactored for ensmallen:
https://github.com/mlpack/mlpack/pull/1225

For the moment, the logistic regression test has been commented out as ensmallen doesn't seem to contain the gaussian distribution. Should this test be deleted, or is it possible to use this distribution from mlpack?

Also, is it possible to preserve the Git commits from the old repository?


zoq:
You could use git cherry pick, to get the commits, about the test, we converted the lr test, see: https://github.com/mlpack/ensmallen/blob/master/tests/adam_test.cpp#L166-L185 for an example.

adl1995:
Thanks, I have cherry-picked the old commits and updated the logistic regression test.

adl1995:
@rcurtin, @zoq - Can this PR be reviewed again? This was previously reviewed in: https://github.com/mlpack/mlpack/pull/1225

rcurtin:
@adl1995 can do, sorry that it fell off the list a bit.  I'll try and take a look when I have time, and mark this 'keep open'.

zoq:
@SuryodayBasak Do you think we could integrate some function from the PR, e.g. test suite? Would be nice to give @adl1995 credit for the code as well.

SuryodayBasak:
@zoq - yes, absolutely! I'm sorry for the delay with my response to this. We can get parts of this code into the project!

adl1995:
@SuryodayBasak - Great! Let me know if you need any help from my side.

birm:
It looks like there is some conflict due to #86 also being related to PSO. Is there any interest in resolving the merge conflict or otherwise keeping both PSO functionalities?

adl1995:
@birm - I can try to resolve the merge conflicts if this change still benefits mlpack.

PR #86, opened 2019-02-28
Introducing LBest PSO for Unconstrained Optimization Problems

SuryodayBasak:
I have included code that implements LBest PSO. This code is from the fork of mlpack owned by chintans111 on GitHub that I made compatible in ensmallen. I've included tests from his fork of mlpack as well.


include/ensmallen_bits/pso/pso.hpp
```diff
@@ -0,0 +1,138 @@
...
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_PSO_PSO_HPP
+#define ENSMALLEN_PSO_PSO_HPP
+
+#include "update_policies/lbest_update.hpp"
+#include "init_policies/default_init.hpp"
+#include "init_policies/constrained_init.hpp"
+#include <iostream>
```
zoq: ``include/ensmallen.hpp`` already includes ``<iostream>``.

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+namespace ens {
+
+/**
+ * EXTREMELY DETAILED DESCRIPTION OF THE WORKING OF PSO.
```
zoq: Would be great to have ``EXTREMELY DETAILED DESCRIPTION``.

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * For Particle Swarm Optimization to work, a FunctionType template parameter is
+ * required.
+ * This class must implement the following function:
+ *
+ *   double Evaluate(const arma::mat& coordinates);
+ *   void Gradient(const arma::mat& coordinates,
+ *                 arma::mat& gradient);
+ */
```
zoq: Would be great if you could add a reference to the function defenitions:
https://github.com/mlpack/ensmallen/blob/8d9682ca54483fb0601db8b48b17839e206149eb/include/ensmallen_bits/ftml/ftml.hpp#L41-L43

is just one example.

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -0,0 +1,138 @@
...
+   * Construct the particle swarm optimizer with the given function and
+   * parameters. The defaults here are not necessarily good for the given
+   * problem, so it is suggested that the values used be tailored to the task
+   * at hand.
+   *
+   * @param numParticles Number of particles in the swarm.
+   * @param maxIterations Number of iterations allowed.
+   * @param exploitationFactor Influence of the personal best of the particle.
+   * @param explorationFactor Influence of the neighbours of the particle.
+   * @param enableGradientDescent Enable the use of gradient descent optimizer.
```
zoq: Looks like the next three aren't used, and there is no description for ``velocityUpdatePolicy`` and ``initPolicy``.

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -0,0 +1,138 @@
...
+    const InitPolicy& initPolicy = InitPolicy()) :
+    numParticles(numParticles),
+    maxIterations(maxIterations),
+    exploitationFactor(exploitationFactor),
+    explorationFactor(explorationFactor),
+    velocityUpdatePolicy(velocityUpdatePolicy),
+    initPolicy(initPolicy) { /* Nothing to do */ }
+
+  /**
+   * ADD PSO OPTIMIZATION DESCRIPTION.
```
zoq: We should do that.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -0,0 +1,133 @@
...
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_PSO_UPDATE_POLICIES_LBEST_UPDATE_HPP
+#define ENSMALLEN_PSO_UPDATE_POLICIES_LBEST_UPDATE_HPP
+#include <assert.h>
+
+namespace ens {
+
+/**
+ * ADD SLIGHTLY MORE DETAILED DESCRIPTION HERE.
```
zoq: Agreed.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -0,0 +1,133 @@
...
+  {
+    // Set number of particles.
+    n = numParticles;
+    // Set c1 = exploitationFactor and c2 = explorationFactor.
+    c1 = exploitationFactor;
+    c2 = explorationFactor;
+
+    // Calculate the constriction factor
+    double phi = c1 + c2;
+    assert(phi > 4.0);
```
zoq: We should use ``static_assert`` in combination with the a description error message, instead of plain ``assert``.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -0,0 +1,133 @@
...
+   * @param numParticles The number of particles in the swarm.
+   * @param iterate The user input, used for shaping intermediate vectors.
+   */
+  void Initialize(const double& exploitationFactor,
+                  const double& explorationFactor,
+                  const size_t& numParticles,
+                  const arma::mat& iterate)
+  {
+    // Set number of particles.
+    n = numParticles;
```
zoq: Don't see a reason to make an alias for ``numParticles``, ``exploitationFactor`` and ``explorationFactor``. In fact, I think the other parameter names are much more descriptive.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+    // Calculate the constriction factor
+    double phi = c1 + c2;
+    assert(phi > 4.0);
+    chi = 2.0 / std::abs(2.0 - phi - std::sqrt((phi - 4.0) * phi));
+
+    // Initialize local best indices to self indices of particles.
+    size_t index = 0;
+    localBestIndices.set_size(n);
+    localBestIndices.imbue([&]() { return index++; });
```
zoq: We could use ``linspace`` to achive the same in a single line of code.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -0,0 +1,133 @@
...
+  void Update(const arma::cube& particlePositions,
+              arma::cube& particleVelocities,
+              const arma::cube& particleBestPositions,
+              const arma::vec& particleBestFitnesses)
+  {
+    // Velocity update logic.
+    for (size_t i = 0; i < n; i++)
+    {
+      localBestIndices(i) =
+        particleBestFitnesses(left(i)) < particleBestFitnesses(i) ?
```
zoq: Isn't left and right continues? So, I think we could use ``subvec`` here, to improve the performance; let me know what you think.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -0,0 +1,133 @@
...
+    assert(phi > 4.0);
+    chi = 2.0 / std::abs(2.0 - phi - std::sqrt((phi - 4.0) * phi));
+
+    // Initialize local best indices to self indices of particles.
+    size_t index = 0;
+    localBestIndices.set_size(n);
+    localBestIndices.imbue([&]() { return index++; });
+
+    // Set sizes r1 and r2.
+    r1.set_size(iterate.n_rows, iterate.n_cols);
```
zoq: I think we can remove the two lines here and use ``arma::randu(particleVelocities.n_rows, particleVelocities.n_cols``in the ``Update`` step.

include/ensmallen_bits/pso/init_policies/constrained_init.hpp
```diff
@@ -0,0 +1,126 @@
...
+    arma::mat positionVector;
+    double positionFitness;
+    double inf = std::numeric_limits<double>::max();
+    for (size_t i = 0; i < numParticles; i++)
+    {
+      bool positionFeasible = 0; //Flag for tracking if generated position vector is feasible or not
+      while (positionFeasible != 0)
+      {
+        //Generate random position vector
+        positionVector.randu(iterate.n_rows,iterate.n_cols);
```
zoq: We should generate the position around the starting point (``iterate``), so ``positionVector = arma::randu(iterate.n_rows,iterate.n_cols) + iterate``.

include/ensmallen_bits/pso/init_policies/constrained_init.hpp
```diff
@@ -0,0 +1,126 @@
...
+   * @param iterate Coordinates of the initial point for training.
+   * @param numParticles The number of particles in the swarm.
+   * @param particlePositions Current positions of particles.
+   * @param particleVelocities Current velocities of particles.
+   * @param particleFitnesses Current fitness values of particles.
+   * @param particleBestPositions Best positions attained by each particle.
+   * @param particleBestFitnesses Best fitness values attained by each particle.
+   */
+  template<typename FunctionType>
+  void InitializeParticles(FunctionType& function,
```
zoq: Do you mind to change the name to ``Initialize``, this would be shorter and since this is already in the pso folder the relation between particle is already there.

include/ensmallen_bits/pso/pso.hpp
SuryodayBasak: Thanks for pointing this out. I had used this to try to hack some outputs from the program.

include/ensmallen_bits/pso/pso.hpp
SuryodayBasak: I will do that, yes!

include/ensmallen_bits/pso/pso.hpp
SuryodayBasak: Could you please clarify this? What I'm seeing is that you want me to add references to ftml.hpp?

include/ensmallen_bits/pso/pso.hpp
SuryodayBasak: Yes, I'll remove these and as a part of the 'detailed description', I will add descriptions to each of the literals.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
SuryodayBasak: I'll take a look at this.

include/ensmallen_bits/pso/init_policies/constrained_init.hpp
SuryodayBasak: Okay, I'll do that.

include/ensmallen_bits/pso/init_policies/constrained_init.hpp
SuryodayBasak: I was thinking to incorporate the `min` and `max` values of the initial ranges along each axis into the particle initialization procedure. That will also be useful for initializing particles for constrained optimization.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
SuryodayBasak: I'll resolve this.

include/ensmallen_bits/pso/pso.hpp
zoq: Sorry for the confusion, I was talking about adding a reference to the function type documentation: 
```
For more details, see 
  * the documentation on function types include with this distribution or on the 
 * ensmallen website. 
```

include/ensmallen_bits/pso/init_policies/constrained_init.hpp
zoq: That works as well.

include/ensmallen_bits/pso/pso.hpp
SuryodayBasak: I've done this.

include/ensmallen_bits/pso/pso.hpp
SuryodayBasak: I've done this.

include/ensmallen_bits/pso/pso.hpp
SuryodayBasak: I have done this.

include/ensmallen_bits/pso/init_policies/default_init.hpp
```diff
@@ -0,0 +1,107 @@
...
+   * @param particlePositions Current positions of particles.
+   * @param particleVelocities Current velocities of particles.
+   * @param particleFitnesses Current fitness values of particles.
+   * @param particleBestPositions Best positions attained by each particle.
+   * @param particleBestFitnesses Best fitness values attained by each particle.
+   */
+  template<typename FunctionType>
+  void Initialize(FunctionType& function,
+		   const arma::mat& iterate,
+                   const size_t numParticles,
```
zoq: Picky style comment, the parameter are off by one space.

include/ensmallen_bits/pso/init_policies/feasible_init.hpp
```diff
@@ -0,0 +1,126 @@
...
+    particleVelocities.randu(iterate.n_rows, iterate.n_cols, numParticles);
+
+    // Initialize current fitness values to infinity.
+    particleFitnesses.set_size(numParticles);
+    particleFitnesses.fill(std::numeric_limits<double>::max());
+
+    // Copy to personal best values for first iteration.
+    particleBestPositions = particlePositions;
+    // Initialize personal best fitness values to infinity.
+    particleBestFitnesses.set_size(numParticles);
```
zoq: I guess if we like to parallelize the process at some point we need the fitness of each particle instead of just using an iterative process that saves the current best and worst fitness.

include/ensmallen_bits/pso/init_policies/feasible_init.hpp
```diff
@@ -0,0 +1,126 @@
...
+ public:
+  /**
+   * Constructor for the DefaultInit policy. The policy initializes particle
+   * posiitons in the range [lowerBound, upperBound]. Defaults to [-1, 1].
+   *
+   * @param lowerBound Lower bound of the position initialization range.
+   * @param upperBound Upper bound of the position initialization range.
+   */
+  ConstrainedInit(const double lowerBound = -1.0,
+              const double upperBound = 1.0):
```
zoq: See comment above.

include/ensmallen_bits/pso/init_policies/feasible_init.hpp
```diff
@@ -0,0 +1,126 @@
...
+   * @param numParticles The number of particles in the swarm.
+   * @param particlePositions Current positions of particles.
+   * @param particleVelocities Current velocities of particles.
+   * @param particleFitnesses Current fitness values of particles.
+   * @param particleBestPositions Best positions attained by each particle.
+   * @param particleBestFitnesses Best fitness values attained by each particle.
+   */
+  template<typename FunctionType>
+  void Initialize(FunctionType& function,
+			const arma::mat& iterate,
```
zoq: See comment above.

include/ensmallen_bits/pso/init_policies/feasible_init.hpp
```diff
@@ -0,0 +1,126 @@
+/**
+ * @file default_init.hpp
```
zoq: The filename isn't correct, the description is off as well.

include/ensmallen_bits/pso/init_policies/feasible_init.hpp
```diff
@@ -0,0 +1,126 @@
...
+			const arma::mat& iterate,
+                        const size_t numParticles,
+                        arma::cube& particlePositions,
+                        arma::cube& particleVelocities,
+                        arma::mat& particleFitnesses,
+                        arma::cube& particleBestPositions,
+                        arma::mat& particleBestFitnesses)
+  {
+    // Randomly initialize the particle positions.
+    particlePositions.randu(iterate.n_rows, iterate.n_cols, numParticles);
```
zoq: We could use the `Initialize` method from the `DefaultInit` class as basis and do the weight replacement afterwards.

include/ensmallen_bits/pso/init_policies/feasible_init.hpp
```diff
@@ -0,0 +1,126 @@
...
+      bool positionFeasible = 0; //Flag for tracking if generated position vector is feasible or not
+      while (positionFeasible == 0)
+      {
+        //Generate random position vector
+        positionVector.randu(iterate.n_rows,iterate.n_cols);
+        positionFitness = function.Evaluate(positionVector);
+        if (positionFitness != inf)
+        {
+          particlePositions.slice(i) = positionVector;
+          positionFeasible = 1;
```
zoq: A break in combination with `while(1)` could work as well, if we like to avoid the extra `positionFeasible` parameter.

include/ensmallen_bits/pso/init_policies/default_init.hpp
```diff
@@ -0,0 +1,107 @@
...
+                   arma::cube& particlePositions,
+                   arma::cube& particleVelocities,
+                   arma::mat& particleFitnesses,
+                   arma::cube& particleBestPositions,
+                   arma::mat& particleBestFitnesses)
+  {
+    // Randomly initialize the particle positions.
+    particlePositions.randu(iterate.n_rows, iterate.n_cols, numParticles);
```
zoq: The initial position should be around the user provided one (`iterate`), so something like: 
```cpp
particlePositions = arma::randu(iterate.n_rows, iterate.n_cols, numParticles);
particlePositions.each_slice() += iterate;
```

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -0,0 +1,144 @@
...
+   * at hand.
+   *
+   * @param numParticles Number of particles in the swarm.
+   * @param maxIterations Number of iterations allowed.
+   * @param exploitationFactor Influence of the personal best of the particle.
+   * @param explorationFactor Influence of the neighbours of the particle.
+   * @param VelocityUpdatePolicy Determines how the velocity of the particles are iteratively updated.
+   * @param InitPolicy Determines how the particles are initialized.
+   */
+  PSOType(
```
zoq: Picky style comment, can we start right after the `(`; move the parameters one line up.

SuryodayBasak:
@zoq - If you could look at my documentation style in `ensmallen_bits/pso/update_policies/lbest_update.hpp` and let me know what you think, it would be really helpful. I've done some extra bits and was just wondering if it is okay.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
SuryodayBasak: The reason the aliases are used this way is that the variable names `c1`, `c2`, etc. have these usual meanings in literature.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
SuryodayBasak: Using `static_assert` could make things a little complicated with `constexpr` functions and templates. If you want an appropriate error message, then we can do it with `assert`.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -20,12 +17,44 @@
...
- * REFERENCE TO A RELEVANT ARTICLE.
+ * For more information, refer the following:
  *
+ * @code
+ * @article{Poli2007,
+ *          doi = {10.1007/s11721-007-0002-0},
+ *          url = {https://doi.org/10.1007/s11721-007-0002-0},
+ *          year = {2007},
+ *          month = aug,
+ *          publisher = {Springer Science and Business Media {LLC}},
```
conradsnicta: That company is typically referred to as "Springer", not as "Springer Science and Business Media {LLC}".  
It's actually better to leave out the publisher field as it's not necessary.

SuryodayBasak: Thank you for pointing this out! I'll make the changes.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
zoq: True, probably a good idea to go with the common naming scheme.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -0,0 +1,163 @@
...
+   * the iteration process. It calculates the value of the constriction
+   * coefficent, initializes the local best indices of each particle to itself,
+   * and sets the shape of the r1 and r2 vectors.
+   *
+   * @param exploitationFactor Influence of personal best achieved.
+   * @param explorationFactor Influence of neighbouring particles.
+   * @param numParticles The number of particles in the swarm.
+   * @param iterate The user input, used for shaping intermediate vectors.
+   */
+  void Initialize(const double& exploitationFactor,
```
zoq: No need to use a reference for fundamental types like ``double`` or ``size_t``.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -0,0 +1,163 @@
...
+  {
+    // Copy values to aliases.
+    n = numParticles;
+    c1 = exploitationFactor;
+    c2 = explorationFactor;
+
+    // Calculate the constriction factor
+    static double phi = c1 + c2;
+    assert(phi > 4.0 && "The sum of the exploitation and exploration factors "
+                        "must be greater than 4.");
```
zoq: Indentation should be 4 spaces, if this is a new line.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -0,0 +1,163 @@
...
+        right(i) : i;
+    }
+
+    for (size_t i = 0; i < n; i++)
+    {
+      // Generate random numbers for current particle.
+      r1.randu();
+      r2.randu();
+      particleVelocities.slice(i) = chi * (particleVelocities.slice(i) +
+          c1 * r1 %
```
zoq: See comment about indentations above.

tests/pso_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+  REQUIRE(finalValue <= 1e-5);
+  for (size_t j = 0; j < 4; ++j)
+    //BOOST_REQUIRE_SMALL(coordinates[j], 1e-3);
+	REQUIRE(coords[j] <= 1e-3); //work on this
+}
+
+
+TEST_CASE("ConstrainedSquaredFunctionTest", "[PSOTest]")
+{
+  class SquaredFunction
```
zoq: Are you planning to switch to the other functions from: https://github.com/mlpack/ensmallen/pull/117?

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -0,0 +1,163 @@
...
+#define ENSMALLEN_PSO_UPDATE_POLICIES_LBEST_UPDATE_HPP
+#include <assert.h>
+
+namespace ens {
+
+/**
+ * The local best version (lbest) of PSO in which particles communicate with
+ * only two neighbours each, thus forming a ring topology amongst them. This
+ * approach allows PSO to converge at the global minimum, but takes
+ * significantly more iterations to do so.
```
zoq: This sounds/looks good to me.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
SuryodayBasak: But it is a part of the previous line of code. That's the reason I aligned it with the text. I'll change it if required.

tests/pso_test.cpp
SuryodayBasak: Yes, that's my plan for the coming week. I'll spend a few days working on the new test functions from PR#117. The merge was timely as I'm just beginning to make PSO run properly.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
SuryodayBasak: Okay! Thanks!

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
zoq: Missed the comment; I think we should follow the style as used here:

https://github.com/mlpack/mlpack/blob/2326bf8bd0b380c17eed6b5b1c34fbfab1563774/src/mlpack/tests/sfinae_test.cpp#L102-L108

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
SuryodayBasak: Yeah, left and right are continuous, but why would we need `subvec`, since we just have to look at the particles in positions `i-1` and `i+1`? The `left` and `right` functions are user defined helper functions that just helps us achieve this.

zoq: On a second look ``subvec`` isn't an option here, what I was thinking to use as something like:

```c++
localBestIndices.subvec(...) = particleBestFitnesses.subvec(..) < particleBestFitnesses.subvec(...);
```

But we don't have the result for each in a vector.

SuryodayBasak: Even if we do, I think we'll need the indexing, and thus considering subvec views will be redundant.

zoq: You are right, it does't provide any improvement.

include/ensmallen_bits/pso/init_policies/feasible_init.hpp
SuryodayBasak: When PSO is being run, there are essentially two loops. An _iteration_ in PSO would be an entire inner loop wherein the positions of all the particles are updated.

In order to stick to the original algorithm to the extent possible, what I want to do is implement parallelism on the updating _iteration_. You start a multi-thread or multi-process before iteration, and when the iteration completes, you wait for the multi-thread or multi-process to end. By doing so, we will be able to stick to the original algorithm without changes, but we will be able to make it faster by a constant amount of time.

SuryodayBasak: @zoq - does this sound alright? Would you be okay with this?

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
SuryodayBasak: Take a look at this one more time in the code, @zoq .

`r1` and `r2` are random vectors, and these need to change in every iteration.

In lines 127 and 128, we have `r1.randu();` and `r2.randu();`. Thus, in every update iteration, `r1` and `r2` will have different values, as is necessary in the PSO algorithm.

SuryodayBasak: Let me know what you think.

include/ensmallen_bits/pso/init_policies/feasible_init.hpp
zoq: Sounds good to me, I think the plan is to implement the standard method first, and afterwards implement parallelization.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
zoq: We just change the expression in the update step from:

```c++
r1.randu();
```

to

```
arma::randu(particleVelocities.n_rows, particleVelocities.n_cols);
```

it would be the same, just that we don't have need to set the size before. If you prefer the current solution that's fine with me as well.

SuryodayBasak: Okay we can do this.

tests/pso_test.cpp
```diff
@@ -0,0 +1,99 @@
...
+ * 3-clause BSD license along with mlpack.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#include <ensmallen.hpp>
+#include "catch.hpp"
+
+using namespace ens;
+using namespace ens::test;
+
+TEST_CASE("LBestPSOSphereFunctionTest", "[PSOTest]")
```
zoq: This looks really good, would be interesting to see if the current version is able to minimise: ``SchafferFunctionN2`` and  ``SchafferFunctionN4`` as well.

SuryodayBasak: Sure it does, but I have noted that the swarm is initialized either over the optimum or close to the optimum. It had kind of slipped my mind, but I need to tweak the API to incorporate user-selected lower-bounds and upper bounds. I don't want to make it too complicated, I'll initialize the population within a hypercube.

SuryodayBasak: Once I implement this, I'll deliberately initialize the swarm far from the optimum. If it succeeds then, we will know that it is fairly robust.

SuryodayBasak:
@zoq : the test for Schaffer N4 fails 2/10 times in my computer. At any rate, it is a massive improvement from CNE.

SuryodayBasak:
@zoq : I have changed the API to include `lowerBound` and `upperBound` as arguments to the initialization function of `DefaultInit`. That's one change that I would like you to closely examine and scrutinize.

I thought trying some template tricks to pass the bounds, but I think we will have to take a look at template specializations and that might make things more cumbersome in general. So I've removed the member variables in `DefaultInit` that corresponded to the bounds and I am passing them as arguments in the `Initialization` function of DefaultInit.

Another thing that you should notice is that I've changed the data type of `lowerBound` and `upperBound` from `const double` to `arma::vec`. This will give us better control and will allow us to set and upper bound and a lower bound for every dimension or variable in an objective.

Let me know your thoughts on these.

zoq:
> @zoq : the test for Schaffer N4 fails 2/10 times in my computer. At any rate, it is a massive improvement from CNE.

Great, I did some test with the CNE code and the main issue I have seen is that the optimization stops after a few iterations. Mainly because the best population remains the same over two runs in which case

```c++
if (std::abs(lastBestFitness - fitnessValues.min()) < tolerance)
```

is 0 and returns. I can see two solutions, either we set the tolerance so that the condition is always false (``tolerance < 0``) in which case we will run for the maximum number of iterations specified or we implement some form of patience parameter, in which case we wait for a couple of epochs before we check the difference.

zoq:
> I thought trying some template tricks to pass the bounds, but I think we will have to take a look at template specializations and that might make things more cumbersome in general. So I've removed the member variables in `DefaultInit` that corresponded to the bounds and I am passing them as arguments in the `Initialization` function of DefaultInit.

Sounds reasonable to me, the PSO constructor is pretty much the same, and I don't think a user has to pass the bounds to the policy at construction time.

> Another thing that you should notice is that I've changed the data type of `lowerBound` and `upperBound` from `const double` to `arma::vec`. This will give us better control and will allow us to set and upper bound and a lower bound for every dimension or variable in an objective.
> 
> Let me know your thoughts on these.

It might be useful to provide another constructor that takes the bounds as ``double``, internally we can create a vector of size 1 to store the value, that way we don't have to support two types. Let me know what you think.

SuryodayBasak:
> I can see two solutions, either we set the tolerance so that the condition is always false (`tolerance < 0`) in which case we will run for the maximum number of iterations specified or we implement some form of patience parameter, in which case we wait for a couple of epochs before we check the difference.

Thanks for the observation! I shouldn't have missed it myself. I would like to go with solution 2: store the best positions in a buffer and see if there is no improvement for a certain number of turns, whose value the user may define. That's the first thought I had when I saw that you pointed out this problem. Letting the method run till maximum iterations is not intelligent.

SuryodayBasak:
> It might be useful to provide another constructor that takes the bounds as `double`, internally we can create a vector of size 1 to store the value, that way we don't have to support two types. Let me know what you think.

I was adamant about not doing this but the more I think about it, the more it seems to me that this could be really useful. If there are many variables in the objective, then this kind of solution would desirable.

However, I'm not sure what you mean when you say:
>I don't think a user has to pass the bounds to the policy at construction time.
And then:
>It might be useful to provide another constructor that takes the bounds as `double`

It's just the implementation part that I'm a little confused about.

SuryodayBasak:
@zoq : Actually, implementing the above with multiple `PSOType` constructors will be pretty easy! The API will essentially remain unchanged from this point. However, we must now know that if we use `const double` types, we will be able to work only with hypercubes with each axis bounded by `lowerBound` and `upperBound`. This is a good option to have. With the recent API with the `arma::vec` lower and upper bounds, we will be able to create uniform hyperrectangles, which could be desirable in other cases.

If you think this is good, let me know and I'll make the changes.

zoq:
> @zoq : Actually, implementing the above with multiple `PSOType` constructors will be pretty easy! The API will essentially remain unchanged from this point. However, we must now know that if we use `const double` types, we will be able to work only with hypercubes with each axis bounded by `lowerBound` and `upperBound`. This is a good option to have. With the recent API with the `arma::vec` lower and upper bounds, we will be able to create uniform hyperrectangles, which could be desirable in other cases.
> 
> If you think this is good, let me know and I'll make the changes.

Agreed, as long as we can make this clear in the documentation I think this is fine.

SuryodayBasak:
Lastly, we should implement the following:

> implement some form of patience parameter, in which case we wait for a couple of epochs before we check the difference.

Please allow me some time to study existing methods to accomplish this.

zoq:
> Lastly, we should implement the following:
> 
> > implement some form of patience parameter, in which case we wait for a couple of epochs before we check the difference.

Agreed, for now I used -1 as tolerance (https://github.com/mlpack/ensmallen/pull/119/commits/2eda0cb86ba661e7072c27795581a89da17209f6) and much lower values for the iterations parameter and population size.

> Please allow me some time to study existing methods to accomplish this.

Sure, let me know if you need any help.

SuryodayBasak:
Okay, I've done the needful. I think this is a fairly stable implementation. In the latest push, I've implemented the lookback-horizon using `std::queue`. @zoq : thanks for the suggestion!

Commenting throughout the code is left to be done. I'll start working on that. Besides anything regarding the commenting, let me know if you would like anything more implemented in this version.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
SuryodayBasak: I think that if we follow the previous solution, the code will not have to reinitialize the variables `r1` and `r2`, but would only have to fill them with new random numbers. That's why I prefer the previous one.

include/ensmallen_bits/pso/pso_impl.hpp
```diff
@@ -0,0 +1,178 @@
...
+    particleFitnesses(i) = f.Evaluate(particlePositions.slice(i));
+    particleBestFitnesses(i) = particleFitnesses(i);
+  }
+  
+  // Declare queue to keep track of improvements over a number of iterations.
+  queue <double> performanceHorizon;
+  // Variable to store the position of the best particle.
+  size_t bestParticle = 0;
+  //Find the best fitness.
+  double bestFitness = arma::datum::inf;
```
zoq: In the init class you used ``std::numeric_limits<double>::max()``, to be consistent we should probably use the same here as well.

include/ensmallen_bits/pso/pso_impl.hpp
```diff
@@ -0,0 +1,178 @@
...
+  {
+    // Calculate fitness and evaluate personal best.
+    for (size_t j = 0; j < numParticles; j++)
+    {
+      particleFitnesses(j) = f.Evaluate(particlePositions.slice(j));
+      // Compare and copy fitness and position to particle best.
+      if (particleFitnesses(j) < particleBestFitnesses(j))
+      {
+        particleBestFitnesses(j) = particleFitnesses(j);
+        particleBestPositions.slice(j) = particlePositions.slice(j);
```
zoq: I'm wondering if it makes sense to assign the best position after the iteration, just to keep the number of copies low, another solution might be to use an advanced constructor, and create an alias:

```cpp
particleBestPositions.slice(j) = arma::Mat<double>(particlePositions.slice(j).memptr(), particlePositions.slice(j).n_rows, particlePositions.slice(j).n_cols, false, false);
```

http://arma.sourceforge.net/docs.html#Mat

include/ensmallen_bits/pso/pso_impl.hpp
```diff
@@ -0,0 +1,178 @@
...
+    for (size_t j = 0; j < numParticles; j++)
+    {
+      if (particleBestFitnesses(j) < bestFitness)
+      {
+        bestParticle = j;
+        bestFitness = particleBestFitnesses(bestParticle);
+      }
+    }
+
+    //Append bestFitness to performanceHorizon.
```
zoq: Missing a space right after ``//``.

include/ensmallen_bits/pso/pso_impl.hpp
```diff
@@ -0,0 +1,178 @@
...
+    performanceHorizon.push(bestFitness);
+
```
zoq: We can remove the empty line here.

include/ensmallen_bits/pso/pso_impl.hpp
```diff
@@ -0,0 +1,178 @@
...
+    
+    // Calculate fitness and evaluate personal best.
+    for (size_t j = 0; j < numParticles; j++)
+    {
+      particleFitnesses(j) = f.Evaluate(particlePositions.slice(j));
+      // Compare and copy fitness and position to particle best.
+      if (particleFitnesses(j) < particleBestFitnesses(j))
+      {
+        particleBestFitnesses(j) = particleFitnesses(j);
+        particleBestPositions.slice(j) = particlePositions.slice(j);
```
zoq: See comment above.

tests/pso_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+/**
+ * Test the PSO optimizer on the Rosenbrock Function.
+ */
+
+TEST_CASE("LBestPSORosenbrockTest","[PSOTest]")
+{
+  RosenbrockFunction f;
+
+  //Setting bounds for the initial swarm population.
```
zoq: Missing space right after ``//``.

tests/pso_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+  arma::vec coords = f.GetInitialPoint();
+  if (!s.Optimize(f, coords))
+    FAIL("LBest PSO optimization reported failure for Sphere Function.");
+  
+  double finalValue = f.Evaluate(coords);
+  //BOOST_REQUIRE_SMALL(result, 1e-5);
+  REQUIRE(finalValue <= 1e-5);
+  for (size_t j = 0; j < 4; ++j)
+    //BOOST_REQUIRE_SMALL(coordinates[j], 1e-3);
```
zoq: Looks like we can remove this one.

tests/pso_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+{
+  SphereFunction f(4);
+  LBestPSO s;
+
+  arma::vec coords = f.GetInitialPoint();
+  if (!s.Optimize(f, coords))
+    FAIL("LBest PSO optimization reported failure for Sphere Function.");
+  
+  double finalValue = f.Evaluate(coords);
+  //BOOST_REQUIRE_SMALL(result, 1e-5);
```
zoq: See comment below.

tests/pso_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+/**
+ * Test the PSO optimizer on the Rosenbrock function with lowerBound and
+ * upperbound of type double.
+ */
+
+TEST_CASE("LBestPSORosenbrockDoubleTest","[PSOTest]")
+{
+  RosenbrockFunction f;
+
+  //Setting bounds for the initial swarm population.
```
zoq: Missing space here as well, same for all the tests below.

include/ensmallen_bits/pso/pso_impl.hpp
SuryodayBasak: I've changed it in the init class, instead.

zoq: Yeah, that works as well.

zoq: On a second thought, we should go with ``std::numeric_limits<double>::max()``, in https://github.com/mlpack/ensmallen/pull/113 Ryan templatized each optimiser so that it not only works with ``arma::mat``, which means we would have to use ``std::numeric_limits<ElemType>::max()`` later anyway.

include/ensmallen_bits/pso/init_policies/default_init.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+    //Check if lowerBound is equal to upperBound. If equal, reinitialize.
+    arma::umat lbEquality = (lowerBound == upperBound);
+    if (lbEquality.n_rows == 1 && lbEquality(0, 0) == 1)
+    {
+      lowerBound = arma::ones<arma::vec>(iterate.n_rows);
+      lowerBound = -lowerBound;
```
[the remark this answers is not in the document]
SuryodayBasak: Yeah! Thanks for the pointer!

include/ensmallen_bits/pso/pso_impl.hpp
SuryodayBasak: Okay! Done!

include/ensmallen_bits/pso/pso_impl.hpp
SuryodayBasak: If we assign the best position after the iteration is over, we will have to keep a track of which particles' positions should be changed and which particles' positions should not be changed. The particles do not have a memory of their previous positions, obviously, as that would take more memory. All that a particle needs to record is the best position that is has visited and if it's most recent position is the best position, then the condition in line 94 is satisfied and the best position is changed.

That is the reason the change happens this way. And thus, there won't really be multiple copies since the `particleBestPositions` cube may not necessarily be updated after an iteration, and that the `particlePositions` cube will be updated after each iteration (but not necessarily to a more optimal position).

I think that if we use a memory pointer to a `particlePositions` we will just have two copies, and after every update of `particlePositions`, the particle best positions will be changed, too.

Let me know if this is correct or if I may have misunderstood.

zoq: Thanks for the clarification, I see what you mean, at first I thought we iterate over the same particles, which isn't the case, so I don't think we can avoid the copy here.

> I think that if we use a memory pointer to a particlePositions we will just have two copies, and after every update of particlePositions, the particle best positions will be changed, too.

That's correct.

tests/pso_test.cpp
```diff
@@ -0,0 +1,281 @@
...
+  SchafferFunctionN4 f;
+ 
+  // Setting bounds for the initial swarm population.
+  arma::vec lowerBound(2);
+  arma::vec upperBound(2);
+  lowerBound.fill(40);
+  upperBound.fill(50);
+  
+  LBestPSO s(15000, lowerBound, upperBound, 3000, 200, 1e-20, 1.5, 3.0);
+  arma::mat coordinates = arma::mat("0; 10");
```
zoq: Perhaps it makes sense to use an initial starting point that is closer to the solution.

include/ensmallen_bits/pso/pso_impl.hpp
SuryodayBasak: Thank you! Then I will mark this as resolved.

zoq: Good idea.

tests/pso_test.cpp
SuryodayBasak: That's the one thing I did not want to do. I wanted to test PSO for its robustness with non-convex problems. As a dynamical system, I'm not sure if selecting a closer initial point in a problem this hard will make much of a difference. Regardless, let me try this out. If it works, then we'll have a new observation. Also, let me try to initialize the solutions over a larger area that encompasses the solution. That's one thing that we haven't tried.

SuryodayBasak: I've been trying for the past two hours with many different values. Can't seem to land a 100% success rate. If you're okay with it, may we exclude it for the time being?

zoq: Sure, let's comment the test for now.

zoq:
Before we can merge this, can you update https://github.com/mlpack/ensmallen/blob/master/HISTORY.md and add PSO to https://github.com/mlpack/ensmallen/blob/master/doc/optimizers.md.

SuryodayBasak:
Yes, I will do that gladly! :)

zoq:
Looks like there is a swap file ("include/ensmallen_bits/pso/.pso.hpp.swp") that we should remove from the PR.

doc/optimizers.md
```diff
@@ -1336,6 +1336,86 @@ optimizer.Optimize(f, coordinates);
...
+At present, only the local-best variant of PSO is present in ensmallen. The optimizer may be initialized using the class type `LBestPSO`.
+
+#### Examples:
+
+```c++
+SphereFunction f(4);
+arma::vec coordinates = f.GetInitialPoint();
+
+LBestPSO s;
+const double result = s.Optimize(f, coords)
```
zoq: We should use ``coordinates`` here; can you check the other examples as well?

doc/optimizers.md
```diff
@@ -1336,6 +1336,86 @@ optimizer.Optimize(f, coordinates);
...
+```
+
+```c++
+RosenbrockFunction f;
+arma::vec coordinates = f.GetInitialPoint();
+
+// Setting bounds for the initial swarm population of size 2.
+arma::vec lowerBound(2);
```
zoq: Hm, I wonder if writing this as `` arma::vec lowerBound("50 50");`` might be more clear. Let me know what you think.

doc/optimizers.md
SuryodayBasak: Yes, I'll fix this.

doc/optimizers.md
SuryodayBasak: Certainly!

zoq:
This looks good, will run multiple test iterations, to make sure the pass rate is reasonable.

SuryodayBasak:
Okay. I have already tested it multiple times, but take your time to test it!!

zoq:
Always good to test it on multiple machines, with different seeds each time. But haven't seen a single failure in 100+ runs.

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -0,0 +1,242 @@
...
+   * @param lowerBound Lower bound of the coordinates of the initial population.
+   * @param upperBound Upper bound of the coordinates of the initial population.
+   * @param maxIterations Number of iterations allowed.
+   * @param horizonSize Size of the lookback-horizon for computing improvement.
+   * @param impTolerance Improvement threshold for termination.
+   * @param exploitationFactor Influence of the personal best of the particle.
+   * @param explorationFactor Influence of the neighbours of the particle.
+   */
+  PSOType(const size_t numParticles = 64,
+          arma::vec lowerBound = arma::ones<arma::vec>(1),
```
zoq: Just noted, we should probably use ``arma::vec& lowerBound`` same for the ``upperBound`` parameter, sorry missed this.

SuryodayBasak: Yeah, that makes sense. I'm sorry for missing this, too.

SuryodayBasak: There's one problem - if we use `arma::vec& lowerBound`, then we cannot use a default value for `lowerBound` and `upperBound`.

zoq: ``const arma::vec& lowerBound = arma::ones<arma::vec>(1)`` should work.

SuryodayBasak: Thanks for the pointers. I've done this!

zoq:
Thanks for a great contribution, excited to use this in practice.

PR #87, opened 2019-02-28
Adjust SPSA Logistic Regression test optimizer parameter

zoq:
Adjust SPSA Logistic Regression test optimizer parameter.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,7 @@
+### ensmallen 1.14.1
+###### ????-??-??
+  * Fixes for SPSA (#87).
+
 ### ensmallen 1.14.0
 ###### 2019-02-20
   * Add DE optimizer (#77).
--- a/doc/function_types.md
+++ b/doc/function_types.md
@@ -24,6 +24,7 @@ The following optimizers can be used to optimize an arbitrary function:
  - [Simulated Annealing](#simulated-annealing-sa)
  - [CNE](#cne)
  - [DE](#de)
+ - [SPSA](#simultaneous-perturbation-stochastic-approximation-spsa)
 
 Each of these optimizers has an `Optimize()` function that is called as
 `Optimize(f, x)` where `f` is the function to be optimized (which implements
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -1510,30 +1510,28 @@ optimizer.Optimize(f, coordinates);
 
 ## Simultaneous Perturbation Stochastic Approximation (SPSA)
 
-*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+*An optimizer for [arbitrary functions](#arbitrary-functions).*
 
 The SPSA algorithm approximates the gradient of the function by finite
 differences along stochastic directions.
 
-### Constructors
+#### Constructors
 
- * `SPSA(`_`alpha, batchSize, gamma, stepSize, evaluationStepSize, maxIterations, tolerance, shuffle`_`)`
+ * `SPSA(`_`alpha, gamma, stepSize, evaluationStepSize, maxIterations, tolerance, shuffle`_`)`
 
- #### Attributes
+#### Attributes
 
 | **type** | **name** | **description** | **default** |
 |----------|----------|-----------------|-------------|
 | `double` | **`alpha`** | Scaling exponent for the step size. | `0.602` |
-| `size_t` | **`batchSize`** | Batch size to use for each step. | `32` |
 | `double` | **`gamma`** | Scaling exponent for evaluation step size. | `0.101` |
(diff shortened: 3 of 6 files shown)


rcurtin:
I found that this helps, but there are still occasional failures.  I had better success by running the test three times, like we do for some tests, and then making sure it succeeded at least once.  Here's the code I used:

```
TEST_CASE("SPSALogisticRegressionTest", "[SPSATest]")
{
  bool success = false;
  arma::Row<size_t> responses, testResponses, shuffledResponses;

  for (size_t trial = 0; trial < 3; ++trial)
  {
    LogisticRegressionTestData(data, testData, shuffledData,
    LogisticRegression<> lr(shuffledData, shuffledResponses, 0.5);

    SPSA optimizer(0.5, 1, 0.102, 0.16, 0.3, 1000, 1e-7);
    arma::mat coordinates = lr.GetInitialPoint();
    optimizer.Optimize(lr, coordinates);

    // Ensure that the error is close to zero.
    const double acc = lr.ComputeAccuracy(data, responses, coordinates);
    const double testAcc = lr.ComputeAccuracy(testData, testResponses,
        coordinates);
    if (acc == Approx(100.0).epsilon(0.003) &&
        testAcc == Approx(100.0).epsilon(0.006))
      success = true;
    
    if (success)
      break;
  }

  REQUIRE(success == true);
}
```

Note that I also found that each run of `Optimize()` took surprisingly long---so I reduced the number of iterations to only 1000.  I think that maybe sometimes SPSA will go a bad direction and doesn't come back, so restarts are a solution.

However, I then dug into the SPSA implementation and found some other issues that I'm happy to fix but I want to get some confirmation on:

 1. SPSA is not meant for differentiable separable functions like the documentation says.  In fact it is for arbitrary functions, as it only calls `Evaluate()`.
 2. SPSA isn't originally defined as something that works on separable functions, but the extension is straightforward; simply call `Evaluate()` on a small batch of points instead of all the points.  Should we implement that extension?  I am okay with that if we can make it clear to users via, perhaps, two separate classes/typedefs like `SPSA` (full batch) and `BatchSPSA` (or some similar name).
 3. I'm confused by this block inside `SPSA::Optimize()`:

```
    gradient.zeros();
    for (size_t b = 0; b < batchSize; b++)
    { 
      // Stochastic directions.
      spVector = arma::conv_to<arma::mat>::from(
          arma::randi(iterate.n_rows, iterate.n_cols,
          arma::distr_param(0, 1))) * 2 - 1;
      
      iterate += ck * spVector;
      const double fPlus = function.Evaluate(iterate, 0, iterate.n_elem);
      
      iterate -= 2 * ck * spVector;
      const double fMinus = function.Evaluate(iterate, 0, iterate.n_elem);                                                                                                       
      iterate += ck * spVector;

      gradient += (fPlus - fMinus) * (1 / (2 * ck * spVector));
    }

    gradient /= (double) batchSize;
```

That should be the approximation of the gradient.  However, we're simply looping `batchSize` times to approximate the gradient, and we're also calling `Evaluate()` very strangely, with a strange batch size of `iterate.n_elem`.  It seems to me that:

 * if we are doing full-batch SPSA (as defined earlier) that we should be computing `gradient` as:

```
    spVector = arma::conv_to<arma::mat>::from(
        arma::randi(iterate.n_rows, iterate.n_cols,
        arma::distr_param(0, 1))) * 2 - 1;

    iterate += ck * spVector;
    const double fPlus = function.Evaluate(iterate);
      
    iterate -= 2 * ck * spVector;
    const double fMinus = function.Evaluate(iterate);                                                                                                       
    iterate += ck * spVector;

    gradient = (fPlus - fMinus) * (1 / (2 * ck * spVector));
```

A key point there is that there is no loop over the batch size.

  * If we are doing small-batch SPSA, then we should call `function.Evaluate(iterate, 0, batchSize)` instead.

Let me know what you think.  Like I said I'm happy to make the changes, but I want to double-check that you agree first.

zoq:
Interesting, I Interesting, I tested the optimizer like 500 times with a single error (using a new random seed each time), but if you still see errors I'm fine to run the test multiple times, I guess since you adjusted the number of iterations it's still fast enough. About SPSA/BatchSPSA; I agree, switching to ``function.Evaluate(iterate)`` makes a lot more sense. I think at this point there is no need to provide a batch variant unless you see one. Do you like to open a PR?

zoq:
We should definitely switch to ``function.Evaluate(iterate)``, and if it runs faster I guess it makes sense to adjust that one as well. I can do this inside of this PR or we can open a new one, but I don't think we should merge this one in. Let me know what you think.

rcurtin:
I'll work up a patch and get it to you shortly, and we can apply all the changes in this PR. :+1:

rcurtin:
Here's a patch: https://gist.github.com/a1deacc1debefbcd90b9d280364469a0

Issue #70, opened 2019-01-19
CMA-ES inconsistencies


rcurtin:
This is a continued version of mlpack/mlpack#1475, ported to the ensmallen repository because this is a more appropriate place.  This was originally opened by @FloopCZ with the following text:

> Hi everyone, first of all, thank you for the huge amount of work you have done on this project.
> After a brief pass through the CMA-ES implementation, I found a few inconsistencies with the documentation.
>
> 1. The documentation of the Optimize function states that the last parameter (iterate) is also used as the starting point of the algorithm. However, this parameter seems to be basically ignored and the starting point is generated randomly here.
>
> 2. When the algorithm does not converge, it suggests to use a smaller step size here, yet there is no way to change the step size.
>
> 3. The lowerBound and upperBound variables don't seem to be true bounds for the variables. As far as I understand the code, those are only used to deduce the initial sigma value and the random starting point.
>
> Have a nice day!

There are some responses in the original issue about how we should handle each point, but the issue is not yet solved.

sshkhr:
@rcurtin I could take a look at this once I'm finished with mlpack/mlpack#1510 

This could be a good place to get started for the NEAT project idea for gsoc too which I'm interested in

zoq:
CMA-ES isn't related to NEAT, but you get to know the optimizer infrastructure.

zoq:
Right, no need to restrict the starting point.

favre49:
@zoq For point 2, besides your suggestions, could we also include the initial step size as a parameter? Currently, we experience a problem when we use larger bounds, since the initial step size becomes too large. This causes the CMAES implementation to fail tests when the parameters are changed to include larger bounds (For e.g., [-25,25] for the LogisticRegressionTest). This problem is fixed if we change the step size.

zoq:
Sounds reasonable to me, I think this is already included in #88?

favre49:
No, I wanted to make sure it was alright before including it. I'll make the required changes to #88 soon

zoq:
Sounds good, thanks!

rcurtin:
Do you want me to wait to release the next version of ensmallen until #88 is merged?

favre49:
I'm not sure whether that question was meant for me, but I should let you know that i might not be able to work on #88 much for the next couple weeks, since I'm going to be busy with exams and stuff.

rcurtin:
[...]

PR #88, opened 2019-03-01
CMAES inconsistency fix

favre49:
This is a fix for issue #70 . The boundary handler has been adapted from the python code for CMA-ES
I had to change the bounds in the test functions since they did not align with the solution.
I will update the documentation for the code in a couple hours if there are no problems

### DIFF
--- a/include/ensmallen_bits/cmaes/cmaes.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes.hpp
@@ -107,9 +107,9 @@ class CMAES
   //! Modify the lower bound of decision variables.
   double& LowerBound() { return lowerBound; }
 
-  //! Get the upper bound of decision variables
+  //! Get the upper bound of decision variables.
   double UpperBound() const { return upperBound; }
-  //! Modify the upper bound of decision variables
+  //! Modify the upper bound of decision variables.
   double& UpperBound() { return upperBound; }
 
   //! Get the batch size.
@@ -132,14 +132,14 @@ class CMAES
   //! Modify the selection policy.
   SelectionPolicyType& SelectionPolicy() { return selectionPolicy; }
 
-  //! Get the step size damping
+  //! Get the step size damping.
   double StepSizeDamping() const { return ds; }
-  //! Modify the step size damping
+  //! Modify the step size damping.
   double& StepSizeDamping() { return ds; }
 
-  //! Get the cumulation constant
+  //! Get the cumulation constant.
   double CumulationConstant() const { return cc; }
-  //! Modify the cumulation constant
+  //! Modify the cumulation constant.
   double& CumulationConstant() { return cc; }
 
  private:
@@ -152,7 +152,7 @@ class CMAES
   //! Lower bound of decision variables.
   double lowerBound;
 
-  //! Upper bound of decision variables
+  //! Upper bound of decision variables.
   double upperBound;
 
   //! The batch size for processing.
@@ -167,13 +167,13 @@ class CMAES
   //! The selection policy used to calculate the objective.
   SelectionPolicyType selectionPolicy;
 
-  //! Step size damping
+  //! Step size damping.
   double ds;
 
-  //! Cumulation constant
+  //! Cumulation constant.
   double cc;
 
-  //! Methods used to transform the candidates into the constraints
+  //! Methods used to transform the candidates into the constraints.
   void BoundaryTransform(arma::mat& iterate);
   void BoundaryTransformInverse(arma::mat& iterate);
 };
--- a/include/ensmallen_bits/cmaes/cmaes.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes.hpp
@@ -163,6 +163,9 @@ class CMAES
   //! Cumulation constant
   double cc;
 
+  //! Methods used to transform the candidates into the constraints
+  void BoundaryTransform(arma::mat& iterate);
+  void BoundaryTransformInverse(arma::mat& iterate);
 };
 
 /**
--- a/include/ensmallen_bits/cmaes/cmaes.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes.hpp
@@ -60,6 +60,7 @@ class CMAES
    * equal one pass over the dataset).
    *
    * @param lambda The population size (0 use the default size).
+   * @param initialSigma The initial step size
    * @param lowerBound Lower bound of decision variables.
    * @param upperBound Upper bound of decision variables.
    * @param batchSize Batch size to use for the objective calculation.
@@ -70,6 +71,7 @@ class CMAES
    *     objective.
    */
   CMAES(const size_t lambda = 0,
+        const double initialSigma = 0.6,
         const double lowerBound = -10,
         const double upperBound = 10,
         const size_t batchSize = 32,
@@ -94,6 +96,11 @@ class CMAES
   size_t PopulationSize() const { return lambda; }
   //! Modify the population size.
   size_t& PopulationSize() { return lambda; }
+
+  //! Get the initial step size.
+  double InitialSigma() const { return initialSigma; }
+  //! Modify the initial step size.
+  double& InitialSigma() { return initialSigma; }
 
   //! Get the lower bound of decision variables.
   double LowerBound() const { return lowerBound; }
@@ -138,6 +145,9 @@ class CMAES
  private:
   //! Population size.
   size_t lambda;
+
+  //! Initial step size.
+  double initialSigma;
 
   //! Lower bound of decision variables.
   double lowerBound;
--- a/include/ensmallen_bits/cmaes/cmaes.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes.hpp
@@ -125,6 +125,16 @@ class CMAES
   //! Modify the selection policy.
   SelectionPolicyType& SelectionPolicy() { return selectionPolicy; }
 
+  //! Get the step size damping
+  double StepSizeDamping() const { return ds; }
+  //! Modify the step size damping
+  double& StepSizeDamping() { return ds; }
+
+  //! Get the cumulation constant
+  double CumulationConstant() const { return cc; }
+  //! Modify the cumulation constant
+  double& CumulationConstant() { return cc; }
+
  private:
   //! Population size.
   size_t lambda;
@@ -146,6 +156,13 @@ class CMAES
 
   //! The selection policy used to calculate the objective.
   SelectionPolicyType selectionPolicy;
+
+  //! Step size damping
+  double ds;
+
+  //! Cumulation constant
+  double cc;
+
 };
 
 /**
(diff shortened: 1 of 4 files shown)


zoq:
You can put the update right after the ``Fixes for SPSA (#87).`` line.

zoq:
Ah wrong PR.

favre49:
weird error here, the `sudo apt-get update` failed?

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -129,7 +129,7 @@ double CMAES<SelectionPolicyType>::Optimize(
     const size_t idx1 = i % 2;
 
     // Perform Cholesky decomposition. If the matrix is not positive definite,
-    // add a small value and try again
+    // add a small value and try again.
```
rcurtin: Thanks for adding that. :)

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -146,6 +166,16 @@ class CMAES
 
   //! The selection policy used to calculate the objective.
   SelectionPolicyType selectionPolicy;
+
+  //! Step size damping.
+  double ds;
```
rcurtin: Just for consistency, it might be beter to name this `stepSizeDamping` since the method to access it is `StepSizeDamping()`.  (Similar for `CumulationConstant()`.)  If we are exposing these parameters, should we also add constructor parameters for these also?  I think maybe I don't fully follow what these parameters are for (or how a user would set them).

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -90,19 +92,24 @@ class CMAES
   template<typename DecomposableFunctionType>
   double Optimize(DecomposableFunctionType& function, arma::mat& iterate);
 
-  //! Get the step size.
+  //! Get the population size.
   size_t PopulationSize() const { return lambda; }
-  //! Modify the step size.
+  //! Modify the population size.
```
rcurtin: Nice catch, thanks! :+1:

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -70,6 +71,7 @@ class CMAES
    *     objective.
    */
   CMAES(const size_t lambda = 0,
+        const double initialSigma = 0.6,
```
rcurtin: I'm concerned that this would break reverse compatibility and we'd have to bump the major version (I don't think that's justified by a relatively simple change like this).  According to our versioning policy,

```
 * A MAJOR version bump implies a backwards-incompatible change to the
   user-facing optimizers.  Code using one major version of ensmallen may
   not work with a newer major version of ensmallen.
```

We could instead add `initialSigma` as the last parameter with a default value, and then we could optionally mark that constructor deprecated ("`will be removed in 2.10.0`") and add another constructor just like this one here.  What do you think?

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -67,16 +69,16 @@ double CMAES<SelectionPolicyType>::Optimize(
 
   // Step size control parameters.
   arma::vec sigma(3);
-  sigma(0) = 0.3 * (upperBound - lowerBound);
+  sigma(0) = initialSigma;
   const double cs = (muEffective + 2) / (iterate.n_elem + muEffective + 5);
-  const double ds = 1 + cs + 2 * std::max(std::sqrt((muEffective - 1) /
+  ds = 1 + cs + 2 * std::max(std::sqrt((muEffective - 1) /
       (iterate.n_elem + 1)) - 1, 0.0);
```
rcurtin: I guess, then, that if the user has set `ds` or `cc`, they get overwritten here when `Optimize()` is called?

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -283,6 +287,79 @@ double CMAES<SelectionPolicyType>::Optimize(
...
+  double al = std::min((upperBound - lowerBound) / 2, (1 + std::abs(lowerBound)) / 20 );
+  double au = std::min((upperBound - lowerBound) / 2, (1 + std::abs(upperBound)) / 20 );
+  for (size_t ii = 0; ii < iterate.n_cols; ii++)
+  {
+    for (size_t it = 0; it < iterate.n_rows; it++)
+    {
+      double x = iterate(it,ii);
+
+      // Shift x into [lowerBound-al, upperBound+au].
+      if (x < lowerBound - 2 * al - (upperBound - lowerBound) / 2 || x > upperBound + 2 * au + (upperBound - lowerBound) / 2)
```
rcurtin: Style issue---this line is over 80 characters long.  Can you wrap it please?

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -283,6 +287,79 @@ double CMAES<SelectionPolicyType>::Optimize(
...
+template<typename SelectionPolicyType>
+void CMAES<SelectionPolicyType>::BoundaryTransform(arma::mat& iterate)
+{
+  double al = std::min((upperBound - lowerBound) / 2, (1 + std::abs(lowerBound)) / 20 );
+  double au = std::min((upperBound - lowerBound) / 2, (1 + std::abs(upperBound)) / 20 );
+  for (size_t ii = 0; ii < iterate.n_cols; ii++)
+  {
+    for (size_t it = 0; it < iterate.n_rows; it++)
+    {
+      double x = iterate(it,ii);
```
rcurtin: Tiny style issue, this should be `iterate(it, ii)`.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -283,6 +287,79 @@ double CMAES<SelectionPolicyType>::Optimize(
   return overallObjective;
 }
 
+// Transforms the candidate into the given bounds.
+template<typename SelectionPolicyType>
+void CMAES<SelectionPolicyType>::BoundaryTransform(arma::mat& iterate)
```
rcurtin: So, if I'm understanding that this function, given `iterate`, will transform each element of iterate such that it falls between `lowerBound` and `upperBound`, you could just use `arma::clamp()`, I think.  Correct me if I'm wrong and there's a bit that I missed.  (It looks like maybe there is some handling for when `upperBound` and `lowerBound` are too close together?  I have not figured out exactly what that does.)

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -151,8 +151,10 @@ double CMAES<SelectionPolicyType>::Optimize(
           pStep.slice(idx(j));
 
       // Calculate the objective function.
+      BoundaryTransform(pPosition.slice(idx(j)));
       pObjective(idx(j)) = selectionPolicy.Select(function, batchSize,
           pPosition.slice(idx(j)));
+      BoundaryTransformInverse(pPosition.slice(idx(j)));
```
rcurtin: Hmm, I guess I'm a little confused here.  I would imagine it would be easier to take `pPosition.slice(idx(j))` and simply clamp it to be within the desired range, but I'm not sure what the reverse transformation is.  Can you clarify the functionality here a bit?  Maybe I misunderstood.

include/ensmallen_bits/cmaes/cmaes.hpp
favre49: I'll change the name for sure. I included those two based on Marcus' suggestion in the issue discussion. These parameters are used in the update of the variables. It provides some extra flexibility for the user, but I wonder if including them in the constructor would become more confusing to the user, or how useful they would be (I use CMA-ES a lot for my work, and have never really toyed with these values). Maybe @zoq can shed some light on this?

include/ensmallen_bits/cmaes/cmaes.hpp
favre49: Yup, didn't think of that. I'll change it :)

include/ensmallen_bits/cmaes/cmaes.hpp
rcurtin: Ah, that's fair.  It doesn't matter to me if the variables are included or not, but if we do include them then I think we should make them accessible from the constructor. :+1:

gaurav-singh1998:
Hi, @favre49, @zoq, @rcurtin since @favre49 has implemented and addressed the majority part regarding the issue #70 I want to take up this and implement the necessary changes that @rcurtin suggested above. Can I move forward with it?

gaurav-singh1998:
Also, @favre49 can you provide the link of the source code or the paper that you referred for applying box constraints for each parameter [here](https://github.com/favre49/ensmallen/blob/CMAESIssue/include/ensmallen_bits/cmaes/cmaes_impl.hpp#L292)?

favre49:
@gaurav-singh1998 That would be great. I'll try to find the sources I used and get back to you

favre49:
If I'm not wrong, I used [pycma](https://github.com/CMA-ES/pycma)

favre49:
Also, what I failed to address in Ryan's comment pertained to how Nikolas Hansen, the author of the paper on CMA-ES advised handling bound constraints. I am unable to find the resources that I read this from, but the gist of it was that at every step, the result obtained would be "projected" into another space, where it would exist in the bounds (if I remember correctly). The way this has been done is in my code (though probably flawed) as well as in the pycma code. If I'm able to find it again, I'll let you know.

gaurav-singh1998:
Hi, @favre49 sorry for responding so late on this. After going through [this](https://github.com/gaurav-singh1998/c-cmaes/blob/master/src/boundary_transformation.c) implementation of boundary handling and an [example](https://github.com/gaurav-singh1998/c-cmaes/blob/master/src/example_boundary.c) which shows how to use it in the final CMAES implementation I found that they are not applying the inverse function after transforming them into the given box constraints like you have implemented [here](https://github.com/favre49/ensmallen/blob/CMAESIssue/include/ensmallen_bits/cmaes/cmaes_impl.hpp#L157) and [here](https://github.com/favre49/ensmallen/blob/CMAESIssue/include/ensmallen_bits/cmaes/cmaes_impl.hpp#L173). Kindly help me with this before moving forward. Thanks.

favre49:
I think instead of calling inverse [here](https://github.com/gaurav-singh1998/c-cmaes/blob/master/src/boundary_transformation.c#L67) they write the same thing as in the function [here](https://github.com/gaurav-singh1998/c-cmaes/blob/master/src/boundary_transformation.c#L120). Not sure why, though

gaurav-singh1998:
Hi, @favre49 in the `libcmaes` library, C++ version of CMAES by N Hansen, box constraints are handles in the same way i.e. they just provided a function for the inverse [here](https://github.com/nikohansen/libcmaes/blob/master/src/pwq_bound_strategy.cc). So, before moving further with this wanted to clarify should I also apply this concept for handling the box constraints? Thanks.

favre49:
Yeah, I think that would make sense

gaurav-singh1998:
Hi, @favre49 sorry to disturb you again and again but wanted to know can I open a new PR for this?

favre49:
No issues, I'll close this when you open your PR

PR #94, opened 2019-03-10
Add issue templates.

rcurtin:
These are just the same ones from the mlpack repository, slightly adapted where necessary.


coatless:
@rcurtin consider sticking these in an _org_-level repo. c.f. 

https://help.github.com/en/articles/creating-a-default-community-health-file-for-your-organization

https://help.github.com/en/articles/about-issue-and-pull-request-templates

rcurtin:
Nice, I didn't know about that.  For now though I'd just like to add the templates here; there are some other repositories we have where these templates wouldn't make as much sense. :+1:

PR #92, opened 2019-03-10
Add newline to end of `de.hpp` file

coatless:
Added a newline to the end of `de.hpp` to quiet a CRAN warning.

> Found the following sources/headers not terminated with a newline:
>   inst/include/ensmallen_bits/de/de.hpp

PR #95, opened 2019-03-11
`cyclical_decay.hpp` BibTeX fix

rcurtin:
It had two titles... not sure how that happened. :)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,8 @@
 ###### ????-??-??
   * Minor documentation fixes (#95).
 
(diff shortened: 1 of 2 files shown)


HISTORY.md
```diff
@@ -1,3 +1,7 @@
+### ensmallen 1.14.2
+###### ????-??-??
+  * Minor documentation fixes (#95).
```
zoq: Not really related, but what about adding https://github.com/mlpack/ensmallen/pull/92 to the history as well.

rcurtin: Sure :)

PR #100, opened 2019-04-05
fix: don't assume that arma::eig_sym() always succeeds

conradsnicta:
if we assume arma::eig_sym() succeeds, but in fact it fails, accessing elements in the evals vector will result either in a crash or an out-of-bounds error.

### DIFF
--- a/tests/sdp_primal_dual_test.cpp
+++ b/tests/sdp_primal_dual_test.cpp
@@ -182,8 +182,11 @@ ConstructMaxCutSDPFromLaplacian(const std::string& laplacianFilename)
 
 static bool CheckPositiveSemiDefinite(const arma::mat& X)
 {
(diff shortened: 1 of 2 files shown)


tests/sdp_primal_dual_test.cpp
```diff
@@ -182,8 +182,11 @@ ConstructMaxCutSDPFromLaplacian(const std::string& laplacianFilename)
 
 static bool CheckPositiveSemiDefinite(const arma::mat& X)
 {
-  const auto evals = arma::eig_sym(X);
-  return (evals(0) > 1e-20);
+  arma::vec evals;
+  if(!arma::eig_sym(evals, X))
```
zoq: Just a minor style issue, missing space right after the `ìf``, can be fixed in the merge process.

PR #101, opened 2019-04-06
Documentation fixup

greatsharma:
This PR fixes some documentations

### DIFF
--- a/include/ensmallen_bits/bigbatch_sgd/adaptive_stepsize.hpp
+++ b/include/ensmallen_bits/bigbatch_sgd/adaptive_stepsize.hpp
@@ -63,6 +63,7 @@ class AdaptiveStepsize
    * @param iterate Parameters that minimize the function.
    * @param gradient The gradient matrix.
    * @param gradientNorm The gradient norm to be used for the given iteration.
+   * @param sampleVariance Variance in the sample.
    * @param offset The batch offset to be used for the given iteration.
    * @param batchSize Batch size to be used for the given iteration.
    * @param backtrackingBatchSize Backtracking batch size to be used for the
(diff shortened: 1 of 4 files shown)


include/ensmallen_bits/bigbatch_sgd/adaptive_stepsize.hpp
```diff
@@ -63,6 +63,7 @@ class AdaptiveStepsize
    * @param iterate Parameters that minimize the function.
    * @param gradient The gradient matrix.
    * @param gradientNorm The gradient norm to be used for the given iteration.
+   * @param sampleVariance Variance in the sample.
```
rcurtin: This comment doesn't really make sense to me.  Can you clarify what the parameter actually means?

include/ensmallen_bits/gradient_descent/gradient_descent.hpp
```diff
@@ -45,11 +45,10 @@ class GradientDescent
  public:
   /**
    * Construct the Gradient Descent optimizer with the given function and
-   * parameters.  The defaults here are not necessarily good for the given
+   * parameters. The defaults here are not necessarily good for the given
```
rcurtin: Can we not make these changes everywhere?  It is okay to use one or two spaces (i.e., the world has not decided which of these is right and which of these isn't).  So let's leave it how it is.

include/ensmallen_bits/bigbatch_sgd/adaptive_stepsize.hpp
zoq: Agreed, we already had the same discussion in the mlpack repository.

PR #103, opened 2019-04-09
Add decay policies to SGD

greatsharma:
In this PR I added time-based, drop-based and exponential-based decay policies for stochastic gradient descent.
 
These learning rate schedulers are very common and highly used by machine learning users, providing such functionality would be great.

Very soon I will add tests for all these decay policies.

### DIFF
--- a/include/ensmallen_bits/sgd/decay_policies/drop_decay.hpp
+++ b/include/ensmallen_bits/sgd/decay_policies/drop_decay.hpp
@@ -1,0 +1,87 @@
+/**
+ * @file drop_decay.hpp
+ * @author Gaurav Sharma
+ *
+ * Drop based decay policy for Stochastic Gradient Descent. 
+ *
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef ENSMALLEN_SGD_DECAY_POLICIES_DROP_DECAY_HPP
+#define ENSMALLEN_SGD_DECAY_POLICIES_DROP_DECAY_HPP
+
+#include <cmath>
+
+namespace ens {
+
+/**
+ * Definition of the DropDecay class.
+ */
+class DropDecay
+{
+ public: 
+  /**
+   * This constructor is called before the first iteration.
+   * @param initialStepSize Step Size at the beginning.
+   * @param dropRate Factor by which stepSize is dropped.
+   * @param epochDrop Number of epochs after which stepSize is dropped.
+   */	
+	DropDecay(const double initialStepSize = 0.01,
+			  const double dropRate = 0.1,
+			  const size_t epochDrop = 1000) :
+	initialStepSize(initialStepSize),
+	dropRate(dropRate),
+	epochDrop(epochDrop),
+	epoch(0),
+	effectiveBatchSize(1)
+	{ /* Nothing to do here. */}
+	
+  /**
+   * This function is called in each iteration after the policy update.
+   * It sets the value of effective batch size.
+   *
+   * @param effectiveBatchSize current effective batch size.
+   */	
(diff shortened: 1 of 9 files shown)


greatsharma:
` /**
  * This function is called in each iteration after the policy update.
  * It sets the value of effective batch size.
  *
  * @param effectiveBatchSize current effective batch size.
  */	
  void setEffectiveBatchSize(const size_t& effBatchSize)
  {
		effectiveBatchSize = effBatchSize;
  }
`

include/ensmallen_bits/sgd/decay_policies/drop_decay.hpp
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * This function is called in each iteration after the policy update.
+   * It sets the value of effective batch size.
+   *
+   * @param effectiveBatchSize current effective batch size.
+   */	
+  void setEffectiveBatchSize(const size_t& effBatchSize)
+  {
+	effectiveBatchSize = effBatchSize;
+  }
```
greatsharma: I added this new function setEffectiveBatchSize to the decay policy template so to get the effectiveBatchSize and update the epoch according to it. Although initially I didn't make any changes to the decay policy template provided by author @zoq  in no_decay.hpp and everything indeed worked by just doing ++epoch. But I personally think that this approach is better and  more robust.

include/ensmallen_bits/sgd/decay_policies/drop_decay.hpp
```diff
@@ -0,0 +1,87 @@
...
+   * @param gradient The gradient matrix.
+   */
+  void Update(arma::mat& /* iterate */,
+              double& stepSize,
+              const arma::mat& /* gradient */)
+  {
+	epoch += effectiveBatchSize;
+    stepSize = initialStepSize * pow(dropRate, floor((1.0 + epoch) / epochDrop));
+  }
+
```
greatsharma: I then used the effectiveBatchSize set by setEffectiveBatchSize method for updating epoch. Although without using effectiveBatchSize ie., just doing ++epoch and keeping the decay policy template intact things worked but this seems more promising to me.

Apart from this if @zoq  wants me to remove the added function setEffectiveBatchSize() and do completely according to no_decay.hpp template then I have no problem and good to go.

Rest is your call.

include/ensmallen_bits/sgd/decay_policies/no_decay.hpp
```diff
@@ -27,14 +28,25 @@ class NoDecay
...
+ /**
+  * This function is called in each iteration after the policy update.
+  * It sets the value of effective batch size.
+  *
+  * @param effectiveBatchSize current effective batch size.
+  */
+  void setEffectiveBatchSize(const size_t& /* effectiveBatchSize */)
+  {
+	/* Nothing to do here. */
+  }
```
greatsharma: Added new function to decay policy template.

zoq:
Since this already uses the existing decay policy that can be used with the SGD optimizer, but I agree we shouldn't merge this until there is a proper documentation for using either different decay methods or there are corresponding typedefs. Also, do you have any further references you used, I wonder if there any benchmarks out there.

greatsharma:
@zoq frankly speaking there is no such specific reference from where I followed this, I studied this some time back and don't exactly remember the source but I found two three great sources online and all are exactly the same, here is one of them 

https://www.google.com/url?sa=t&source=web&rct=j&url=https://towardsdatascience.com/learning-rate-schedules-and-adaptive-learning-rate-methods-for-deep-learning-2c8f433990d1&ved=2ahUKEwjOkpaYjcvhAhWMfysKHUQQAXcQjjgwAXoECAYQAQ&usg=AOvVaw0ToDf2tZSseqKmZeHGD-Ra

Apart from this you guys tell me where to add documentation, I will try my best to explain then you guys can verify that. And also please have a look at the code because in ensmallen we did a bit different from other libraries I mean instead of taking the number of epochs we take the maximum iterations ( where one iteration mean 1 single point ) which was a bit different from other libraries but I tried it and infact I did these decay policy with different approach earlier but get the same results both ways so I opted this as this properly follow the template mentioned by @zoq in no_decay.hpp

Issue #104, opened 2019-04-10
Infinite loop bug in SGD


greatsharma:
When we run SGD with 0 batch size then the loop goes infinite. Although it is very stupid thing to use 0 batch size but user can accidentally set batch size to 0 ( as I did it accidentally ). And as training on very large data set and complex machine learning models do take large amount of time so if an user accidentally set batch size 0 for such situation then they may think that the optimizer is taking so long time probably due to large data set and complex model but indeed they were stuck in a infinite loop and were not aware.

And if we set batch size to 0 then logically our optimizer shouldn't do anything i.e., it will not update the iterate ( initial points given by user ) and the overallObjective return by the optimizer is the value of the function which is to be optimized at that initial point given by the user.

Apart from this as software engineering tells us that while testing we should handle such boundary conditions.

greatsharma:
This issue is solved in #105.

rcurtin:
Agreed, it would be useful to catch this condition.  Thanks for looking into it.

zoq:
Personally, I find it difficult to draw the line between what should be checked and what not, as you pointed out, using a batch size of 0 doesn't make much sense. Also, I think we can create a bunch of optimizer parameter combinations that will trigger a similar behavior.

greatsharma:
@zoq I do agree with your point but I mentioned it just because I stuck in it accidentally and after few minutes when I close the execution and look closer to sgd class then I realised this. But maybe user don't look at the sgd class and they just use mlpack for developing purpose and not for contribution. Rest as mentioned by @rcurtin in #105 I do throw an exception for that.

PR #105, opened 2019-04-10
fix infinite loop bug in SGD

greatsharma:
This PR fixes the bug reported in the issue #104.

If we set batch size to 0 then our optimizer shouldn't do anything i.e., it will not update the iterate ( initial points given by user ) and the overallObjective return by the optimizer is the value of the function which is to be optimized at that initial point given by the user. And this is what I get by doing changes mentioned in this PR.

### DIFF
--- a/include/ensmallen_bits/sgd/sgd_impl.hpp
+++ b/include/ensmallen_bits/sgd/sgd_impl.hpp
@@ -116,6 +116,11 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
         std::min(batchSize, actualMaxIterations - i),
         numFunctions - currentFunction);
 
+    if( effectiveBatchSize == 0 )
+	{
+	  break;
+	}
+
     // Technically we are computing the objective before we take the step, but
     // for many FunctionTypes it may be much quicker to do it like this.
     overallObjective += f.EvaluateWithGradient(iterate, currentFunction,
@@ -138,9 +143,16 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
   overallObjective = 0;
   for (size_t i = 0; i < numFunctions; i += batchSize)
   {
-    const size_t effectiveBatchSize = std::min(batchSize, numFunctions - i);
-    overallObjective += f.Evaluate(iterate, i, effectiveBatchSize);
+    size_t effectiveBatchSize = std::min(batchSize, numFunctions - i);
+
+	if (effectiveBatchSize == 0)
+	{
+	  effectiveBatchSize = 1;
+	  ++i;
+	}		
+	overallObjective += f.Evaluate(iterate, i-1, effectiveBatchSize);
   }
+
   return overallObjective;
 }
 
--- a/include/ensmallen_bits/sgd/sgd_impl.hpp
+++ b/include/ensmallen_bits/sgd/sgd_impl.hpp
@@ -117,9 +117,9 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
         numFunctions - currentFunction);
 
     if( effectiveBatchSize == 0 )
-	{
-	  break;
-	}
+    {
+      break;
+    }
 
     // Technically we are computing the objective before we take the step, but
     // for many FunctionTypes it may be much quicker to do it like this.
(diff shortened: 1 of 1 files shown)


include/ensmallen_bits/sgd/sgd_impl.hpp
```diff
@@ -116,6 +116,11 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
         std::min(batchSize, actualMaxIterations - i),
         numFunctions - currentFunction);
 
+    if( effectiveBatchSize == 0 )
+	{
+	  break;
+	}
```
greatsharma: If effectiveBatchSize is 0 then we should break from this loop otherwise we will be trapped in infinite looping.

include/ensmallen_bits/sgd/sgd_impl.hpp
```diff
@@ -138,9 +143,16 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
...
-    overallObjective += f.Evaluate(iterate, i, effectiveBatchSize);
+    size_t effectiveBatchSize = std::min(batchSize, numFunctions - i);
+
+	if (effectiveBatchSize == 0)
+	{
+	  effectiveBatchSize = 1;
+	  ++i;
+	}		
+	overallObjective += f.Evaluate(iterate, i-1, effectiveBatchSize);
   }
```
greatsharma: First we calculate effectiveBatchSize and if it is 0 then make effectiveBatchSize to 1 for calculating overallObjective and increment increment i by one. Then call evaluate with iterate, i-1( as we incremented it before passing ) and effectiveBatchSize.

include/ensmallen_bits/sgd/sgd_impl.hpp
```diff
@@ -138,9 +143,16 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
...
+
```
greatsharma: First we calculate effectiveBatchSize and if it is 0 then make effectiveBatchSize to 1 for calculating overallObjective and increment increment i by one. Then call evaluate with iterate, i-1( as we incremented it before passing ) and effectiveBatchSize. 

++i is done to move the loop forward because batchSize is 0 which means i+=batchSize doesn't update i and we will be trapped in infinite looping.

include/ensmallen_bits/sgd/sgd_impl.hpp
```diff
@@ -138,9 +143,16 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
   overallObjective = 0;
   for (size_t i = 0; i < numFunctions; i += batchSize)
   {
-    const size_t effectiveBatchSize = std::min(batchSize, numFunctions - i);
-    overallObjective += f.Evaluate(iterate, i, effectiveBatchSize);
+    size_t effectiveBatchSize = std::min(batchSize, numFunctions - i);
```
greatsharma: First we calculate effectiveBatchSize and if it is 0 then make effectiveBatchSize to 1 for calculating overallObjective and increment increment i by one. Then call evaluate with iterate, i-1( as we incremented it before passing ) and effectiveBatchSize. 

++i is done to move the loop forward because batchSize is 0 which means i+=batchSize doesn't update i and we will be trapped in infinite looping.

include/ensmallen_bits/sgd/sgd_impl.hpp
```diff
@@ -138,9 +143,16 @@ double SGD<UpdatePolicyType, DecayPolicyType>::Optimize(
...
+
+	if (effectiveBatchSize == 0)
+	{
+	  effectiveBatchSize = 1;
+	  ++i;
+	}		
+	overallObjective += f.Evaluate(iterate, i-1, effectiveBatchSize);
```
greatsharma: First we calculate effectiveBatchSize and if it is 0 then make effectiveBatchSize to 1 for calculating overallObjective and increment increment i by one. Then call evaluate with iterate, i-1( as we incremented it before passing ) and effectiveBatchSize. 

++i is done to move the loop forward because batchSize is 0 which means i+=batchSize doesn't update i and we will be trapped in infinite looping.

greatsharma:
@rcurtin I will get back on this very soon.

PR #106, opened 2019-04-11
correct code in momentum_sgd_test.cpp

greatsharma:


greatsharma:
I agree and even noticed at the moment I pointed this that this will not change anything as both are exactly same just change in variable name and nothing else,  but it is more clear to use f1 corresponding to coordinates1 and s1.

Rest to you guys, I have no issue with any decision.

zoq:
Thanks for the contribution.

Issue #108, opened 2019-04-25
Add option to avoid computing exact objective at end of optimization


rcurtin:
At the end of most of the SGD-like optimizers, we have a loop that loops over all parts of the decomposable function and computes the final objective.  But this is problematic in that the decomposable function could have *very many* functions, meaning that the final objective computation takes a very long time.  This could easily be seen with, e.g., training a machine learning model, where you might take 20 passes over the data and each pass could take an hour or more.  Computing the exact final objective is equivalent to one additional pass over the data.

To solve this, we could add an extra parameter to each of our decomposable function optimizers: `exactObjective`.  If `exactObjective` is true, then the final objective will be computed exactly, just like it is now.  If it's false (which should be the default), then just the estimate of the final objective obtained on the last pass over the data will be used.

In general, only estimates of the final objective function are given back by all deep learning toolkits and machine learning optimization toolkits I have encountered.  So computing the exact objective every time is basically just a computation disadvantage for us, and although it can be useful, we should allow the user to specify whether they want it (and assume that they don't, to match what others are doing).

SinaMojtahedi:
Dear all,
Lets consider a system of equations are as below:
((1-x(1))(10+x(5)))^2/(((x(1)^2)x(3)^2)+((1-x(1))^2x(2)^2)) = 0.154;
((1-x(1))(10-x(5)))^2/(((x(1)^2)x(3)^2)+((1-x(1))^2x(2)^2)) = 0.254;
(x(1)(10+x(4)))^2/(((x(1)^2)x(3)^2)+((1-x(1))^2x(2)^2)) = 0.545;
(x(1)(10-x(4)))^2/(((x(1)^2)x(3)^2)+((1-x(1))^2x(2)^2)) = 0.034;
(10+(x(1)*x(4))+(1-x(1))*x(5))^2/(((x(1)^2)x(3)^2)+((1-x(1))^2x(2)^2)) = 0.994;
(10-(x(1)*x(4))-(1-x(1))x(5))^2/(((x(1)^2)x(3)^2)+((1-x(1))^2x(2)^2)) = 0.063;
and I want to find x(1) ... x(5) values so that the equalities hold.
I already tried so many optimization methods but the problem is, after finding the values when I replace them with x values the error values I came across are very large. That is if we consider for example
[...]

PR #109, opened 2019-04-28
Providing option to calculate exact objective

walragatver:
Hi, this is for #108. It looks like we need to modify some `tests` in the `mlpack` repository after this change. I will start working on it once this gets approved.


include/ensmallen_bits/ada_delta/ada_delta.hpp
```diff
@@ -64,6 +64,9 @@ class AdaDelta
    * @param tolerance Maximum absolute tolerance to terminate algorithm.
    * @param shuffle If true, the function order is shuffled; otherwise, each
    *        function is visited in linear order.
+   * @param exactObjective Flag that determines whether actual objective over
```
zoq: Sounds a little bit vague to me, what about "Calculate the exact objective (Default: estimate the final objective obtained on the last pass over the data)."

include/ensmallen_bits/ada_delta/ada_delta.hpp
```diff
@@ -128,6 +132,11 @@ class AdaDelta
   //! Modify whether or not the individual functions are shuffled.
   bool& Shuffle() { return optimizer.Shuffle(); }
 
+  //! Get whether or not the actual objective is calculated after training.
```
zoq: Not sure, I would probably remove the "after training" part, see this more as an optimisation process.

include/ensmallen_bits/ada_delta/ada_delta.hpp
walragatver: Yes, this is quite clear.

include/ensmallen_bits/ada_delta/ada_delta.hpp
walragatver: Sorry, I just got carried away.

include/ensmallen_bits/cne/cne_impl.hpp
```diff
@@ -119,7 +121,11 @@ double CNE::Optimize(DecomposableFunctionType& function, arma::mat& iterate)
   // Set the best candidate into the network parameters.
   iterate = population.slice(index(0));
 
-  return function.Evaluate(iterate);
+  if (exactObjective)
+  {
+    lastBestFitness = function.Evaluate(iterate);
+  }
+  return lastBestFitness;
```
rcurtin: Since CNE uses only an arbitrary function type, there's no need for `exactObjective` here.  It only applies to optimizers that optimize decomposable functions.  I think the documentation is a little wrong for CNE though (I updated it in another PR that I'm about to open).

include/ensmallen_bits/scd/scd_impl.hpp
```diff
@@ -89,8 +91,12 @@ double SCD<DescentPolicyType>::Optimize(ResolvableFunctionType& function,
...
       << "terminating optimization." << std::endl;
 
-  // Calculate and return final objective.
-  return function.Evaluate(iterate);
+  // Calculate and final objective if exactObjective is set to true.
+  if (exactObjective)
+  {
+    lastObjective = function.Evaluate(iterate);
+  }
+  return lastObjective;
```
rcurtin: Just like CNE, SCD doesn't work on decomposable functions, so I think we can leave SCD as it is. :+1:

include/ensmallen_bits/spsa/spsa_impl.hpp
```diff
@@ -98,8 +100,12 @@ inline double SPSA::Optimize(
...
   }
 
-  // Calculate final objective.
-  return function.Evaluate(iterate);
+  // Calculate final objective if exactObjective is set to true.
+  if (exactObjective)
+  {
+    overallObjective = function.Evaluate(iterate);
+  }
+  return overallObjective;
```
rcurtin: SPSA doesn't work on decomposable functions, so we can avoid the change here too.

walragatver:
Hi @rcurtin , @zoq ,
This is now ready to merge. I have currently kept tommorow's date in `History.md`.

HISTORY.md
```diff
@@ -1,3 +1,7 @@
+### ensmallen 1.1?.?
+###### ????-??-??
+  * Add option to avoid computing exact objective at end of optimization (#109).
```
zoq: Super picky comment: "at the end of the optimization".

doc/optimizers.md
```diff
@@ -1485,11 +1497,12 @@ the following other optimizers:
  | `size_t` | **`max_iterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
  | `double` | **`tolerance`** | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
  | `bool` | **`shuffle`** | If true, the function order is shuffled; otherwise, each function is visited in linear order. | `true` |
+ | `bool` | **`exactObjective`** | Calculate the exact objective (Default: estimate the final objective obtained on the last pass over the data). | `false` |
  | `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call; otherwise, their values are retained. | `true` |
 
  The attributes of the optimizer may also be modified via the member methods
  `StepSize()`, `BatchSize()`, `Beta1()`, `Beta2()`, `Eps()`, `MaxIterations()`,
- `Tolerance()`, `Shuffle()`,`V1()`,`V2()`, and `ResetPolicy()`.
+ `Tolerance()`, `Shuffle()`, `ExactObjective()`, `V1()`, `V2()`, and `ResetPolicy()`.
```
zoq: Nice catch!

walragatver:
Hi @zoq,
Thanks for the final approval. I have made the required changes.

coatless:
Regarding the parameter ordering, the current function call is:

```
AdaDelta(stepSize, batchSize, rho, epsilon, maxIterations, tolerance, shuffle, resetPolicy)
```

In this update, the function call switches to: 

```
AdaDelta(stepSize, batchSize, rho, epsilon, maxIterations, tolerance, shuffle, **exactObjective**, resetPolicy)
```

Perhaps the addition of `exactObjective` should be at the end of the function definition?

```
AdaDelta(stepSize, batchSize, rho, epsilon, maxIterations, tolerance, shuffle, resetPolicy, **exactObjective**)
```

walragatver:
@coatless, Thanks for raising the issue. I also wanted to discuss that. If you look in some of the cases `UpdateRule`  parameter comes after the `shuffle` parameter. If we keep `exactObjective` after that then in order to use it a user will need to provide the `updateRule` parameter always. I thought it's not a great idea. So I kept `exactObjective` after `shuffle` as both are `bools` indicating some optional functionality. 
Let me know if this is incorrect.

zoq:
Good point, I think I rather change the update policies instead of changing the behavior of the objective calculation. So personally I would add the value at the end. Note this will be part of the next major release so we don't necessarily have to worry about backward compatibility, but I think it would be great to have anyway.

zoq:
Let us keep this open, I think this is close to being merged.

walragatver:
Hi @rcurtin, 
 Sorry for the slow response. Instead of adding a new constructor I have added `exactObjective` to the last position. I did that because it was difficult to `overload `the constructor because the last three arguments are of same `type`(`bool`). I have modified everything and this is ready to merge.

rcurtin:
@walragatver awesome work, thanks again! :+1:

Issue #110, opened 2019-05-05
error: using namespace ens


merria28:
I installed "ensmallen" successfully. Then an error occur when i calling by "using namespace ens".
The error is following:
Error	C2664	'arma::Mat<double>::Mat(arma::Mat<double> &&)': cannot convert argument 1 from 'arma::uword' to 'const arma::SizeMat &' 

at  line 92 in file "atoms.hpp" :
arma::mat gradient(size(x));

conradsnicta:
Try the following workaround:

    arma::mat gradient(arma::size(x));

Note that MSVC is full of bugs. For any real work, I suggest using a proper compiler like gcc or clang.

merria28:
@conradsnicta  thanks a lot.  This issue is solved.

PR #112, opened 2019-05-06
workaround for MSVC bug

conradsnicta:
add arma:: prefix to size(x)
see https://github.com/mlpack/ensmallen/issues/110

### DIFF
--- a/include/ensmallen_bits/fw/atoms.hpp
+++ b/include/ensmallen_bits/fw/atoms.hpp
@@ -89,7 +89,7 @@ class Atoms
       // Solve for current gradient.
       arma::mat x;
       RecoverVector(x);
-      arma::mat gradient(size(x));
+      arma::mat gradient(arma::size(x));
       function.Gradient(x, gradient);
 
       // Find possible atom to be deleted.


conradsnicta:
As far as I can tell, gcc and clang follow the rules for Argument Dependent Lookup (ADL): 
https://en.cppreference.com/w/cpp/language/adl
https://en.wikipedia.org/wiki/Argument-dependent_name_lookup

Say we have an object `bar::x` and a function `bar::foo(x)`, where `bar` is a namespace.  Let's say `bar::x` is instantiated and the user calls `foo(x)` without the `bar::` namespace prefix. The C++ compiler will "helpfully" search the namespace where `x` came from, to see if there is a function called `foo()` that accepts `x`.

I think the problem might be that some compilers are rather trigger happy and search all available namespaces (or search them in the wrong order) for functions that might accept `x`, or accept `x` after silently converting it to another type (possibly through multiple conversion hops). The entire process looks and smells brittle, especially for a common function name like `size()`.

All of this ADL chicanery seems like it was poorly thought out in the first place by the C++ committee. The primary use case for ADL is easy access to overloaded operators, in order to keep code readable. For example, ADL makes stuff like `arma::mat A, B; arma::mat C = A+B` possible, where `arma::operator+` is inferred for the two `arma::mat` objects.  However, allowing ADL for all functions (not just operators) is just asking for trouble -- it is anti-intuitive.

conradsnicta:
See also this overview of the std::size() disaster: https://quuxplusone.github.io/blog/2018/11/04/std-hash-value/

rcurtin:
@conradsnicta ah, I did not know about ADL, thanks for the links.  I think your take is right---yet another potentially dangerous C++ feature that's handled differently by different compilers. :)

Issue #62, opened 2018-12-11
Templatized `Optimize()`


rcurtin:
Each optimizer currently has a signature like

```
template<typename FunctionType>
void Optimize(FunctionType& f, arma::mat& coordinates);
```

but really it might be nicer if we did

```
template<typename FunctionType, typename MatType>
void Optimize(FunctionType& f, MatType& coordinates);
```

The code change should be straightforward for each optimizer but there is one main issue from my end: all the documentation as written suggests writing `FunctionType` functions with `arma::mat`.  Making that documentation fully templatized would make it harder to provide really simple examples, so I have to wonder if adding some extra section on "advanced functionality" that mentions that different types of Armadillo matrices can typically be used for optimization.  If anyone has any other ideas, feel free to propose them.

Also, it's probably worth writing at least one test for a different-precision Armadillo matrix for each optimizer, but existing tests can be easily adapted, I think.  (I suppose also that means that the `Evaluate()` and `Gradient()` for everything in `problems/` will need to be adapted, but again that too should be straightforward.)

zoq:
> all the documentation as written suggests writing FunctionType functions with arma::mat. Making that documentation fully templatized would make it harder to provide really simple examples, so I have to wonder if adding some extra section on "advanced functionality" that mentions that different types of Armadillo matrices can typically be used for optimization. If anyone has any other ideas, feel free to propose them.

I think an extra section works just fine, would be great if we could keep the documentation/examples as simple as possible.

conradsnicta:
I recommend not making the MatType "too flexible", as people can and will try doing silly things like feeding integer matrices to the optimisers.  We probably don't want to deal with complex types either, as (based on experience) that can complicate a lot of the code.  That leaves only two viable element types: float and double.

One way to place the restriction is to first modify the signature for Optimize() et al to something like:

    template<typename FunctionType, typename eT>
    void Optimize(FunctionType& f, Mat<eT>& coordinates);

The type restriction can then be placed on the output type of the function:

    template<typename FunctionType, typename eT>
    typename arma::enable_if2< arma::is_real<eT>::value, void>::result
    Optimize(FunctionType& f, Mat<eT>& coordinates);

... or as a dummy argument:

    template<typename FunctionType, typename eT>
    void
    Optimize(FunctionType& f, Mat<eT>& coordinates, const typename arma::arma_real_only<eT>::result* junk = 0);

[...]

PR #113, opened 2019-05-12
Templated Optimize()

rcurtin:
It's taken me a long time to put this together but this addresses #62 and provides a path for both optimizing sparse matrices (which is nice) and optimizing Bandicoot matrices (which is my real longer-term goal).

The short description of this PR is that now instead of having `double Optimize(FunctionType& function, arma::mat& coordinates)` as the signature of every optimizer, it's

```
template<typename FunctionType, typename MatType, typename GradType>
typename MatType::elem_type Optimize(FunctionType& function, MatType& coordinates);
```

instead.  This allows the user to pass any type they want---that could be `arma::mat`, `arma::fmat`, `arma::sp_mat`, and so forth.  By default, the type is checked for validity (which usually means it needs to be `arma::mat`, `arma::fmat`, `arma::sp_mat`, or `arma::sp_fmat`, but sometimes the sparse versions are disabled because they can't be used).  But, type checking can be disabled by specifying the `ENS_DISABLE_TYPE_CHECKS` macro before including `ensmallen.hpp`.  Static assertions are used to provide (somewhat) nice error messages.

The long description of this PR with individual changes highlighted to make review easier:

  1. All the `Optimize()` signatures changed, and in general I added tests to ensure that the `Optimize()` overloads would work with `MatType` set to `arma::fmat` or `arma::sp_mat` where applicable.

  2. In order to get the type checking right, each of the function forms like `EvaluateForm`, `GradientForm`, etc., had to be modified and put inside of a templated struct called `TypedForms`.  So, now, the `Function<>` template also takes the `MatType` and `GradType` parameters, and so do the type checking functions like `CheckNonDifferentiableFunctionTypeAPI()` and others.

  3. For the sparse functionality, newer versions of Armadillo are needed so that sparse `vectorise()` and `min()` and `max()` are available.  I simply put `#ifdef`s depending on the Armadillo version around the relevant tests instead of backporting that functionality.

  4. The SDP code needed some amount of overhaul for this to make sense.  So the API has changed and so has the documentation.  I deprecated the old versions; so they will still work until we remove them in ensmallen 2.10.0.

  5. I added a section to the documentation about templatized `Optimize()`, so users can realize that they can use any type.  I will also add to the website's "Questions" section and perhaps one or two other places for visibility (not sure about that yet).


include/ensmallen_bits/config.hpp
```diff
@@ -46,7 +46,18 @@
...
   #define ENS_PRAGMA_OMP_ATOMIC   _Pragma("omp atomic")
+  #define ENS_PRAGMA_OMP_CRITICAL _Pragma("omp critical")
 #else
   #define ENS_PRAGMA_OMP_PARALLEL
   #define ENS_PRAGMA_OMP_ATOMIC
+  #define ENS_PRAGMA_OMP_CRITICAL
+#endif
+
+
+// Define arma_deprecated for deprecated functionality.
```
conradsnicta: ens_deprecated

include/ensmallen_bits/config.hpp
```diff
@@ -46,7 +46,18 @@
...
+
+#if defined(_MSC_VER)
+  #define ens_deprecated __declspec(deprecated)
+#else
+  #define ens_deprecated __attribute__((__deprecated__))
```
conradsnicta: `__attribute__` is only available with gcc and clang, so at a minimum a guard is needed here to allow this code only for these compilers.

For sanity, each compiler should be treated separately though.  I suggest treating gcc and clang as two separate compilers.  Last time I checked clang reports that it's gcc 4.2, but in reality its support for attributes is closer to gcc 4.9.

For gcc, the guard can be along these lines:

    #if defined(__GNUG__) && (!defined(__clang__))
      #undef  ens_deprecated
      #define ens_deprecated __attribute__((__deprecated__))
    #endif

For clang, I suggest to use a guard and clang specific code:
    
    #if defined(__clang__) 
      #if __has_attribute(__deprecated__)
      #undef  ens_deprecated
      #define ens_deprecated __attribute__((__deprecated__))
    #endif

The undefines are for good hygiene.

include/ensmallen_bits/sdp/lin_alg.hpp
```diff
@@ -33,43 +33,49 @@ inline size_t SvecIndex(size_t i, size_t j, size_t n)
...
-  const size_t n = input.n_rows;
+  arma::SpMat<ElemType> iMat(input);
+
+  const size_t n = iMat.n_rows;
   const size_t n2bar = n * (n + 1) / 2;
 
   output.zeros(n2bar, 1);
 
-  for (auto it = input.begin(); it != input.end(); ++it)
+  for (auto it = iMat.begin(); it != iMat.end(); ++it)
```
conradsnicta: We don't know how well a given compiler will handle the `it != iMat.end();` part, ie. whether it will assume that `iMat.end();` is constant and hence not recreate the iterator for each iteration of the loop.  If it decides it can't assume that, we are wasting cycles here.

To avoid the above situation, I suggest declaring `auto it_end = iMat.end();` before the loop, and changing the loop declaration to:

    for (auto it = iMat.begin(); it != it_end; ++it)

include/ensmallen_bits/iqn/iqn_impl.hpp
```diff
@@ -89,36 +113,37 @@ double IQN::Optimize(DecomposableFunctionType& function, arma::mat& iterate)
...
+            Q[it] / arma::as_scalar(s.t() * Q[it] * s);
 
         // Update aggregate Hessian approximation.
-        B += (1.0 / numBatches) * (stochasticHessian - Q.slice(it));
+        B += (1.0 / numBatches) * (stochasticHessian - Q[it]);
 
         // Update aggregate Hessian-variable product.
-        u += (1.0 / numBatches) * (stochasticHessian * iterateVec -
-            Q.slice(it) * t.slice(it));
+        u += arma::reshape((1.0 / numBatches) * (stochasticHessian *
```
conradsnicta: Without thorough analysis, I don't know whether this call to reshape() recreates the given matrix with a different total number of elements (either less or more elements).  However, bear in mind that in the current implementation of SpMat::reshape(), the total number of elements can't be changed.

include/ensmallen_bits/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -103,29 +141,34 @@ double ParallelSGD<DecayPolicyType>::Optimize(
...
+        // TODO: support for batch size > 1 could be really useful.
         function.Gradient(iterate, visitationOrder[j], gradient, 1);
 
         // Update the decision variable with non-zero components of the
         // gradient.
         for (size_t i = 0; i < gradient.n_cols; ++i)
         {
           // Iterate over the non-zero elements.
-          for (arma::sp_mat::iterator cur = gradient.begin_col(i);
+          for (typename GradType::iterator cur = gradient.begin_col(i);
```
conradsnicta: I suggest to declare `cur_end = gradient.end_col(i);` outside of the loop, and then use `cur_end` inside the loop declaration.

coatless:
@rcurtin are you planning on rolling v1.15.x with this soon? If so, I'll hold off on pushing a new version of the _R_ package to CRAN.

rcurtin:
Depends on the definition of soon---this will probably be a couple weeks before it's merged.  (I guess it could be faster but it will probably take a little while for review.)  I will release 1.15.0 when it's merged though.

conradsnicta:
@rcurtin 
> I will release 1.15.0 when it's merged though.

1.15.0 doesn't really communicate the scope of the changes since 1.14.x.  This should be looked at from the point of view of cumulative changes since the initial 1.10.0 release.

I suggest calling this 2.10, or at the very least 1.20.  This is for two reasons:
1. To indicate a major new feature. This ties in with marketing -- people love new shiny things -- which leads to greater adoption. A bump in the major version is not solely reserved to API breaking changes.
2. The changes to allow fmat+etc tantamount to a massive rewrite of ensmallen, which brings its own set of risks. Regressions tests notwithstanding, there may very well be breakage with user code, as it's impossible for foresee the side-effects of such large changes. C++ is complicated enough that an innocuous looking change can screw things up (speaking from multiple first hand experiences). By having the version bumped to 2.10 communicates that breakage is possible.
3. It's easier to tell people to use ensmallen 2.x+ if they want to handle fmat, sp_mat, etc, in contrast to "use version later than 1.1492653589793, but not earlier"  :)

rcurtin:
@conradsnicta---you're right on all three points.  The reason I didn't suggest that is I was hoping to also get the callbacks done before ensmallen 2.10.0, then 2.x would be "callbacks and templated `Optimize()`".  I think, then, that we should wait to merge this until both are ready then.  (Also, my next priority is callbacks, so it won't be too long.)  @coatless that will mean it will be a little while longer until the release, definitely more than two weeks at this point.

include/ensmallen_bits/config.hpp
rcurtin: You've done way more fighting with strange compiler versions than I have, so I trust your input here.  But... could I trust your input authoritatively by doing `#define ens_deprecated arma_deprecated`?

include/ensmallen_bits/config.hpp
rcurtin: Thanks, fixed.  You maybe can see where I based the code off of. :)

include/ensmallen_bits/sdp/lin_alg.hpp
rcurtin: Agreed, thanks for pointing it out. :+1:

include/ensmallen_bits/parallel_sgd/parallel_sgd_impl.hpp
rcurtin: Agreed, thanks. :+1:

conradsnicta:
@rcurtin - it's always possible to "defer" callbacks to 3.10. I suspect there might be a bit of unintended fallout with the templated Optimize(), leading to a bunch of necessary 2.x releases.  This is actually okay and expected when such large changes are made.

Rather than waiting until everything is complete so we have a "pristine" 2.x release, it might be better to follow the motto of "release early and release often" (see Refs. [1] and [2] below).  This will allow quicker feedback from users.

I don't have a problem with bumping the major version from 2.x to 3.x in order to indicate that callbacks are now supported.

[1] https://en.wikipedia.org/wiki/The_Cathedral_and_the_Bazaar  
[2] https://en.wikipedia.org/wiki/Release_early,_release_often

include/ensmallen_bits/ada_delta/ada_delta_update.hpp
```diff
@@ -104,18 +61,77 @@ class AdaDeltaUpdate
...
+     * @param iterate Parameters that minimize the function.
+     * @param stepSize Step size to be used for the given iteration.
+     * @param gradient The gradient matrix.
+     */
+    void Update(MatType& iterate,
+                const double stepSize,
+                const GradType& gradient)
+    {
+      // Accumulate gradient.
+      meanSquaredGradient *= parent.Rho();
```
zoq: Again maybe this was intentional, but personally I would have used `parent.rho`.

include/ensmallen_bits/config.hpp
conradsnicta: I'd suggest to have a dedicated `ens_deprecated` with its own declarations. The `arma_deprecated` define is internal to Armadillo and subject to change and/or removal.

Though you can certainly copy'n'paste the current declarations from Armadillo into ensmallen.

(Sidenote: at a future point I may clean up Armadillo and bump the minimum C++ version to C++14 or C++17, which have the [`[[deprecated]]`](https://en.cppreference.com/w/cpp/language/attributes/deprecated) statement builtin. This would make `arma_deprecated` redundant and subject to removal.  Keeping compatibility with C++98/C++03 is starting to get annoying; it's kept to allow use on the ancient RHEL 6, and R folks using Windoze with [fossilised tool chains](https://cran.r-project.org/bin/windows/Rtools/)).

rcurtin:
@conradsnicta I'm totally with you in terms of general release policy ideas.  (I didn't used to be, some years ago.  But I'd like to say I've learned a few things.  That may even be true. :))  But given that I am expecting callbacks to be rather soon, I don't want to release 2.x and then 3.x just a few weeks later.  If callbacks end up looking like they will take way longer than planned, then we can release 2.x with just templatized `Optimize()` and then 3.x once callbacks are ready.

include/ensmallen_bits/ada_delta/ada_delta_update.hpp
rcurtin: Oh, hey, I learned something today:

```
"A nested class is a member and as such has the same access rights as any other member. The members of an enclosing class have no special access to members of a nested class; the usual access rules shall be obeyed."
```

I'll change things back. :)

include/ensmallen_bits/ada_delta/ada_delta.hpp
```diff
@@ -87,10 +87,14 @@ class AdaDelta
    * @param iterate Starting point (will be modified).
    * @return Objective value of the final point.
    */
-  template<typename DecomposableFunctionType>
-  double Optimize(DecomposableFunctionType& function, arma::mat& iterate)
+  template<typename DecomposableFunctionType,
+           typename MatType,
```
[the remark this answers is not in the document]
rcurtin: Ah, no, it was just an oversight, thanks.  Fixed in d7cb084.

include/ensmallen_bits/config.hpp
rcurtin: Actually, it turns out when you filter down the Armadillo detection mechanism it simplifies to basically what I had plus your bits for gcc and clang.  :)

(Sidenote response: C++14 might be too fast, maybe C++11 is better to aim for?  It seems like the "big three" (gcc, clang, and MSVC) all have reasonably complete C++14 support though, so maybe there is no problem there.)

include/ensmallen_bits/iqn/iqn_impl.hpp
rcurtin: Agreed---in this case, it is known that the number of elements will remain unchanged.

zoq:
I figured we should run the static analyses job before we merge this in as well.

zoq:
Wrong PR... I'll remove this one.

rcurtin:
The failure here in the tests should be resolved by #118.

include/ensmallen_bits/cne/cne_impl.hpp
```diff
@@ -104,7 +104,7 @@ typename MatType::elem_type CNE::Optimize(ArbitraryFunctionType& function,
       << std::endl;
 
   // Find the fitness before optimization using given iterate parameters.
-  ElemType lastBestFitness = function.Evaluate(iterate);
+  size_t lastBestFitness = function.Evaluate(iterate);
```
zoq: I think using ``ElemType``makes sense, the main issue is that the optimisation process stops if the current and previous fitness/loss is the same:

https://github.com/mlpack/ensmallen/blob/9d2a5f07c63e292d116c5b182a7682615798f04e/include/ensmallen_bits/cne/cne_impl.hpp#L108

Depending on the problem this is more or less likely. As a workaround I used ``-1`` as threshold in https://github.com/mlpack/ensmallen/pull/119/commits/2eda0cb86ba661e7072c27795581a89da17209f6, if we do this we can use a much lower population size and number of iterations as well.

rcurtin: Ah, ok.  Since that commit is in #119, should we merge this as-is despite the failing build and then merge #119 just afterwards?

zoq: Are you talking about merging #113 as-is? Not sure, have to double check, but I think #119 does contain every commit from this PR, so I guess, it would be fine to just merge #119, once it is done (the ``StepTake`` additions is almost done)?

rcurtin: Oh, I see---yeah, we should just merge #119 all at once then.  Maybe I made your life a little hard with the merge 3c72dae though; let me know if you want me to reproduce that merge on the `callbacks` branch (and cherry-pick a8a2ff0).

zoq: The last time I just used cherry-pick on your branch, so I think I can do the same for the latest commits, but let's see.

coatless:
@rcurtin + @zoq probably need to tweek the @mlpack-bot to avoid closing mlpack org member PRs?

rcurtin:
Nah, actually in this case #119 supersedes this one so we can let it close.  mlpack-bot did our job for us this time :)

Issue #114, opened 2019-05-20
 `shuffle` parameter in `spsa` is not used


coatless:
#### Issue description

The `shuffle` parameter in `spsa` is not used.

```
 ../inst/include/ensmallen_bits/spsa/spsa.hpp:120:8: 
warning: private field 'shuffle' is not used [-Wunused-private-field]
  bool shuffle;
       ^
```

#### Your environment

 * version of ensmallen: 1.15.0
 * operating system: macOS
 * compiler: clang 7.0.0
 * version of Armadillo: 9.400
 * any other environment information you think is relevant:

#### Steps to reproduce

Compile ensmallen.

rcurtin:
Since #115 is merged, this is now solved.  Thanks for the PR. :+1:

PR #115, opened 2019-05-20
Fix compilation warnings

coatless:
Addresses the variable initialization issue highlighted under -Wreorder, c.f.

```
 ../inst/include/ensmallen_bits/qhadam/qhadam_update.hpp:57:5: 
warning: field 'v2' will be initialized after field 'iteration' [-Wreorder]
```

@rcurtin + @zoq  I can tackle #114 if need be in this as well.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+### ensmallen 1.15.1
+###### 2019-MM-DD
+  * Fix -Wreorder in `qhadam` warning (#115).
+  * Fix -Wunused-private-field warning in `spsa` (#115).
+
 ### ensmallen 1.15.0
 ###### 2019-05-14
   * Added QHAdam and QHSGD optimizers (#81).
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -1636,7 +1636,7 @@ differences along stochastic directions.
 
 #### Constructors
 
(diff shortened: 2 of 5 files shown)


zoq:
Thanks for the report; have to set up the style checks, I think I said that before. Anyway, if you like let's include a fix for #114 here as well. Just checked and I think it's save to remove the shuffle parameter.

coatless:
@zoq added into this PR the removal of `shuffle` parameter in `spsa` highlighted by #114. 

In another PR, I'll likely add the flags to the CMAKE Build here:

https://github.com/mlpack/ensmallen/blob/master/CMakeLists.txt#L36-L42

From the R docs, the flags I'll add are: `-Wall -pedantic`

https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Writing-portable-packages

rcurtin:
I realized we have to update `doc/optimizers.md` to remove the `shuffle` parameter for SPSA but that's easy and we can do it during merge.  (Or if you want @coatless you can do it now, up to you.)

coatless:
@rcurtin removed the `shuffle` parameter from the function usage documentation.

**Note:** It wasn't included in the table of parameters.

rcurtin:
> Note: It wasn't included in the table of parameters.

Yeah, only in the constructor, surprising we missed that during review.  Thanks for fixing it. :+1:

zoq:
Let's see if that fixes the style issues in https://github.com/mlpack/ensmallen/pull/116.

PR #116, opened 2019-05-20
Enable minimal code style checks

coatless:
When building ensmallen on TravisCI, verify the code passes traditional ISO checks given by:

```
-Wall -Wpedantic -Wextra
```

If not, treat any warnings that arise as errors with `-Werror`. Show the errors clearly by colorizing the diagnostic output using `-fdiagnostics-color`.

These checks are generally used by CRAN:

https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Writing-portable-packages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_install:
   - cmake . && make && sudo make install && cd ..
 
 install:
-  - mkdir build && cd build && cmake .. && make -j2
+  - mkdir build && cd build && cmake -DSHOW_COMPILE_WARNINGS=ON .. && make -j2
 
 script:
   - CTEST_OUTPUT_ON_FAILURE=1 travis_wait 30 ctest -j2
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_install:
   - cmake . && make && sudo make install && cd ..
 
 install:
-  - mkdir build && cd build && cmake -DSHOW_COMPILE_WARNINGS=ON .. && make -j2
+  - mkdir build && cd build && cmake -DCMAKE_CXX_FLAGS="-Werror" -DCMAKE_C_FLAGS="-Werror" .. && make -j2
 
 script:
   - CTEST_OUTPUT_ON_FAILURE=1 travis_wait 30 ctest -j2
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ cmake_minimum_required(VERSION 2.8.10)
 project(ensmallen C CXX)
 
 option(USE_OPENMP "If available, use OpenMP for parallelization." ON)
+option(SHOW_COMPILE_WARNINGS "Verify compilation is free from warnings." OFF) 
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/CMake")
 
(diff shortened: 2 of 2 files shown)


coatless:
@zoq here is an implementation of the warning checks discussed in #115. These will only be enabled on TravisCI as I've set a `SHOW_BUILD_WARNINGS`. 

**Note:** For the initial go, I've upped the compilation standard to error on warnings (`-Werror`). This probably can be relaxed, but the `-Wall -Wpedantic` should be retained.

coatless:
The initial error was from Travis' compiler being less than gcc 4.9, which first included `-fdiagnostics-color`. To resolve this, I've removed it. 

The next error occurs because #115 needs to be merged. (It's detecting those problems).

.travis.yml
```diff
@@ -18,7 +18,7 @@ before_install:
   - cmake . && make && sudo make install && cd ..
 
 install:
-  - mkdir build && cd build && cmake .. && make -j2
+  - mkdir build && cd build && cmake -DSHOW_COMPILE_WARNINGS=ON .. && make -j2
```
rcurtin: Hmm, could we simplify by just calling `cmake -DCMAKE_CXX_FLAGS='-Wall -Wpedantic -Wextra -Werror' ..`?  Or is there a compelling reason to add the `SHOW_COMPILE_WARNINGS` option? (maybe it makes life easier downstream?)

CMakeLists.txt
```diff
@@ -41,6 +42,15 @@ else ()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unknown-pragmas")
 endif ()
 
+# Verify the code conforms to ISO standards regarding C++ as
+# indicated in the GCC options
+# https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
+# Moreover, treat any compilation warning as an error.
+if ( SHOW_COMPILE_WARNINGS )
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wpedantic -Wextra -Werror")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wpedantic -Wextra -Werror")
```
conradsnicta: Be careful with using `-Wextra`. It can produce overly pedantic warnings.  Coupling this with `-Werror` could get annoying real fast.  Perhaps a selected subset of the warnings by enabled by `-Wextra` can be useful. Examples: `-Wunused-parameter` and `-Wunused-but-set-parameter`.  I'd recommend to also enable `-Wshadow`

coatless: @conradsnicta will do. Thanks for the suggestion!

.travis.yml
coatless: Definitely. Though, how would you feel if I just mirrored MLPACK's setup?

https://github.com/mlpack/mlpack/blob/3389a40ecd686a7e1e01a63b475b042e804440ff/CMakeLists.txt#L65-L78

rcurtin: That's totally fine with me.  We can use the warning flags on every build if you like (except `-Werror` I would say, I would agree with Conrad that that can be really annoying).  But perhaps it is  good to enable `-Werror` just for the Travis build (via `CMAKE_CXX_FLAGS`) so we avoid committing any code that causes warnings.

coatless:
After merging in PR #115 that corrected the compile warnings, this PR builds nicely.

The flags set now are:

```
-Wall -Wpedantic -Wunused-parameter
```

MLPACK uses `-Wall -Wextra`.

PR #118, opened 2019-05-27
Clean up tests (runtime and failure probability)

rcurtin:
I noticed that the tests took a long time, so I investigated and found that the CNE test can take a really long time.  So, I switched it to optimize simpler functions.  Then, I checked to make sure that the test failure probability was low, and had to tune a couple tests.

But probably the more interesting thing I did here was that I noticed the `BigBatchSGD` test sometimes failed after taking a `NaN` step size.  After some investigation, I traced this down to the estimation of the curvature---when the step size gets really small, there is no noticeable different in the iterates, and so `arma::norm(iterate - lastIterate, 2) == 0`, and the curvature estimate divides by that.

Therefore, the solution I settled on was just to avoid step size decay in this situation by taking the curvature estimate to be 0.  @manish7294 I know you were working with `BigBatchSGD` most recently---let me know what you think of this fix or if something better should be done.

### DIFF
--- a/include/ensmallen_bits/bigbatch_sgd/adaptive_stepsize.hpp
+++ b/include/ensmallen_bits/bigbatch_sgd/adaptive_stepsize.hpp
@@ -126,20 +126,24 @@ class AdaptiveStepsize
 
     // Update sample variance & norm of the gradient.
     sampleVariance = vB;
-    gradientNorm = std::pow(arma::norm(gradient / backtrackingBatchSize, 2), 2.0);
+    gradientNorm = std::pow(arma::norm(gradient / backtrackingBatchSize, 2),
+        2.0);
 
-    // Compute curvature.
-    double v = arma::trace(arma::trans(iterate - iteratePrev) *
-        (gradient - gradPrevIterate)) /
-        std::pow(arma::norm(iterate - iteratePrev, 2), 2.0);
+    // Compute curvature.  If it can't be computed (typically due to floating
+    // point representation issues), call it 0.  If the curvature is 0, the step
+    // size will not decay.
+    const double vNum = arma::trace(arma::trans(iterate - iteratePrev) *
+        (gradient - gradPrevIterate));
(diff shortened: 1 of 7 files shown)


include/ensmallen_bits/bigbatch_sgd/adaptive_stepsize.hpp
```diff
@@ -126,20 +126,24 @@ class AdaptiveStepsize
...
-    double v = arma::trace(arma::trans(iterate - iteratePrev) *
-        (gradient - gradPrevIterate)) /
-        std::pow(arma::norm(iterate - iteratePrev, 2), 2.0);
+    // Compute curvature.  If it can't be computed (typically due to floating
+    // point representation issues), call it 0.  If the curvature is 0, the step
+    // size will not decay.
+    const double vNum = arma::trace(arma::trans(iterate - iteratePrev) *
+        (gradient - gradPrevIterate));
+    const double vDenom = std::pow(arma::norm(iterate - iteratePrev, 2), 2.0);
+    const double v = (vNum == 0.0 && vDenom == 0.0) ? 0.0 : (vNum / vDenom);
```
conradsnicta: From purely numerical perspective (not taking into account the nature of the algorithm), there's the possibility of `vNum / 0 = Inf`, so maybe this would be more robust:  
`const double v = (vDenom == 0.0) ? 0.0 : (vNum / vDenom);`

Alternatively, to take into account more possible corner cases, perhaps the check should be more general:  
`const double v_tmp = vNum / vDenom;`  
`const double v = std::is_finite(v_tmp) ? v_tmp : 0.0;`

rcurtin: I think the only way we could have such a situation would be if floating-point errors caused `arma::norm(iterate - iteratePrev, 2)^2` to be 0, but `iterate - iteratePrev` was not 0.  In any case, yeah, probably good to handle that, and I think your solution that any non-finite vNum / vDenom should result in us taking `0.0` instead.

PR #49, opened 2018-11-12
Callback idea

zoq:
Initial idea to handle callbacks, perhaps this is helpful to continue the discussion in https://github.com/mlpack/mlpack/issues/1481

Issue #3, opened 2018-10-05
mlpack bits (logging)


zoq:
In a couple of files we use mlpack's logging capabilities (``Log::Info, Log::Warn``). Are we going to use: ``arma::get_cout_stream()`` and ``arma::get_cerr_stream()``. Any thoughts?

conradsnicta:
Both are fine.  `arma::get_cout_stream()` is probably best suited to "normal" and expected output, such as printing progress (eg. while executing an iterative algorithm). `arma::get_cerr_stream()` is best suited for printing warnings and errors.

Alternatively, we can use `cout` and `cerr` directly.

In combination with either of the options above, it might be useful to adapt the simple printing approach used by armadillo.  Internally, armadillo uses macros such as `arma_debug_print` and `arma_debug_warn`, which can be made inactive by the user by defining `ARMA_DONT_PRINT_ERRORS`.

It's probably not a good idea to directly use `arma_debug_print` and `arma_debug_warn`, as these are internal functions. However, it would be easy to create analogues for ensmallen. For example, `ens_debug_warn`, which use optional defines such as `ENS_DONT_PRINT_ERRORS`.

For all the gory details, see armadillo's [debug.hpp](https://gitlab.com/conradsnicta/armadillo-code/blob/9.200.x/include/armadillo_bits/debug.hpp)

rcurtin:
I think that maybe it might be a good idea here to emulate Armadillo's `get_cout_stream()` and `get_cerr_stream()`, so that we can do in some mlpack header, e.g.,

```
ens::set_cout_stream(Log::Info);
ens::set_cerr_stream(Log::Warn);
```

and then for any `Log::Fatal` call we can just throw an exception.

I think that it won't be quite as simple as that; I'll try and find some time in the next day or two to think about if that will actually work.  Something simple is good for now, but in the longer term this may also be related to mlpack#1481, where the user would pass some functor that would have some behavior when they wanted to print the loss or something like this.

rcurtin:
Actually, looks like I wrote my comment too late. :)  71e6979 seems to handle most of this already.  Still I'll think about how this can be hooked up more nicely to the mlpack infrastructure when desired.  Maybe this?

```
#ifdef ENS_CUSTOM_INFO_LOG
Info = ENS_CUSTOM_INFO_LOG
#endif

#ifdef ENS_CUSTOM_WARN_LOG
Warn = ENS_CUSTOM_WARN_LOG
#endif
```

[...]

PR #119, opened 2019-06-02
Callbacks

zoq:
This is based on https://github.com/mlpack/ensmallen/pull/113; see https://github.com/mlpack/ensmallen/pull/49 for more details.


include/ensmallen_bits/utility/arma_traits.hpp
```diff
@@ -0,0 +1,117 @@
...
+  const static bool value = true;
+};
+
+// template<>
+template<typename eT>
+struct IsArmaType<arma::SpRow<eT> >
+{
+  const static bool value = true;
+};
+
```
conradsnicta: trait for plain subview is probably also required:

    template<typename eT>
    struct IsArmaType< arma::subview<eT> >
    {
      const static bool value = true;
    };

include/ensmallen_bits/utility/arma_traits.hpp
```diff
@@ -0,0 +1,117 @@
...
+// template<>
+template<typename eT>
+struct IsArmaType<arma::Cube<eT> >
+{
+  const static bool value = true;
+};
+
+} // namespace ens
+
```
conradsnicta: There's also `subview_cube<eT>`

include/ensmallen_bits/callbacks/timer_stop.hpp
```diff
@@ -0,0 +1,82 @@
...
+ */
+class TimerStop
+{
+ public:
+  /**
+   * Set up the print loss callback class with the width and output stream.
+   *
+   * @param time The timeout in seconds.
+   */
+  TimerStop(const double time) : time(time), reset(false)
```
conradsnicta: I suggest to disambiguate the name of the input variable from the name of the class member variable. Example:

    TimerStop(const double in_time) : time(in_time), reset(false)

Otherwise certain lower-quality compilers may get confused about which variable we're referring to, and we can end up with `time` not being initialised at all.

When in doubt, assume the compiler will do the wrong thing :)

include/ensmallen_bits/callbacks/timer_stop.hpp
```diff
@@ -0,0 +1,82 @@
...
+    if (timer.toc() > time)
+    {
+      Info << "Timer timeout reached; terminate optimization." << std::endl;
+      optimizer->Terminate() = true;
+    }
+  }
+
+ private:
+  //! The timeout in seconds.
+  double time;
```
conradsnicta: Pedantic suggestion to help maintenance: rather than `time`, naming the variable `duration` would be more precise.  It also makes it easier to understand the statement `if (timer.toc() > duration)`.

include/ensmallen_bits/de/de_impl.hpp
```diff
@@ -75,21 +92,21 @@ inline double DE::Optimize(DecomposableFunctionType& function,
         l = arma::as_scalar(arma::randi<arma::uvec>(
```
conradsnicta: @rcurtin - as of Armadillo 8.400, it should be possible to use a simpler form instead of generating a temporary vector: `l = arma::randi<uword>( arma::distr_param(0, populationSize - 1) );`

Ubuntu 18.04 (current LTS) has Armadillo 8.400, while Fedora has more recent versions.  It might be worthwhile to bump the minimum required version to 8.400, as that brings in more functionality compared to 6.500. This includes the many improvements in sparse matrices.

include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -0,0 +1,103 @@
...
+ public:
+  /**
+   * Set up the progress bar callback class with the given width and output
+   * stream.
+   *
+   * @param width Width of the bar.
+   * @param ostream Ostream which receives output from this object.
+   */
+  ProgressBar(const size_t width = 70,
+              std::ostream& output = arma::get_cout_stream()) :
```
conradsnicta: `arma::get_cout_stream()` is available from Armadillo 8.100 onwards.  Currently ensmallen requires Armadillo 6.500 or later. I suggest bumping the minimum required version to 8.400, as that's available in Ubuntu 18.04 (current LTS).

include/ensmallen_bits/callbacks/print_loss.hpp
```diff
@@ -0,0 +1,57 @@
...
+ */
+class PrintLoss
+{
+ public:
+  /**
+   * Set up the print loss callback class with the width and output stream.
+   *
+   * @param ostream Ostream which receives output from this object.
+   */
+  PrintLoss(std::ostream& output = arma::get_cout_stream()) : output(output)
```
conradsnicta: `arma::get_cout_stream()` is in Armadillo 8.100 onwards.  See comment for include/ensmallen_bits/callbacks/progress_bar.hpp
https://github.com/mlpack/ensmallen/pull/119/commits/b6edf29b5862decf2ce02d534cf1de1ec7325e21#r290214017

include/ensmallen_bits/callbacks/progress_bar.hpp
zoq: Okay, so I guess we missed this here: https://github.com/mlpack/ensmallen/blob/8bea8d214b40be3cb42e817328c0791541fbcd6c/include/ensmallen_bits/log.hpp#L74

I don't mind to bump the version here, what I could also do is to backport the functionality since it's just simple enough. Personally I don't have a strong opinion here, we would have to bump the version for mlpack as well @rcurtin do you have any preference?

rcurtin: Ahh, unfortunately the reason we are stuck with 6.500 is Ubuntu 16.04 (even RHEL6 has 7.300, although I'm less concerned about RHEL6).  16.04 is still in pretty wide usage---the more recent LTS version, 18.04, has "only" been out for a year, which in my experience is not enough time for awful slow bureaucratic IT departments to upgrade.  If it's easy to backport for us, then I'd suggest that because it will make it easier for our users (who likely already have Armadillo installed or, if they are in one of these awful IT situations, they can ask the sysadmin to install it via the package manager).

In this case, I wonder if we'd be better off just defining an inline function whose implementation depends on the version of Armadillo available... if 8.100 or newer is available, it returns `arma::get_cout_stream()`, otherwise it returns `std::cout`.  What do you think?

include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -0,0 +1,103 @@
...
+                const MatType& /* coordinates */,
+                const size_t epoch,
+                const double /* objective */)
+  {
+    if (!optimizer->MaxIterations())
+    {
+      Warn << "Maximum number of iterations not defined (no limit),"
+           << " no progress bar shown." << std::endl;
+      return;
+    }
```
rcurtin: Do you think it would be better to have one progress bar per epoch, like Keras does?  In this case we would need an additional `StepTaken()` callback available (or something like this).

include/ensmallen_bits/callbacks/timer_stop.hpp
```diff
@@ -0,0 +1,82 @@
...
+  void EndEpoch(OptimizerType& optimizer,
+                FunctionType& /* function */,
+                const MatType& /* coordinates */,
+                const size_t /* epoch */,
+                const double /* objective */)
+  {
+    if (timer.toc() > time)
+    {
+      Info << "Timer timeout reached; terminate optimization." << std::endl;
+      optimizer->Terminate() = true;
```
rcurtin: Another idea would be to return a `bool` instead indicating whether optimization should continue.  That might allow us to avoid adding a `terminate` field to every optimizer.

include/ensmallen_bits/function.hpp
```diff
@@ -84,28 +105,196 @@ class Function :
...
+
+  /* Callbacks entry points. */
+
+  /**
+   * Register the optimizer and the given callbacks.
+   *
+   * @param optimizer reference to the optimizer.
+   * @param callbacks The functions to call.
+   */
+  void Register(OptimizerType optimizer)
```
rcurtin: Should this be a reference?

include/ensmallen_bits/utility/any.hpp
```diff
@@ -5,6 +5,11 @@
  * A simple and dangerous implementation of 'Any' that can be used when the
  * class needs to hold some specific information for which the type is not
  * known.
+ *
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
```
rcurtin: Thanks, I forgot this! :+1:

include/ensmallen_bits/function.hpp
```diff
@@ -15,7 +16,10 @@
 
 namespace ens {
 
-template<typename FunctionType, typename MatType, typename GradType>
+template<typename FunctionType,
+         typename MatType,
+         typename GradType,
+         typename OptimizerType>
```
rcurtin: Do you think it would be easier to have a standalone `Callbacks` class that has static methods of the form

```
  template<typename OptimizerType, typename FunctionType, typename MatType, typename... CType>
  static bool
  Evaluate(const OptimizerType& optimizer,
           const FunctionType& function,
           const MatType& coordinates,
           size_t begin,
           size_t batchSize,
           CType&... callbacks);
```

or something like that?  This would separate the callback infrastructure from the infrastructure that adds missing functions, and might be a little easier to understand.

include/ensmallen_bits/callbacks/progress_bar.hpp
conradsnicta: `arma::get_cout_stream()` has been used in ensmallen since 1.10, and so far nobody has complained :)  Maybe this indicates that Ubuntu 18.04 has spread sufficiently.

Using `std::cout` directly is problematic for the folks in the R world (ie. RcppEnsmallen and RcppArmadillo packages @coatless @eddelbuettel).  `arma::get_cout_stream()` has a configurable mechanism used by Rcpp -- allows the use of an R specific output stream.  

Otherwise any 3rd party R packages relying on `std::cout` tend to be rejected by the CRAN gatekeepers. This in turn can reduce the impact of ensmallen.

rcurtin: Oh, good point, you're right and I forgot about that.  This is in `log.hpp`:

```
#ifdef ENS_PRINT_INFO
  static std::ostream& Info = arma::get_cout_stream();
#else
  static NullOutStream Info;
#endif

#ifdef ENS_PRINT_WARN
  static std::ostream& Warn = arma::get_cerr_stream();
#else
  static NullOutStream Warn;
#endif
```

So, I think already we are in a situation where if `ENS_PRINT_INFO` or `ENS_PRINT_WARN` are defined, Armadillo 8.100 or newer is required.  I looked briefly at backporting Armadillo's `get_cout_stream()` and I think it's easy*---it won't require anything other than the following code dumped into the `arma` namespace, only when the version is below 8.100:

```
template<typename T>
inline
std::ostream&
arma_cout_stream(std::ostream* user_stream)
  {
  static std::ostream* cout_stream = &(ARMA_COUT_STREAM);

  if(user_stream != NULL)  { cout_stream = user_stream; }

  return (*cout_stream);
  }



template<typename T>
inline
std::ostream&
arma_cerr_stream(std::ostream* user_stream)
  {
  static std::ostream* cerr_stream = &(ARMA_CERR_STREAM);

  if(user_stream != NULL)  { cerr_stream = user_stream; }

  return (*cerr_stream);
  }



inline
void
set_cout_stream(std::ostream& user_stream)
  {
  arma_cout_stream<char>(&user_stream);
  }



inline
void
set_cerr_stream(std::ostream& user_stream)
  {
  arma_cerr_stream<char>(&user_stream);
  }



inline
std::ostream&
get_cout_stream()
  {
  return arma_cout_stream<char>(NULL);
  }



inline
std::ostream&
get_cerr_stream()
  {
  return arma_cerr_stream<char>(NULL);
  }
```

So, perhaps that's the best way to go.  I don't have a problem with the backporting, but also we're backed into a little bit of a corner here by this situation.  So I also don't have a huge problem with bumping the minimum version to 8.100.0.  @zoq up to you.

* ignoring very strange situations that *could* happen if the user has a very weird setup.

conradsnicta: In general I'm not a big fan of backporting, as that creates franken-software that nobody has tested properly. In this particular case the changes are relatively small so it might be doable, though I'd still prefer bumping the minimum required version.

coatless: I'm with @conradsnicta regarding backporting. I think it would be better to bump the required version since there are a lot of nice things in 8.yyy line (needless to say, also a licensing change). 

Also, as @conradsnicta mentioned regarding CRAN, any `std::cout` or `std::cerr` in the code will trigger an auto rejection. Details can be found at the ol' logging post: https://github.com/mlpack/ensmallen/issues/3#issuecomment-443578471

include/ensmallen_bits/callbacks/timer_stop.hpp
zoq: True, this replicated keras and I guess also is shorted since we don't have to write ``terminate |= Callback(...);`` for each callback. But I can see the benefit, to avoid the extra ``Terminate`` function, so not sure if you like to go with the ``bool`` I make the change.

include/ensmallen_bits/function.hpp
zoq: My intention was to reduce the number of lines necessary to use the callbacks, using an extra class would mean one extra function for ``Evaluate`` and `´Gradient``, but I guess since there are some optimisers that don't use the ``Function`` class it might be easier to just go with an extra class? Let me know what you think.

include/ensmallen_bits/callbacks/progress_bar.hpp
zoq: Thanks for the input, will bump the version.

zoq:
> Do you think it's worth adding a few more callback hooks? I had a list of proposed hooks in #49, but we can also put that off until after this is merged and add those later.

Absolutely, just wanted to make sure we can agree on the API first (``Function`` or extra class).

> I think we should test the callbacks, but that shouldn't be too hard.

Sure, I wrote some here: https://github.com/mlpack/ensmallen/blob/a82a4aa256467b11b7e4d2a0bbe254b7440b5a3f/tests/callbacks_test.cpp

> Right now the callbacks are only implemented for SGD (and anything derived from that class). Are you planning to add them to all classes here, or in a future PR?

Already, did this for the other optimisers as well, once we agreed on the API, I will push the changes.

> We should add some documentation (maybe a doc/callbacks.md) for the callback infrastructure that shows users how to use the callbacks we already have, and how to implement their own.

Working on that part right now.

include/ensmallen_bits/callbacks/timer_stop.hpp
rcurtin: Yeah, personally I think the `bool` approach is better here, since it doesn't require us to add a member to every optimizer.

include/ensmallen_bits/function.hpp
rcurtin: Hmm, I think in this case maintainability is a little easier if we separate out the callbacks and function-adding functionality.  And you're right, not all optimizers use `Function`.  In fact, right now, I think it is only needed or useful for differentiable functions or differentiable separable functions.

tests/callbacks_test.cpp
```diff
@@ -0,0 +1,382 @@
...
+  arma::mat coordinates = f.GetInitialPoint();
+
+  // Instantiate the optimizer with a number of iterations that will take a
+  // long time to finish.
+  StandardSGD s(0.0003, 1, 10000000000, -10, true);
+
+  // The optimization process should return in one second.
+  const double result = s.Optimize(f, coordinates, EarlyStopAtMinLoss(100));
+
+  REQUIRE(s.Terminate());
```
rcurtin: Ah, I guess if we remove `Terminate()` then this test gets harder.  I guess that we could use a really simple quadratic function to optimize, and start very far from the optimum with a very small step size.  Then, if termination happens very early, then we would expect the point to be far closer to the starting point than to the optimum.

tests/callbacks_test.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+  std::stringstream stream;
+  s.Optimize(f, coordinates, ProgressBar(10, stream));
+
+  REQUIRE(stream.str().length() > 0);
+}
+
+/**
+ * Make sure the StoreBestCoordinates callback will store the best coordinates
+ * and objetive.
```
rcurtin: Typo: `objective` :)

include/ensmallen_bits/callbacks/progress_bar.hpp
zoq: Do you think this could be an option? I think depending on the number of iterations the output could be usage. Also can you clarify the ``StepTake()`` idea, not sure I can follow.

rcurtin: So, basically, if we want to have a per-epoch progress bar, we would need to know how far into the epoch we are.  If we had a `StepTaken()` callback that indicated when a step was taken, what the (partial?) objective was, how big the batch size was, etc., we could know how much of the epoch was done and be able to print the progress bar.

include/ensmallen_bits/de/de.hpp
```diff
@@ -137,6 +138,9 @@ class DE
 
   //! The tolerance for termination.
   double tolerance;
+
+  //! Controls early termination of the optimization process.
+  bool terminate;
```
rcurtin: I think the class member is not needed here; if I understand right, it could be done as just a local variable in the `Optimize()` method.  Let me know if I missed something.

include/ensmallen_bits/gradient_descent/gradient_descent_impl.hpp
```diff
@@ -26,31 +26,55 @@ inline GradientDescent::GradientDescent(
...
 
   // Now iterate!
-  arma::mat gradient(iterate.n_rows, iterate.n_cols);
-  for (size_t i = 1; i != maxIterations; ++i)
+  terminate |= Callback::BeginOptimization(*this, f, iterate, callbacks...);
+  BaseGradType gradient(iterate.n_rows, iterate.n_cols);
+  for (size_t i = 1; i != maxIterations && !terminate; ++i)
   {
+    terminate |= Callback::BeginEpoch(*this, f, iterate, i, overallObjective,
+        callbacks...);
```
rcurtin: Hmm, so it looks like we are using `BeginEpoch()` and `EndEpoch()` even when it the function is not separable.  This could be a little confusing, since people don't generally think the word `epoch` when they think about gradient descent (in fact, I suspect the term `epoch` is generally unknown outside of deep learning optimization).  I have a couple ideas:

(1) We could document clearly in the callback documentation that for non-separable functions, `BeginEpoch()` and `EndEpoch()` are called at the beginning and end of a single *iteration*.  (This works, since a non-separable optimizer is just like a separable optimizer except with `batchSIze = NumFunctions()`.)

(2) We could avoid the use of `BeginEpoch()` and `EndEpoch()` in non-separable optimizers, and instead have a `StepTaken()` callback only.  The reason I suggest `StepTaken()` and not `EvaluateWithGradient()` is that some non-separable optimizers (and some separable optimizers too) will propose multiple possible steps before actually taking one.  That can mean that `Evaluate()` or `EvaluateWithGradient()` is called potentially more than once per iteration.

Let me know what you think; maybe there are other options I didn't think of.

tests/callbacks_test.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+/**
+ * Make sure we invoke all callbacks (AdaDelta).
+ */
+TEST_CASE("AdaDeltaCallbacksFullFunctionTest", "[CallbacksTest]")
+{
+  AdaDelta optimizer(1.0, 1, 0.99, 1e-8, 3, 1e-9, true);
+  CallbacksFullFunctionTest(optimizer, true, true, true, true, true, true,
+      false, false);
+}
```
rcurtin: Nice, this is a good way to test that all callbacks are called, and it's easy to add this test for new optimizers.

CMakeLists.txt
```diff
@@ -41,6 +41,13 @@ else ()
...
+# Set the CFLAGS and CXXFLAGS depending on the options the user specified.
+if(CMAKE_COMPILER_IS_GNUCC OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wpedantic -Wunused-parameter")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wpedantic -Wunused-parameter")
+endif()
+
 # The only dependency we need is Armadillo.
 #
 # We keep the minimum version in sync with mlpack, otherwise we could have
```
rcurtin: I guess we should bump the version here still.  If we bump to 8.400.0 I think the RHEL7 package will need to be updated, but actually I think that's my job so I can do it...

doc/callbacks.md
```diff
@@ -0,0 +1,484 @@
...
+
+| **type** | **name** | **description** |
+|----------|----------|-----------------|
+| `OptimizerType` | **`optimizer`** | The optimizer used to update the function. |
+| `FunctionType` | **`function`** | The function to be optimized. |
+| `MatType` | **`coordinates`** | The current function parameter. |
+| `size_t` | **`epoch`** | The index of the current epoch. |
+| `double` | **`objective`** | Objective value of the current point. |
+
+## Built-in Callbacks
```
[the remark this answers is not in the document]
conradsnicta: I'm with Ryan on this.  From an "average use case" perspective, I think it would be better to document the built-in callbacks first, and then explain how to make customised callbacks.

doc/callbacks.md
```diff
@@ -0,0 +1,484 @@
+Callbacks in mlpack are methods that are called at different states during the
```
conradsnicta: I suggest changing "different" to "various".  

In this context the word "different" is grammatically imprecise at best, and too overloaded in meaning at worst (eg. "different states" can raise questions like "how are the states different?", or "what are the states different to?", etc).  
Generally the word "different" should only be used to mean "not the same".

include/ensmallen_bits/gradient_descent/gradient_descent_impl.hpp
zoq: For the second idea, do you still call the Evaluate/Gradient callback?

rcurtin: I think we can leave the Evaluate/Gradient callbacks---they might still be useful if, e.g., you want to count the number of function evaluations or gradient evaluations, I guess.

include/ensmallen_bits/problems/generalized_rosenbrock_function.hpp
```diff
@@ -90,9 +96,10 @@ class GeneralizedRosenbrockFunction
...
    */
-  void Gradient(const arma::mat& coordinates,
+  template<typename MatType, typename GradType>
+  void Gradient(const MatType& coordinates,
                 const size_t begin,
-                arma::mat& gradient,
+                GradType& gradient,
                 const size_t batchSize = 1) const;
 
   /**
```
rcurtin: I think that removing this overload should fix the static code analysis build.

include/ensmallen_bits/callbacks/progress_bar.hpp
zoq: Sorry for the slow response, I can get the objective from the last Evaluate call, but I don't see an easy way to get the batch size, unless the optimizer provides an option. Maybe I missed something.

rcurtin: Right, sorry---the batch size doesn't make sense to pass, because it's optimizer-specific.  If the user wants to know the batch size (and they are writing the callback) they can just call `BatchSize()` from inside the callback.  For a progress bar we could do that (if a `BatchSize()` function exists and the optimizer is a separable function, which we can check with traits I guess).  Let me know what you think.

zoq: Okay, just wanted to make sure I didn't missed anything.

rcurtin:
If the BBSGD tests are failing and don't seem to be a part of this PR, we can open a separate issue for that.  No need to hold this up on those; let me know what you think. :+1:

tests/pso_test.cpp
```diff
@@ -17,28 +17,44 @@ using namespace ens;
...
+  LBestPSO s;
 
+  arma::fmat coords = f.GetInitialPoint<arma::fmat>();
+  if (!s.Optimize(f, coords))
+    FAIL("LBest PSO optimization reported failure for Sphere Function.");
+
+  double finalValue = f.Evaluate(coords);
+  REQUIRE(finalValue <= 1e-5);
+  for (size_t j = 0; j < 4; ++j)
+    REQUIRE(coords[j] <= 1e-3);
```
conradsnicta: I recommend using the [element accessor](http://arma.sourceforge.net/docs.html#element_access) with bounds check here and in other code like this, ie. `coords(j)` instead of `coords[j]`.

Since this is a test case, we should catch all possible errors.  Using `[]` allows some errors to slip through.

zoq: Great, will apply the patch.

PR #121, opened 2019-08-04
Parallelization for PSO

SuryodayBasak:
I'm creating this pull request for the purpose of merging changes that involve parallelization in our PSO implementation.


include/ensmallen_bits/pso/pso.hpp
```diff
@@ -87,6 +88,7 @@ class PSOType
           const double impTolerance = 1e-10,
           const double exploitationFactor = 2.05,
           const double explorationFactor = 2.05,
+	  const size_t numThreads = 4,
```
zoq: A user could use the environment variable ``OMP_NUM_THREADS`` to adjust the number of threads, without recompiling the code.

SuryodayBasak: I was hoping to pass the number of threads that a user would like to use as an argument.

SuryodayBasak:
I've included some basic multithreading taking the number of threads in as an argument. Strangely, if I use a really large number of threads, such as 10 threads, as opposed to using a single thread, the execution actually slows down. Maybe it has something to do with synchronization?

I still have many other blocks to parallelize, and I will work on them shortly.

include/ensmallen_bits/pso/pso.hpp
zoq: The issue I see with overriding ``OMP_NUM_THREADS`` is that it could cause confusion with the mlpack codebase, as the method behaves differently than other methods. Is there any advantage to use a parameter instead of using the env?

zoq:
OpenMP does provide some options to specify which parameter should be shared, that might be helpful here.

include/ensmallen_bits/pso/pso.hpp
SuryodayBasak: No, technically there isn't. But I guess the only thing that I had in mind while suggesting this was a certain level of user-friendliness by specifying the number of threads as a parameter. But I think that you're right about this:

> The issue I see with overriding OMP_NUM_THREADS is that it could cause confusion with the mlpack codebase, as the method behaves differently than other methods.

So I guess it is better to change the API design and make things clear in the documentation about specifying the number of threads using the environment variable.

zoq: Agreed, we should definitely point this out in the documentation.

SuryodayBasak:
I have made some progress and observations. Firstly, I have removed the `numThreads` argument that I was using earlier. Instead, I have used the directive `#pragma omp parallel for` with all the blocks that I want to parallelize.

The second thing that I have been able to gather is that parallelization comes at the cost of synchronization and this can be a big problem in many cases wherein a lot of synchronization needs to be done. With PSO, each parallelizable `for` loop under the update rules is called multiple times, and I believe that does add some overhead. I have also observed that the right number of threads is something that should be set only based on the application - in some cases, a small number of threads (such as 3 or 4 threads) leads to the lowest number of clock cycles used to execute the program! I think that is interesting in itself, but for our purposes at hand, I believe that this is okay since it is a technological limitation.

Let me know your thoughts on this. The `pso_test.cpp` has some extra lines now, which I will remove later, once we've tested the timing more.

include/ensmallen_bits/pso/pso_impl.hpp
```diff
@@ -150,7 +158,9 @@ double PSOType<VelocityUpdatePolicy, InitPolicy>::Optimize(
     // In-place update of particle positions.
     particlePositions += particleVelocities;
 
+    //TODO: Think over this. Probably should not be parallelized.
     //Find the best particle.
+    #pragma omp parallel for
```
zoq: Perhaps it makes sense to calculate ``bestFitness`` afterwards, in an independent run, that way there is no need to synchronize between different runs.

zhlou: What here need is a reduction clause. Since here you essentially need an argmin, there are 2 strategies to do this. The easier way is simply use
```suggestion
    #pragma omp parallel for reduction(min:bestFitness)
```
You can find the `bestParticle` later after the parallelized region. `reduction` is supported in OpenMP since 3.0.

Another way is to define a custom reduction to calculate `bestParticle` and `bestFitness` in one pass by using `declare reduction` directive introduced in OpenMP 4.0. See section 2.15 of https://www.openmp.org/wp-content/uploads/OpenMP4.0.0.pdf

rcurtin: Hey @SuryodayBasak, just checking in---do you think you'll have a chance to handle this?  I think @zhlou's suggestion here is a good one.  It's ok if not, of course; I just want to make sure it's still on the radar if you are interested in finishing it up and getting it merged. :smile:

SuryodayBasak:
Hi all! Could this issue be kept open for another two to three weeks? I'll get back to working on it as soon as my semester is done! :)

SuryodayBasak:
Hi! Please keep this open. I'll get back to it in a few weeks - I'm just
wrapping up my semester now.

*Suryoday Basak*

*Graduate Student*, Department of Computer Science and Engineering, *The
University of Texas at Arlington*

Website: *suryodaybasak.info <http://suryodaybasak.info>*
Follow me on Medium:
*https://medium.com/@suryodaybasak
<https://medium.com/@suryodaybasak>*Astroinformatics
Research Group: *http://astrirg.org <http://astrirg.org>*


* <http://ascl.net/code/v/1475>*

rcurtin:
Of course, no problem!  Good luck with the end of the semester. :)

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -122,6 +124,8 @@ class LBestUpdate
           right(i) : i;
     }
 
+
+    //TODO: This should probably not be parallelized.
```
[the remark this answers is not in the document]
SuryodayBasak: Okay I will try this out in the code.

Issue #123, opened 2019-08-09
PSO variable type buglet


coatless:
#### Issue description

Mismatch on return types in [PSO](https://github.com/mlpack/ensmallen/blob/b46986d1344551b1d8480784661b7c547304ba04/include/ensmallen_bits/pso/pso.hpp)

```
../inst/include/ensmallen_bits/pso/pso.hpp: In member function ‘size_t ens::PSOType<VelocityUpdatePolicy, InitPolicy>::LowerBound() const’:
../inst/include/ensmallen_bits/pso/pso.hpp:162:38: error: cannot convert ‘const vec’ {aka ‘const arma::Col<double>’} to ‘size_t’ {aka ‘long unsigned int’} in return
  162 |   size_t LowerBound() const { return lowerBound; }
      |                                      ^~~~~~~~~~
```

Defined as `arma::vec`:

https://github.com/mlpack/ensmallen/blob/b46986d1344551b1d8480784661b7c547304ba04/include/ensmallen_bits/pso/pso.hpp#L207-L210

But returning `size_t`:

https://github.com/mlpack/ensmallen/blob/b46986d1344551b1d8480784661b7c547304ba04/include/ensmallen_bits/pso/pso.hpp#L161-L171

#### Your environment

 * version of ensmallen: 1.16.0
 * operating system: debian
 * compiler: gcc-9
 * version of Armadillo: 9.600

zoq:
Thanks for the report, should be fixed in https://github.com/mlpack/ensmallen/pull/124, I will put some time into the test setup, so that we can avoid such issues in the future.

rcurtin:
Fixed in #124.  Thanks for the report. :+1:

rcurtin:
Ah, I discovered that #126 will also be necessary here.

coatless:
@rcurtin sorry about that. I should have grabbed the entire log. :/

coatless:
Closed in #126 and #123

PR #124, opened 2019-08-09
Fix PSO bound type

zoq:
Fix for issue reported in https://github.com/mlpack/ensmallen/issues/123.

### DIFF
--- a/include/ensmallen_bits/pso/pso.hpp
+++ b/include/ensmallen_bits/pso/pso.hpp
@@ -40,19 +40,17 @@ namespace ens {
  * For more information, refer to:
  *
  * @inproceedings{Kennedy,
- *    doi = {10.1109/icnn.1995.488968},
- *    url = {https://doi.org/10.1109/icnn.1995.488968},
- *    publisher = {{IEEE}},
- *    author = {J. Kennedy and R. Eberhart},
  *    title = {Particle swarm optimization},
- *    booktitle = {Proceedings of {ICNN}{\textquotesingle}95 - 
- *                 International Conference on Neural Networks}
+ *    author = {J. Kennedy and R. Eberhart},
+ *    booktitle = {Proceedings of {ICNN}{\textquotesingle}95 -
+ *                 International Conference on Neural Networks},
+ *    publisher = {{IEEE}}
  * }
  *
  * PSO can optimize arbitrary functions. For more details, see the documentation
  * on function types included with this distribution or on the ensmallen
  * website.
- * 
+ *
  * For PSO to work, the function being optimized must implement an
  * ArbitraryFunctionType template parameter. The respective class must implement
  * the following function:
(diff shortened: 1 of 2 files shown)


include/ensmallen_bits/pso/pso.hpp
```diff
@@ -159,16 +159,16 @@ class PSOType
   size_t& NumParticles() { return numParticles; }
 
   //! Retrieve value of lowerBound.
-  size_t LowerBound() const { return lowerBound; }
+  arma::vec LowerBound() const { return lowerBound; }
```
lozhnikov: Shouldn't it be a const reference?
```c++
const arma::vec& LowerBound() const { return lowerBound; }
```

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -159,16 +159,16 @@ class PSOType
...
 
   //! Modify value of lowerBound.
-  size_t& LowerBound() { return lowerBound; }
+  arma::vec& LowerBound() { return lowerBound; }
 
   //! Retrieve value of upperBound.
-  size_t UpperBound() const { return upperBound; }
+  arma::vec UpperBound() const { return upperBound; }
```
lozhnikov: See the comment above.

include/ensmallen_bits/pso/pso.hpp
zoq: Right, updated.

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -159,16 +159,16 @@ class PSOType
   size_t& NumParticles() { return numParticles; }
 
   //! Retrieve value of lowerBound.
-  size_t LowerBound() const { return lowerBound; }
+  const arma::vec LowerBound() const { return lowerBound; }
```
lozhnikov: Did you forget to add a reference `&`?
```c++
const arma::vec&
```
I thought the copy-constructor of `arma::vec` isn't trivial.

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -159,16 +159,16 @@ class PSOType
...
 
   //! Modify value of lowerBound.
-  size_t& LowerBound() { return lowerBound; }
+  arma::vec& LowerBound() { return lowerBound; }
 
   //! Retrieve value of upperBound.
-  size_t UpperBound() const { return upperBound; }
+  const arma::vec UpperBound() const { return upperBound; }
```
lozhnikov: See the comment above.

include/ensmallen_bits/pso/pso.hpp
zoq: You are absolutely right, thanks again for pointing this out.

rcurtin:
@lozhnikov thanks for catching the const reference issue; I completely overlooked it. :)

Issue #122, opened 2019-08-09
Release notes


coatless:
When pulling from [`HISTORY.md`](https://github.com/mlpack/ensmallen/blob/master/HISTORY.md), the release notes:

1. miss the "code name" that can be found in `ensmallen.hpp` and [GitHub release page](https://github.com/mlpack/ensmallen/releases).
2. miss hard links to the PR, e.g. `#86` -> `[#86](https://github.com/mlpack/ensmallen/pull/86)`

Would it be possible to always include these components?

rcurtin:
Oh, I see, that's easy enough to do.  Let me manually make those changes now.  I don't think it's worth issuing 1.16.1 with just changes to HISTORY.md, but we can make sure that happens in the future.  Does that sound good?

coatless:
Sounds good! I agree, there is no need to release these history changes as 1.16.1.

rcurtin:
Fixed in #125.  I'll make sure that HISTORY.md follows this format in the future.  To make your life easier, should we do this for mlpack also?

coatless:
@rcurtin I'm not sure how important that would be as the `RcppMLPACK` bridge relies on the system-specific version of `MLPACK` installed whereas the `RcppEnsmallen` package  distributes the header-only library.

rcurtin:
Ah, got it, thanks for the clarification. :+1:

PR #125, opened 2019-08-09
Add version names and use Markdown links in HISTORY.md

rcurtin:
Addresses #122.  @coatless let me know if this is what you were thinking or if you had something else in mind. :)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,113 +1,130 @@
-### ensmallen 1.16.0
+### ensmallen 1.16.1: "Loud Alarm Clock"
+###### 2019-??-??
+  * Update HISTORY.md to use Markdown links to the PR and add release names.
+
+### ensmallen 1.16.0: "Loud Alarm Clock"
 ###### 2019-08-09
   * Add option to avoid computing exact objective at the end of the optimization
-    (#109).
+    ([#109](https://github.com/mlpack/ensmallen/pull/109)).
 
-  * Fix handling of curvature for BigBatchSGD (#118).
+  * Fix handling of curvature for BigBatchSGD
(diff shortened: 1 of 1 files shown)


coatless:
@rcurtin perfect. Thank you.

gauravk97:
Hi Gaurav here, from mlpack issue (https://github.com/mlpack/mlpack/issues/1974). I think in addition to the above changes, the following changes are also required: 
in file pso.hpp

  //! Retrieve value of impTolerance.
  size_t ImpTolerance() const { return impTolerance; }
  //! Modify value of impTolerance.
  size_t& ImpTolerance() { return impTolerance; }

here size_t -> double because impTolerance is of double datatype. Because of this, it was breaking in my system.

rcurtin:
Yep, also needed is #126.  It'll be merged shortly and I'll do another release.  Sorry for the inconvenience.

gauravk97:
@rcurtin I'm sorry for bugging you again and again :P but I have one more question and I hoped you could answer that for me. 
I am using the compiler gcc 8.3.0 while my friend was using gcc 5.4. And we both were building the mlpack library in our systems. In his system, this return type error didn't come up at all and the build executed perfectly even with ensmallen 1.16.0. I was curious how that happened.
Otherwise, with these fixes, even my build is now completing without issues. Thanks a lot. :)

rcurtin:
@gauravk97 no worries---different versions of compilers will handle issues differently.  In earlier gcc's case, apparently it wasn't being strict about the return types because those functions aren't actually used in any of the code you're compiling.  Newer gcc is stricter (which I think is generally a good thing). :)

PR #128, opened 2019-09-07
Add release process and details to the repository

rcurtin:
I thought it would be useful to add the release process and scripts to do it automatically to this repository; that way, anyone can easily and quickly make a release, and I am not a bottleneck. :)

### DIFF
--- a/rel/ensmallen-release.sh
+++ b/rel/ensmallen-release.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+#
+# Release a new version of ensmallen.
+#
+# Arguments:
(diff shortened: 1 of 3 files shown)


rel/ensmallen-release.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
zoq: Hm, should we use ``#!/usr/bin/env bash`` here to be more portable?

rcurtin: Good point, thanks!  Fixed in 083a9a6.

PR #131, opened 2019-09-08
Minor compatibility fixes for ensmallen 1.x

rcurtin:
These simple fixes make it easier for older code to work with ensmallen 2.x.  There's no functionality change (although `GetInitialPoint()` for `LRSDP` had a bug, and this fixes that).


rcurtin:

rcurtin:

rcurtin:
The mlpack master build test will fail until mlpack/mlpack#2011 is merged, so I'll merge this here now, but release ensmallen 2.10.1 with the patch only after mlpack/mlpack#2011 is merged so that everyone's builds don't break.

PR #133, opened 2019-09-11
ensure armadillo version check uses the same numbers as CMakeLists.txt

conradsnicta:
CMakeLists.txt enforces Armadillo 8.400, so for consistency we need to do that in `include/ensmallen.hpp` as well

https://github.com/mlpack/ensmallen/blob/master/CMakeLists.txt#L55


zoq:
Strange haven't seen the "Can one of the admins verify this patch?" message in a long time; let's see what happens if I delete it from the config.

PR #134, opened 2019-09-13
best practice: set bash option to terminate shell upon expected error

barak:
When earlier commands fail, it might not be such a great idea to run later git commands which do things like tagging and pushing.


zoq:
Thanks again.

PR #136, opened 2019-09-16
Check L-BFGS scaling factor for convergence

rcurtin:
While debugging some mlpack test failures I discovered that it is possible that `scalingFactor` for L-BFGS can be computed as `0.0` when the optimization is getting very close to convergence.  This, then, causes a division by 0 later on in the code, and the returned iterate goes to a bunch of NaN garbage.

So, this change addresses that by catching the case where the scaling factor is computed as 0, and terminates successfully in that case.


rcurtin:

PR #137, opened 2019-09-21
AdaBound + AMSBound

zoq:
Implementation of AdaBound and AMSBound which employ dynamic bounds on
learning rates to achieve a gradual and smooth transition from adaptive methods to
SGD. For more information see "Adaptive Gradient Methods with Dynamic Bound of Learning Rate" by Luo, Liangchen and Xiong, Yuanhao and Liu, Yan and Sun, Xu.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,7 @@
 ### ensmallen ?.??.?
 ###### ????-??-??
+  * Add AdaBound and AMSBound
+    ([#137](https://github.com/mlpack/ensmallen/pull/137)).
 
 ### ensmallen 2.10.5
 ###### 2019-12-13
@@ -27,9 +29,6 @@
 
   * Add additional L-BFGS convergence check
     ([#136](https://github.com/mlpack/ensmallen/pull/136)).
-
-  * Add AdaBound and AMSBound
-    ([#137](https://github.com/mlpack/ensmallen/pull/137)).
 
 ### ensmallen 2.10.2: "Fried Chicken"
 ###### 2019-09-11
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+### ensmallen ?.??.?
+###### ????-??-??
+  * Add AdaBound and AMSBound
+    ([#137](https://github.com/mlpack/ensmallen/pull/137)).
+
 ### ensmallen 2.10.2: "Fried Chicken"
 ###### 2019-09-11
   * Add release script to rel/ for maintainers
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -38,6 +38,10 @@ The attributes of the optimizer may also be modified via the member methods
 
 #### Examples
 
+<details open>
+<summary>Click to collapse/expand example code.
+</summary>
+
 ```c++
 SphereFunction f(2);
 arma::mat coordinates = f.GetInitialPoint();
@@ -45,6 +49,8 @@ arma::mat coordinates = f.GetInitialPoint();
 AdaBound optimizer(0.001, 2, 0.1, 1e-3, 0.9, 0.999, 1e-8, 500000, 1e-3);
 optimizer.Optimize(f, coordinates);
 ```
+
+</details>
 
 #### See also:
 
@@ -327,6 +333,10 @@ The attributes of the optimizer may also be modified via the member methods
 
 #### Examples
 
+<details open>
+<summary>Click to collapse/expand example code.
+</summary>
+
 ```c++
 SphereFunction f(2);
 arma::mat coordinates = f.GetInitialPoint();
@@ -334,6 +344,16 @@ arma::mat coordinates = f.GetInitialPoint();
 AMSBound optimizer(0.001, 2, 0.1, 1e-3, 0.9, 0.999, 1e-8, 500000, 1e-3);
 optimizer.Optimize(f, coordinates);
 ```
+
+</details>
+
+#### See also:
+
+ * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
+ * [SGD](#standard-sgd)
+ * [Adaptive Gradient Methods with Dynamic Bound of Learning Rate](https://arxiv.org/abs/1902.09843)
+ * [Adam: A Method for Stochastic Optimization](http://arxiv.org/abs/1412.6980)
+ * [Differentiable separable functions](#differentiable-separable-functions)
 
 ## AMSGrad
 
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -1,3 +1,59 @@
+## AdaBound
+
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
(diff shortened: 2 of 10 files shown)


zoq:

rcurtin:
Wow, a month has gone by already?  It's still on my todo list to review...

zoq:
No worries.

include/ensmallen_bits/ada_bound/ada_bound_update.hpp
```diff
@@ -0,0 +1,207 @@
...
+
+      const ElemType fl = parent.finalLr * stepSize / initialStepSize;
+      const ElemType lower = fl * (1.0 - 1.0 / (parent.gamma *
+          parent.iteration + 1));
+      const ElemType upper = fl * (1.0 + 1.0 / (parent.gamma *
+          parent.iteration));
+
+       // Applies bounds on actual learning rate.
+      iterate -= arma::clamp((stepSize *
+          std::sqrt(biasCorrection2) / biasCorrection1) / (arma::sqrt(v) +
```
rcurtin: The paper is rather confusing here.  Their notation is `V_t = diag(v_t)` but I truly don't understand the need for the `diag()` as `v_t` is a vector.  Anyway if I just forget about the `diag()` or assume it is the identity, which I think is the right thing to do, this implementation seems correct.

include/ensmallen_bits/ada_bound/ams_bound_update.hpp
```diff
@@ -0,0 +1,214 @@
...
+                 const double gamma = 1e-3,
+                 const double epsilon = 1e-8,
+                 const double beta1 = 0.9,
+                 const double beta2 = 0.999) :
+    finalLr(finalLr),
+    gamma(gamma),
+    epsilon(epsilon),
+    beta1(beta1),
+    beta2(beta2),
+    iteration(0)
```
rcurtin: I think these should be doubly indented. :+1:

include/ensmallen_bits/ada_bound/ada_bound_update.hpp
zoq: `v_t` could be a matrix if the `gradient` is a matrix as well, so I will add the `diag`; it shouldn't matter.

rcurtin: I'm not sure that's right though.  In the paper they restrict the parameters to be a vector, and the intention of Adam is to scale the momentum of each individual parameter.  It seems like then every element should be used, not just the diagonal.

zoq: https://openreview.net/forum?id=Bkg3g2R9FX&noteId=S1eizWWuCQ

But you are right, it doesn't really make sense. Also I can't see they use it in there reference implementation.

rcurtin: Okay, I think it is a poor and confusing response that the authors gave, but I think it gave me enough of a hint to understand.  In their case `diag()` does not mean "take the diagonal vector of a matrix" like in Armadillo but instead it must be more like Armadillo's `diagmat()` where the meaning is "create a matrix with this vector on the diagonal".

Then, the concept of `sqrt()` is equivalent to element-wise square root: if `A` is diagonal, then `sqrt(A)_{ii} == sqrt(A_ii)`.

But then they must be implicitly intending that the expression `(\alpha / \sqrt(V_t))` in line 6 of the algorithm has an implicit flattening of the 2-d result: i.e., `\hat{eta}_t = Clip(vectorise(\alpha / \sqrt(V_t)), \eta_l(t), \eta_u(t))` (to abuse notation a little bit).

So, in the end, I think this means your original implementation was correct... let me know if you think otherwise.

HISTORY.md
```diff
@@ -3,6 +3,9 @@
   * Add additional L-BFGS convergence check
     ([#136](https://github.com/mlpack/ensmallen/pull/136)).
 
+  * Add AdaBound and AMSBound
+    ([#137](https://github.com/mlpack/ensmallen/pull/137)).
```
rcurtin: I think we'll need to rebase this on current master---easy enough to do during merge.

doc/optimizers.md
```diff
@@ -1,3 +1,59 @@
+## AdaBound
```
rcurtin: I just remembered, we should add `AdaBound` and `AMSBound` to the list of optimizers that can be used with separable differentiable functions in `doc/function_types.md`. :+1:

include/ensmallen_bits/ada_bound/ams_bound_update.hpp
```diff
@@ -0,0 +1,214 @@
...
+          parent.iteration);
+
+      const ElemType fl = parent.finalLr * stepSize / initialStepSize;
+      const ElemType lower = fl * (1.0 - 1.0 / (parent.gamma *
+          parent.iteration + 1));
+      const ElemType upper = fl * (1.0 + 1.0 / (parent.gamma *
+          parent.iteration));
+
+      // Element wise maximum of past and present squared gradients.
+      vImproved = arma::max(vImproved, v.diag());
```
rcurtin: I think everything is correct except this, so if you want to revert 91b7653 or point out what I missed, then I think this is ready for merge. :+1:

zoq:
Thanks for the review, I've merged the latest changes into the PR and updated `HISTORY.md`.

coatless:
@rcurtin are you planning on rolling another release after this goes in? (And the 1-step ahead PR?)

rcurtin:
@coatless yeah, I was.  I was hoping to review the k-step PR, and if that looks like it will be a quick merge (it probably will, just guessing) then I'll wait until then to release ensmallen 2.11.0.  I guess you are probably trying to decide whether to package 2.10.5? :)

include/ensmallen_bits/ada_bound/ada_bound.hpp
```diff
@@ -0,0 +1,208 @@
...
+  //! are reset before Optimize call.
+  bool& ResetPolicy() { return optimizer.ResetPolicy(); }
+
+ private:
+  //! The Stochastic Gradient Descent object with QHAdam policy.
+  SGD<UpdatePolicyType, DecayPolicyType> optimizer;
+};
+
+using AdaBound = AdaBoundType<AdaBoundUpdate>;
+using AMSBound = AdaBoundType<AdaBoundUpdate>;
```
rcurtin: Ahh, caught this as I was scrolling by :)
```suggestion
using AMSBound = AdaBoundType<AMSBoundUpdate>;
```

zoq: Nice catch, thanks!

coatless:
>  I guess you are probably trying to decide whether to package 2.10.5? :)

@rcurtin that am I .

rcurtin:
@coatless I think we can aim for a release in roughly a week.  Hope that helps the decision. :)

rcurtin:
Awesome, excited to have this new optimizer available. :+1:

PR #138, opened 2019-09-25
Lookahead Optimizer: k steps forward, 1 step back

zoq:
Implementation of "Lookahead Optimizer: k steps forward, 1 step back" by Michael R. Zhang, James Lucas, Geoffrey Hinton, Jimmy Ba.

### DIFF
--- a/doc/function_types.md
+++ b/doc/function_types.md
@@ -551,7 +551,7 @@ and f'(x) compute some of the same intermediate quantities.
 Each of the implemented methods is allowed to have additional cv-modifiers
 (`static`, `const`, etc.).
 
-The following optimizers can be used with differentiable functions:
+The following optimizers can be used with differentiable separable functions:
 
  - [AdaDelta](#adadelta)
  - [AdaGrad](#adagrad)
--- a/doc/function_types.md
+++ b/doc/function_types.md
@@ -561,6 +561,7 @@ The following optimizers can be used with differentiable functions:
  - [Big Batch SGD](#big-batch-sgd)
  - [IQN](#iqn)
  - [Katyusha](#katyusha)
+ - [Lookahead](#lookahead)
  - [Momentum SGD](#momentum-sgd)
  - [Nadam](#nadam)
  - [NadaMax](#nadamax)
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -986,6 +986,62 @@ optimizer.Optimize(f, coordinates);
  * [Limited-memory BFGS in Wikipedia](https://en.wikipedia.org/wiki/Limited-memory_BFGS)
  * [Differentiable functions](#differentiable-functions)
 
+## Lookahead
+
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+Lookahead is a stochastic gradient based optimization method which chooses a
+search direction by looking ahead at the sequence of "fast weights" generated
+by another optimizer.
+
+#### Constructors
+
+ * `Lookahead(`_`baseOptimizer`_`)`
+ * `Lookahead(`_`baseOptimizer, stepSize`_`)`
+ * `Lookahead(`_`baseOptimizer, stepSize, k`_`)`
+ * `Lookahead(`_`baseOptimizer, stepSize, k, maxIterations, tolerance, decayPolicy, exactObjective`_`)`
+
+Note that `Lookahead` is based on the templated type
+`Lookahead<`_`BaseOptimizerType, DecayPolicyType`_`>` with _`DecayPolicyType`_` = NoDecay`.
+
+Any optimizer that implements the differentiable separable functions interface
+can be paired with the `Lookahead` optimizer.
+
+#### Attributes
+
+| **type** | **name** | **description** | **default** |
+|----------|----------|-----------------|-------------|
+| `BaseOptimizerType` | **`baseOptimizer`** |  Optimizer for the forward step. | **n/a** |
+| `double` | **`stepSize`** | Step size for each iteration. | `0.001` |
+| `size_t` | **`k`** | The synchronization period. | `5` |
+| `size_t` | **`max_iterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+| `double` | **`tolerance`** | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+| `DecayPolicyType` | **`decayPolicy`** | Instantiated decay policy used to adjust the step size. | `DecayPolicyType()` |
+| `bool` | **`exactObjective`** | Calculate the exact objective (Default: estimate the final objective obtained on the last pass over the data). | `false` |
+
+The attributes of the optimizer may also be modified via the member methods
+`BaseOptimizer()`, `StepSize()`, `K()`, `MaxIterations()`,
+`Tolerance()`, `DecayPolicy()` and `ExactObjective()`.
+
+#### Examples
+
+```c++
+RosenbrockFunction f;
+arma::mat coordinates = f.GetInitialPoint();
+
+Adam adam(0.01, 2, 0.7, 0.999, 1e-8, 5, 1e-3, false);
+Lookahead<Adam> optimizer(adam, 0.5, 5, 100000, 1e-5);
+
+optimizer.Optimize(f, coordinates);
+```
+
+#### See also:
+
+ * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
+ * [SGD](#standard-sgd)
+ * [Lookahead Optimizer: k steps forward, 1 step back](https://arxiv.org/abs/1907.08610)
+ * [Differentiable separable functions](#differentiable-separable-functions)
+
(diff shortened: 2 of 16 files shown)


zoq:
Let us keep this open.

conradsnicta:
This optimiser seems related (at least in spirit) to the Late Acceptance Hill-Climbing (LAHC) optimiser, which has been around since about 2008 in one form or another.  A recent-ish journal article on LAHC is here: https://doi.org/10.1016/j.ejor.2016.07.012  
Recent improvements on LAHC are here: https://doi.org/10.1007/978-3-030-03991-2_29

Curiously, the arxiv pre-print for the Lookahead optimiser doesn't seem to mention LAHC at all.  Maybe this is a case of convergent evolution.

zoq:
Thanks for the reference, I agree it seems related; the improvements from the "Diversified Late Acceptance Search" paper are interesting as well, looks like there is an implementation for LAHC in python: https://github.com/Gunnstein/lahc

doc/optimizers.md
```diff
@@ -986,6 +986,62 @@ optimizer.Optimize(f, coordinates);
  * [Limited-memory BFGS in Wikipedia](https://en.wikipedia.org/wiki/Limited-memory_BFGS)
  * [Differentiable functions](#differentiable-functions)
 
+## Lookahead
+
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+Lookahead is a stochastic gradient based optimization method which chooses a
+search direction by looking ahead at the sequence of "fast weights" generated
+by another optimizer.
```
rcurtin: `Lookahead` seems different from our other optimizers in that it requires an inner optimizer type.  I think we could either point that out in the documentation here, meaning that the constructors documentation should be adapted to `Lookahead<BaseOptimizerType>(...)`, or, we could make `Adam` the default and also provide constructor overloads that don't need `baseOptimizer` so a user can construct it without needing to think about the base optimizer.  What do you think?

include/ensmallen_bits/callbacks/traits.hpp
```diff
@@ -371,15 +373,20 @@ struct HasStepTakenSignature
...
-  using BatchSizeForm = size_t(C::*)(void) const;
+  using BatchSizeConstForm = size_t(C::*)(void) const;
+
+  template<typename C>
+  using BatchSizeForm = size_t(C::*)(void);
 
   const static bool value =
-      HasBatchSize<OptimizerType, BatchSizeForm>::value;
+      HasBatchSize<OptimizerType, BatchSizeForm>::value ||
+      HasBatchSize<OptimizerType, BatchSizeConstForm>::value;
```
rcurtin: Ahh, nice fix :+1:

include/ensmallen_bits/eve/eve.hpp
```diff
@@ -88,9 +88,9 @@ class Eve
       const bool exactObjective = false);
 
   /**
-   * Optimize the given function using stochastic gradient descent.  The given
-   * starting point will be modified to store the finishing point of the
-   * algorithm, and the final objective value is returned.
+   * Optimize the given function using Eve. The given starting point will be
+   * modified to store the finishing point of the algorithm, and the final
+   * objective value is returned.
```
rcurtin: Nice catch :+1:

include/ensmallen_bits/lookahead/lookahead.hpp
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * @code
+ * @article{Zhang2019,
+ *   author  = {Michael R. Zhang and James Lucas and Geoffrey E. Hinton and
+ *              Jimmy Ba},
+ *   title   = {Lookahead Optimizer: k steps forward, 1 step back},
+ *   journal = {CoRR},
+ *   year    = {2019},
+ *   url     = {http://arxiv.org/abs/1907.08610}
+ * }
```
rcurtin: I think we need an `@endcode` here too.

include/ensmallen_bits/lookahead/lookahead_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+      BaseGradType>();
+  RequireFloatingPointType<BaseMatType>();
+  RequireFloatingPointType<BaseGradType>();
+  RequireSameInternalTypes<BaseMatType, BaseGradType>();
+
+  BaseMatType& iterate = (BaseMatType&) iterateIn;
+
+  // Check if the optimizer implements HasMaxIterations() and override the
+  // parameter with k.
+  if (callbacks::traits::HasMaxIterationsSignature<BaseOptimizerType>::value)
```
rcurtin: These traits generally don't have to do with the callbacks themselves; if you wanted, maybe we should move them out of the `callbacks` namespace and put them into `function/traits.hpp` or something?  I am not picky either way, and no need to do that in this PR, just wanted to see what you think of the idea.

include/ensmallen_bits/lookahead/lookahead_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+  {
+    baseOptimizer.MaxIterations() = k;
+  }
+  else
+  {
+    Warn << "The BaseOptimizerType does not have a definition of "
+        << "MaxIterations(), don't override the number of iteration.";
```
rcurtin: Maybe instead of `don't override the number of iteration`, it's more clear to say `the given value of k will be ignored` or something like this?  The user should probably be aware that in a situation like this, the inner optimizer will probably run until convergence and that will make this whole Lookahead strategy take way too long.  In fact, I wouldn't be opposed to forcing the existence of a `MaxIterations()` function and failing if it's not there.

include/ensmallen_bits/lookahead/lookahead_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  // Check if the optimizer implements ResetPolicy() and override the reset
+  // policy setting if true.
+  if (callbacks::traits::HasResetPolicySignature<BaseOptimizerType>::value &&
+      baseOptimizer.ResetPolicy())
+  {
+    Warn << "Parameters are reset before every Optimize call; set "
+        << "ResetPolicy() to false.";
+    baseOptimizer.ResetPolicy() = false;
```
rcurtin: Maybe it's worth adding a boolean parameter to `Lookahead` that will set whether or not `ResetPolicy()` is called at the start of every outer iteration?  It's not clear to me what they do in the paper implementation, though, which is probably what we should choose as the default if we offer an option like this.

include/ensmallen_bits/lookahead/lookahead_impl.hpp
```diff
@@ -0,0 +1,191 @@
...
+      << "terminating optimization." << std::endl;
+
+  // Calculate final objective if exactObjective is set to true.
+  if (exactObjective)
+  {
+    // Find the number of functions to use.
+    const size_t numFunctions = f.NumFunctions();
+
+    size_t batchSize = 1;
+    // Check if the optimizer implements the SatchSize() method and use the
```
rcurtin:
```suggestion
    // Check if the optimizer implements the BatchSize() method and use the
```
:)

include/ensmallen_bits/lookahead/lookahead_impl.hpp
zoq: Agreed, not sure why I put them into the callbacks traits.

include/ensmallen_bits/lookahead/lookahead_impl.hpp
zoq: There are some optimizers that don't implement ``MaxIterations()`` like Differential Evolution which could work with the Lookahead optimizer without running forever, so unless you have a strong opinion here, I would go with the rewording.

include/ensmallen_bits/lookahead/lookahead.hpp
zoq: Nice catch, looks like there are a couple more optimizer where the `@endcode` or `@code` is missing, so I will fix those as well.

include/ensmallen_bits/lookahead/lookahead_impl.hpp
zoq: Good idea.

include/ensmallen_bits/lookahead/lookahead_impl.hpp
rcurtin: Yeah, it's not a huge deal either way---if it's more convenient I think it's fine to open an issue or do it in a future PR.

doc/function_types.md
```diff
@@ -561,6 +561,7 @@ The following optimizers can be used with differentiable functions:
  - [Big Batch SGD](#big-batch-sgd)
  - [IQN](#iqn)
  - [Katyusha](#katyusha)
+ - [Lookahead](#lookahead)
```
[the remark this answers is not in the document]
zoq: Nice catch, done.

rcurtin:
Awesome, I'll release a new version of ensmallen with this and #137 now. :)

PR #141, opened 2019-11-09
Add optional test building

robertohueso:
According to `CMakeLists.txt`

> This project has no configurable options---it just installs the headers to the install location, and optionally builds the test program.

but it's not optional at the moment. This PR adds the option in CMake :smile:.


robertohueso:
History updated :) I forgot it :sweat_smile:.

Issue #144, opened 2019-12-08
DE: need for DecomposableFunction not stated in documentation


ElianeBriand:
#### Description of problem

The differential evolution algorithm is documented as needing an arbitrary function, both in the function type and optimizer section of the documentation.

However, using the code in the master branch (commit 9b730c7), the DE algorithm need a NonDifferentiableDecomposableFunctionType according to the static_assert error message.

In addition to the discrepancy between documentation and code, the static_assert error message speaks of decomposable function, but the documentation use the term separable function only. These terms are synonyms, but this may not be obvious to the user. If you ctrl+F in the documentation for decomposable, there is no result, which is unfortunate because it make fixing error in one's code more difficult. I suggest either saying the function need to be separable in the static_assert message, or mentioning the term decomposable function in the documentation.

#### Problem location
http://ensmallen.org/docs.html

doc/functions_types.md
doc/optimizers.md

rcurtin:
Hey @ElianeBriand, thanks for the clear bug report---it made my job easy.  It turns out DE actually should work with arbitrary functions, not arbitrary separable functions.  So I fixed the check in that code.

Also, you are right, we should definitely have documentation whose terms match the error messages.  That was an artifact from early ensmallen development when we hadn't settled on the terms to use yet. :(

I opened #145 which fixes both of these issues; now, error messages will use the terms `separable` and `arbitrary`.

Thanks for pointing out the issue! :+1:

ElianeBriand:
Thanks for the fix. And thanks for maintaining this library, it's really one of the best and easier to user optimization libs available in C++.

rcurtin:
Great to hear that, thanks! :+1:

rcurtin:
I released ensmallen 2.10.5 today with these documentation fixes.  Thanks again for the report!

PR #145, opened 2019-12-10
Update static check terms to match documentation

rcurtin:
This solves #144.  The original code for the static function type checks used the terms `decomposable` and `non-differentiable`, but in the documentation we use `separable` and `arbitrary` to represent those two concepts.  This PR changes all of the code to match the terms in the documentation.

### DIFF
--- a/include/ensmallen_bits/cmaes/cmaes.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes.hpp
@@ -82,7 +82,7 @@ class CMAES
    * modified to store the finishing point of the algorithm, and the final
    * objective value is returned.
    *
-   * @tparam DecomposableFunctionType Type of the function to be optimized.
+   * @tparam SeparableFunctionType Type of the function to be optimized.
    * @tparam MatType Type of matrix to optimize.
    * @tparam CallbackTypes Types of callback functions.
    * @param function Function to optimize.
@@ -90,10 +90,10 @@ class CMAES
    * @param callbacks Callback functions.
    * @return Objective value of the final point.
    */
-  template<typename DecomposableFunctionType,
+  template<typename SeparableFunctionType,
            typename MatType,
            typename... CallbackTypes>
-  typename MatType::elem_type Optimize(DecomposableFunctionType& function,
+  typename MatType::elem_type Optimize(SeparableFunctionType& function,
                                        MatType& iterate,
                                        CallbackTypes&&... callbacks);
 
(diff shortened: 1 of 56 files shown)


include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -82,18 +82,18 @@ class CMAES
...
    * @param iterate Starting point (will be modified).
    * @param callbacks Callback functions.
    * @return Objective value of the final point.
    */
-  template<typename DecomposableFunctionType,
+  template<typename SeparableFunctionType,
            typename MatType,
            typename... CallbackTypes>
-  typename MatType::elem_type Optimize(DecomposableFunctionType& function,
+  typename MatType::elem_type Optimize(SeparableFunctionType& function,
```
zoq: Maybe it makes sense to rewrite the calculation of the first objective:

https://github.com/mlpack/ensmallen/blob/9b730c721294db0df7c0b752994b9d255ef58479/include/ensmallen_bits/cmaes/cmaes_impl.hpp#L109-L119

and use the `ArbitraryFunctionTypeAPI` as well. Let me know what you think, I can open a PR.

rcurtin: It looks like `RandomSelection` depends on the objective function being separable: https://github.com/mlpack/ensmallen/blob/9b730c721294db0df7c0b752994b9d255ef58479/include/ensmallen_bits/cmaes/random_selection.hpp#L49

So maybe we shouldn't?  I guess another idea would be to make `Function` provide non-separable overloads of `Evaluate()` for separable functions (which I've been meaning to do for a while anyway).

zoq: You are right, another overload sounds like the right thing to do. I think we should open another PR for that, what do you think?

rcurtin: Sure, that works.  Do you want me to make the code changes and open the PR, or do you want to?

zoq: I can open a PR in the next days, unless you like to, probably not urgent.

rcurtin: Sounds good, I'll focus on other PRs like #138.

zoq: Sounds good.

PR #147, opened 2019-12-10
Remove 'using namespace std' from problems.

### LINKED ISSUE #146: using std namespace in header files
#### Issue description

There are several header files with `using namespace std;`. This is a bad practise because it pollutes the application namespace when including those header files in your application. 
Here is the list of the header files with such a definition:

ensmallen_bits/problems/ackley_function_impl.hpp
ensmallen_bits/problems/beale_function_impl.hpp
ensmallen_bits/problems/cross_in_tray_function_impl.hpp
ensmallen_bits/problems/goldstein_price_function_impl.hpp
ensmallen_bits/problems/himmelblau_function_impl.hpp
ensmallen_bits/problems/holder_table_function_impl.hpp
ensmallen_bits/problems/levy_function_n13_impl.hpp
ensmallen_bits/problems/schaffer_function_n2_impl.hpp
ensmallen_bits/problems/schaffer_function_n4_impl.hpp
ensmallen_bits/problems/three_hump_camel_function_impl.hpp

#### Your environment

 * version of ensmallen: 2.10.4
 * operating system: Linux Debian 10
 * compiler: gcc 8.3.0
 * version of Armadillo:
 * any other environment information you think is relevant:

#### Steps to reproduce

include ensmallen header file in the code

#### Expected behavior

Including ensmallen in your code must not make std namespace to be used by default.

#### Actual behavior

Including ensmallen in your code causes std namespace to be used by default.

rcurtin:
It made for a bit more `std::pow()`, `std::sin()`, etc., but anyway, now including anything in the `problems/` directory should not cause any problems for downstream code.  This fixes #146.

### DIFF
--- a/include/ensmallen_bits/problems/cross_in_tray_function_impl.hpp
+++ b/include/ensmallen_bits/problems/cross_in_tray_function_impl.hpp
@@ -14,7 +14,6 @@
 
 // In case it hasn't been included yet.
 #include "cross_in_tray_function.hpp"
-using namespace std;
 
 namespace ens {
 namespace test {
@@ -36,9 +35,9 @@ typename MatType::elem_type CrossInTrayFunction::Evaluate(
   const ElemType x1 = coordinates(0);
   const ElemType x2 = coordinates(1);
 
-  const ElemType objective = -0.0001 * pow(abs(sin(x1) * sin(x2) *
-      exp(abs(100 - (sqrt(pow(x1, 2) + pow(x2, 2)) /
-      arma::datum::pi))) + 1), 0.1);
+  const ElemType objective = -0.0001 * std::pow(std::abs(std::sin(x1) *
+      std::sin(x2) * std::exp(std::abs(100 - (std::sqrt(std::pow(x1, 2) +
+      std::pow(x2, 2)) / arma::datum::pi))) + 1), 0.1);
   return objective;
 }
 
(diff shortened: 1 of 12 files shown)


include/ensmallen_bits/problems/cross_in_tray_function_impl.hpp
```diff
@@ -36,9 +35,9 @@ typename MatType::elem_type CrossInTrayFunction::Evaluate(
   const ElemType x1 = coordinates(0);
   const ElemType x2 = coordinates(1);
 
-  const ElemType objective = -0.0001 * pow(abs(sin(x1) * sin(x2) *
-      exp(abs(100 - (sqrt(pow(x1, 2) + pow(x2, 2)) /
-      arma::datum::pi))) + 1), 0.1);
+  const ElemType objective = -0.0001 * std::pow(std::abs(std::sin(x1) *
```
zoq: Looks like you introduced another sinus.

zoq: Nevermind, thought there was a second `std::sin`.

rcurtin:
Still, maybe there is some issue, let me try to debug the Travis build.

rcurtin:
Ah, turns out that we needed a `using namespace std` in `pso_test.cpp`.  I'm glad I took the time to investigate it instead of assuming that it was a random test failure :)

PR #153, opened 2019-12-26
Lookahead - Synchronization period type fix

zoq:
The builds fails (gcc 9.2.1), if the return type is wrong.


zoq:
Once this is merged, we should make another release.

rcurtin:
I'll make another release once #154 is merged. (should be tomorrow morning I guess) :+1:

zoq:
Sounds good.

PR #158, opened 2020-01-09
AppVeyor config

rcurtin:
Here is an attempt at getting AppVeyor to build ensmallen and run the tests...


.appveyor.yml
```diff
@@ -0,0 +1,53 @@
+environment:
+  ARMADILLO_DOWNLOAD: "http://ftp.fau.de/macports/distfiles/armadillo/armadillo-8.400.0.tar.xz"
```
conradsnicta: Since this appveyor config is a new addition, might as well use the latest available version of Armadillo (9.800.3).
Armadillo 8.400 is unsupported and pretty ancient at this point; we're not forced to use that version for this config.
We're only forced to use 8.400 on Ubuntu, due to Ubuntu 18.04 LTS.

(Ubuntu 18.04 will soon be a distant memory anyway, as the new 20.04 LTS release is just around the corner. I suspect many Ubuntu users will quickly update to it.)

rcurtin: I see what you're saying, but actually I think it's important to build against 8.400.0 here, as it will help ensure that someone committing code isn't committing code that would break on our oldest allowed Armadillo version.  When newer Ubuntu (and RHEL) versions become standard we can bump the minimum Armadillo version, but for now I'm actually still hearing from people who are using Ubuntu 16.04...!

coatless:
Any reason why GitHub Actions are not being used?

https://github.com/features/actions

Can test across Linux/macOS/Windows and handle deployment

This avoids having to configure and wait for results from multiple CIs.

rcurtin:
The only reason is unfamiliarity. :)  If someone wants to come along later and update it, no issue from my end, but for now I know how to get AppVeyor setup pretty quickly. :+1:

say4n:
> The only reason is unfamiliarity. :) If someone wants to come along later and update it, no issue from my end, but for now I know how to get AppVeyor setup pretty quickly. 👍

@rcurtin If we are moving to GitHub actions, will it be for static analysis as well as for building for both the platforms (Ubuntu and Windows) right?

I can have a go at this.

rcurtin:
> > The only reason is unfamiliarity. :) If someone wants to come along later and update it, no issue from my end, but for now I know how to get AppVeyor setup pretty quickly. 👍
> 
> @rcurtin If we are moving to GitHub actions, will it be for static analysis as well as for building for both the platforms (Ubuntu and Windows) right?
> 
> I can have a go at this.

Personally I'm not planning on putting any time into Github actions.  In my view once we have the infrastructure working there isn't any problem or need to change it.

say4n:
> In my view once we have the infrastructure working there isn't any problem or need to change it.

Yeah, that is prudent. :)

zoq:
Not sure what your plans are, if you like to solve the failing test as part of the PR, or open another PR. Looking at the result it's close: input `[-5, -5]`, global minimum: `[-2.903, -2.903]`, result: `[-2.94137, -2.94137]`

Maybe increasing:

https://github.com/mlpack/ensmallen/blob/master/include/ensmallen_bits/swats/swats.hpp#L76 

helps, not sure, bascially just a shot in the dark.

rcurtin:
Thanks for looking at that... I'm not sure why it's failing but I strongly suspect that it's due to different handling of floating-point numbers by MSVC and gcc.  That probably makes the optimization terminate too early (maybe the convergence check succeeds with MSVC but not gcc?).  I tested locally (not on Windows) and wasn't able to reproduce the issue at all, even with many different random seeds.

So, given that the optimizer *does* appear to be working, I don't think there's actually a bug, so I just bumped the tolerance.

I think that the AppVeyor configuration will now work.  I can't believe it took so long... I thought it would be easy! :(

.appveyor.yml
```diff
@@ -0,0 +1,60 @@
...
+    -DLAPACK_LIBRARY=%BLAS_LIBRARY%
+    -DBLAS_LIBRARY=%BLAS_LIBRARY%
+    -DCMAKE_BUILD_TYPE=Release ..
+  - >
+    "%MSBUILD%" "ensmallen.sln"
+    /m /verbosity:minimal /nologo /p:BuildInParallel=true
+
+  # Run tests after copying libraries.
+  - ps: cp C:\projects\ensmallen\OpenBLAS.0.2.14.1\lib\native\bin\x64\*.* C:\projects\ensmallen\build\
+  - Release\ensmallen_tests.exe SWATSStyblinskiTangFunctionFMatTest
```
zoq: Looks like this isn't necessary anymore.

rcurtin: Thanks, fixed in c0ea1ed.

rcurtin:
Ok, I'll go ahead and merge once this passes the tests. :+1:

.appveyor.yml
```diff
@@ -0,0 +1,59 @@
+environment:
+  ARMADILLO_DOWNLOAD: "http://ftp.fau.de/macports/distfiles/armadillo/armadillo-8.400.0.tar.xz"
+  ARMADILLO_LIBRARY: "C:/projects/mlpack/armadillo-8.400.0/build/Debug/armadillo.lib"
```
birm: Is this used? I don't think there's even a C:/projects/mlpack/ folder for ensmallen. It seems to build just fine in any case.

rcurtin: I think you're right---nice catch. :+1: I believe that on Windows it will only link against LAPACK and BLAS directly, so `ARMADILLO_LIBRARY` will be ignored.  Let's see if it still builds if I remove it, and if so, I'll go ahead and merge. :+1:

rcurtin:
Awesome, looks like it all works.

PR #159, opened 2020-01-11
Fix Lookahead MaxIterations() check

zoq:
In the previous version each optimizer passed to Lookahead had to implement `MaxIterations()`.


zoq:
Same, just noticed it in the moment I used `Lookahead` with an optimizer  that didn't implement `MaxIterations()`.

PR #160, opened 2020-01-13
New High Dimensional Test Functions

SuryodayBasak:
This is an initial commit - I just changed the argument list in `ackley_function.hpp`.

I'm trying to create functions that will be useful for testing high-dimensional optimization functions, and in particular, PSO. Hence, the functions that I will include will have a parameter `n` to specify the dimensionality.

The functions that I am currently looking at adding to the project, or changing in the project are:

1. Ackley
2. Rosenbrock
3. Griewank Function

On a side note, the other two functions that I will be using in order to test PSO in higher dimensions are Rastrigin and Sphere, and these are already present in the project.

### DIFF
--- a/include/ensmallen_bits/problems/ackley_function.hpp
+++ b/include/ensmallen_bits/problems/ackley_function.hpp
@@ -44,10 +44,11 @@ class AckleyFunction
   /**
    * Initialize the AckleyFunction.
    *
+   * @param n Number of dimensions of the objective function to be optimized.
    * @param c Multiplicative constant with a default value of 2 * pi.
    * @param epsilon Coefficient to avoid division by zero (numerical stability).
    */
-  AckleyFunction(const double n = 2,
+  AckleyFunction(const size_t n = 2,
                  const double c = 2 * arma::datum::pi,
                  const double epsilon = 1e-8);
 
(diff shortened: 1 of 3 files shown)


SuryodayBasak:
With regards to the Rosenbrock function: there are multiple test cases that use this, but I want to edit it to test functions in high dimensions. Would it be preferable to have a separate high-dimensional Rosenbrock function, or edit the current implementation?

In my opinion, it would be better to edit the current implementation but I would have to revise all the test cases that use it, but I'm willing to do that.

zoq:
Adapting the existing function sounds reasonable, we could provide a constructor that provides backward compatibility.

include/ensmallen_bits/problems/ackley_function.hpp
```diff
@@ -47,7 +47,8 @@ class AckleyFunction
    * @param c Multiplicative constant with a default value of 2 * pi.
    * @param epsilon Coefficient to avoid division by zero (numerical stability).
    */
-  AckleyFunction(const double c = 2 * arma::datum::pi,
+  AckleyFunction(const double n = 2,
```
zoq: Can you add a comment for the number of dimensions. Also, I think this isn't part of the public API, so we don't break compatibility.

rcurtin: Agreed, the test functions aren't documented anywhere, so for now I wouldn't consider them public.  It might be nice to make them public at some point, but I'm not sure how useful people would find them, and I don't think we need to do that now.

zoq:
Let us keep this open.

PR #161, opened 2020-01-15
prevent spurious compiler warnings

conradsnicta:
prevent spurious compiler warnings as experienced when compiling mlpack-master with gcc 7.4 on ubuntu 18.04

this makes more important compiler warnings + errors more visible during the compilation process


zoq:
Nice, saw the same issue on other configurations as well (https://dev.azure.com/mlpack/mlpack/_build/results?buildId=277&view=results).

PR #149, opened 2019-12-22
Add NSGA2

say4n:
This PR implements the [Non-dominated Sorting Genetic Algorithm - II](https://www.iitk.ac.in/kangal/Deb_NSGA-II.pdf) by Kalyanmoy Deb.
~It is a Work in Progress.~ :)

TODO:
- [x] Implement NSGA-II
- [x] Add tests
- [x] Add documentation

PR #165, opened 2020-02-10
Early stopping at loss min on validation set

shrit:
This pull request implement the discussion we had in the issue: https://github.com/mlpack/mlpack/issues/2185
This is only a first prototype, I have tested it on my neural network it seems to be working.
I implemented it in ensmallen, I know that @rcurtin requested to be implemented on mlpack side, but during the implementation on mlpack, I have noticed that I have to modify ensmallen callback functions, So I implemented it in ensmallen.
Still missing callback tests and documentation, but I do not know how to test it without data.


include/ensmallen_bits/callbacks/early_stop_at_min_loss_on_validation.hpp
```diff
@@ -0,0 +1,105 @@
...
+ */
+class EarlyStopAtMinLossOnValidation
+{
+ public:
+  /**
+   * Set up the early stop at min loss class, which keeps track of the minimum
+   * loss and stops the optimization process if the loss stops decreasing.
+   * 
+   * @param predictors: data matrix used to predict the responses.
+   *
```
say4n: Please remove the extra newlines here.

include/ensmallen_bits/callbacks/early_stop_at_min_loss_on_validation.hpp
```diff
@@ -0,0 +1,105 @@
+/**
+ * @file early_stop_at_min_loss.hpp
```
say4n: Please change these to reflect the correct information.

include/ensmallen_bits/callbacks/early_stop_at_min_loss_on_validation.hpp
```diff
@@ -0,0 +1,105 @@
...
+#ifndef ENSMALLEN_CALLBACKS_EARLY_STOP_AT_MIN_LOSS_ON_VALIDATION_HPP
+#define ENSMALLEN_CALLBACKS_EARLY_STOP_AT_MIN_LOSS_ON_VALIDATION_HPP
+
+namespace ens {
+
+/**
+ * Early stopping to terminate the optimization process early if the loss stops
+ * decreasing.
+ */
+class EarlyStopAtMinLossOnValidation
```
zoq: I guess instead of writing another callback class, we could also add another constructor to the existing  `EarlyStopAtMinLoss` class, let me know what you think, might avoid some code duplication.

shrit: I agree I prefer one class with two constructors, the only reason for this prototype is our last discussion on the related issue. I will merge the two classes and push the code.

zoq: Right, I guess that changed a bit because it's now part of ensmallen.

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -28,12 +29,37 @@ class EarlyStopAtMinLoss
    * @param patienceIn The number of epochs to wait after the minimum loss has
    *    been reached or no improvement has been made (Default: 10).
    */
-  EarlyStopAtMinLoss(const size_t patienceIn = 10) :
+  EarlyStopAtMinLoss(const size_t patienceIn = 10, 
+                    std::ostream& output = arma::get_cout_stream()) :
```
zoq: Pendantic style issue, this should align with the parameter above.

shrit:
I only overloaded the `Evalute()` function in sgd tests to pass the test, I did not implement it yet. I have tested it on real data set, it seems to be working

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -28,12 +29,37 @@ class EarlyStopAtMinLoss
...
+  EarlyStopAtMinLoss(arma::mat& predictors,
+                     arma::mat& responses,
+                     const size_t patienceIn = 10,
+                     std::ostream& output = arma::get_cout_stream()) :
+      patience(patienceIn),
+      bestObjective(std::numeric_limits<double>::max()),
+      steps(0),
+      output(output)
+  {
+    this->predictors = std::move(predictors);
```
zoq: Not sure we should use move here, maybe a simple reference is okay?

include/ensmallen_bits/function/add_evaluate.hpp
```diff
@@ -60,6 +60,42 @@ class AddEvaluate<FunctionType,
...
+                  false,
+                  false>
+{
+ public:
+  // Reflect the existing Evaluate().
+  typename MatType::elem_type Evaluate(const MatType& coordinates)
+  {
+    return static_cast<FunctionType*>(
+        static_cast<Function<FunctionType,
+                             MatType, GradType>*>(this))->Evaluate(coordinates);
```
zoq: I think this is one of the reasons why @rcurtin proposed to implement this in mlpack. I'm wondering if we can put this into the callback itself, instead of extending the Function Type interface, which we like to keep to a minimum.

shrit: @zoq I agree. However, I tried to implement it first in mlpack, but in mlpack case, we will have to change the architecture of this callback.
I think it is the same, even if we are in mlpack we will to `add_evaluate` in ensmallen. If we had a close look, it seems to me that `EndEpoch()` requires `FunctionType& function` in our case,  which is `ens::Function`. Suppose `ens::Function</*mlpack neural network  model should goes here*/>::Evaluate(arma::mat, arma::mat)` interface does not exist. Can you give me more details about possible implementation in mlpack?

Otherwise, I can try to move the overloaded interface in the callbacks.

shrit: @rcurtin Do you have any thoughts on this, any starting guide from mlpack?

rcurtin: I think that maybe the best way to put this together would be to implement a callback that has structure a little bit like this:

```
class EarlyStoppingOnValidationSet // maybe can be named better?
{
 private:
  FFN<>& network;
  arma::mat& validationPredictors;
  arma::mat& validationResponses;
  double lastResponse; // initialize to DBL_MAX?

 public:
  template<typename OptimizerType, typename FunctionType, typename MatType>
  bool EndEpoch(OptimizerType& /* optimizer */,
                FunctionType& /* function */,
                const MatType& /* coordinates */,
                const size_t /* epoch */,
                const double /* objective */)
  {
    // The FFN we are holding internally should be the same one as what is being optimized.
    const double validationLoss = ffn->Evaluate(validationPredictors, validationResponses);
    if (validationLoss > lastLoss)
       return true; // or do we return false to terminate early?  I can't remember

    return false;
  }
};
```

I've omitted some functions there, but maybe that's a decent enough idea?  It would definitely have to be in the mlpack codebase since it depends on `FFN`.  Let me know what you think---and sorry it took so long to get to this. :+1:

shrit: Now, I understand the idea related to the `FFN` reference. I will implement it and see if it works.
Thanks. :+1:

shrit:
Keep open

shrit:
@zoq can you have a look on this please. Thanks

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -45,11 +71,17 @@ class EarlyStopAtMinLoss
...
-                FunctionType& /* function */,
+                FunctionType& function,
                 const MatType& /* coordinates */,
                 const size_t /* epoch */,
-                const double objective)
+                double objective)
   {
+    if ((!predictors.empty()) and (!responses.empty()))
+    {
+      objective = network.Evaluate(predictors, responses);
```
zoq, line 81: About testing, what we could do is to write a custom/minimal version of the existing`LogisticRegressionFunction` class and add another `Evaluate()` function that internally uses `ComputeAccuracy()` and we interpret that as loss. Does this make sense, I can elaborate on the idea.

shrit: @zoq if I understood well, we have to write a small version of this class: https://github.com/mlpack/ensmallen/blob/82817a6e3b220cd8236121ef2ac2cb8f5e5bd32e/include/ensmallen_bits/problems/logistic_regression_function.hpp#L26
with `ComputeAccuracy()`. But I think we need to have a reference to `AnnType` in order to get to evaluate. Right?
In the actual implementation, I call `Evaluate` from `FFN`. However, in the tests, I do not think we can test ensmallen and having a dependency on mlpack. I am not sure.
A better solution is to re-open this pull request on mlpack side instead of ensmallen, this will be easier to test, but will be more confusing to users.

zoq: Right, but it's not really necessary only for the FFN class, as long as the classifer/model implements `Evaluate(predictors, responses);` you can use the callback, correct me if I'm wrong. So what I had envision is either create a another problem based on `LogisticRegressionFunction` which implements `Evaluate(predictors, responses);` or add `Evaluate(predictors, responses);` to the exsisting one, but not sure that is possible since there already a function with the same arguments, that does something different. Actually, I don't mind to implement a fully dummy function to test the callback as well. Let me know what you think.

shrit: @zoq I agree, what every classifier/model used that implements `Evaluate(predictors, responses);` should be able to use this callback.
It is the implementation details of `Evaluate` that I can not find a solution for it. Since adding a new `Evaluate` in the `add_evaluate.hpp` interface is not recommended, we might have to implement a new one locally (fully dummy function is a great idea). Beacause it is not the `Evaluate` function that we are testing here since whatever model is going to use this callback it will have it's own `Evaluate` that has already been tested. The main objective is to verify that the EarlyStop functionality is tested and works perfectly.

zoq: This is what I had in mind:

```c++
class LogisticRegressionValidationFunction
{
 public:
  LogisticRegressionValidationFunction(LogisticRegression<>& lrIn, arma::mat& coordinatesIn) : lr(lrIn), coordinates(coordinatesIn)
  {
  }

  void Evaluate(arma::mat& predictors, arma::mat& responses)
  {
    lr.ComputeAccuracy(predictors, responses, coordinates);
  }

  LogisticRegression<> lr;
  arma::mat coordinates;
};

arma::mat data, testData, shuffledData;
arma::Row<size_t> responses, testResponses, shuffledResponses;

LogisticRegressionTestData(data, testData, shuffledData,

SMORMS3 smorms3;
LogisticRegression<> lr(shuffledData, shuffledResponses, 0.5);
arma::mat coordinates = lr.GetInitialPoint();
LogisticRegressionValidationFunction lrValidation(lr, coordinates);


EarlyStopAtMinLoss cb(lrValidation, testData, testResponses);

smorms3.Optimize(lr, coordinates, cb);
```

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -28,11 +30,35 @@ class EarlyStopAtMinLoss
...
+   * Set up the early stop at min class, which keeps track
+   * of the minimum loss and stops the optimization process if the loss stops
+   * decreasing on a specific validation set.
+   *
+   * @param predictors: data matrix used to predict the responses.
+   * @param responses: data matrix used to evaluate the predictions.
+   * @param patienceIn The number of epochs to wait after the minimum loss has
+   * been reached or no improvement has been made (Default: 10).
+   */
+  EarlyStopAtMinLoss<AnnType>(AnnType& network,
```
zoq, line 50: I guess since this can be used with other methods as well, we should rename the parameter, what about `function`?

rcurtin: Right, seems like maybe this would be a perfect use case for a `std::function`?  We could be really general about it.  The user can just write a lambda that, when called, evaluates the loss on the given parameters...

```
// assume we built this network somehow and we want to optimize it...
FFN<...> network;
// we have some test set and responses...
arma::mat testSet;
arma::mat testResponses;

EarlyStopAtMinLoss e(
    [network, testSet, testResponses](const arma::mat& /* parameters */) -> double
    {
      return network.Evaluate(testSet, testResponses);
    });

network.Train(trainSet, trainResponses, e, ...);
```

That's just a sketch, but that's sufficiently general that the user can do whatever they want with it, with the easiest example being using a validation/test set to compute a loss.  I guess, from my end, I want to make sure that the callback isn't specific to the machine learning setting (since ensmallen is more general, it's just generic optimization).

zoq: I like the idea.

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
shrit: Thanks, I understand, I will put everything together in new commits

tests/callbacks_test.cpp
```diff
@@ -152,6 +152,31 @@ void CallbacksFullFunctionTest(OptimizerType& optimizer,
...
+  EarlyStopAtMinLoss<LogisticRegressionValidationFunction,
+  arma::mat, arma::Row<size_t>> cb(
+    lrValidation, testData, testResponses);
+
+  optimizer.Optimize(lr, coordinates, cb);
+}
+
+TEST_CASE("EarlyStopAtMinLossCallbackTest", "[CallbacksTest]")
+{
+  SMORMS3 smorms3;
```
shrit, line 177: @zoq  I have added the function as you have proposed, but I am not sure that we are stopping the optimization as we were doing before

tests/callbacks_test.cpp
```diff
@@ -470,24 +495,25 @@ TEST_CASE("SACallbacksFullFunctionTest", "[CallbacksTest]")
...
+  // StandardSGD s(0.0003, 1, 2000000000, -10);
+  // s.ExactObjective() = true;
+
+  // // The optimization process should return in one second.
+  // const double result = s.Optimize(f, coordinates, EarlyStopAtMinLoss(100));
+
+  // REQUIRE(result == Approx(-1.0).epsilon(0.0005));
+  // REQUIRE(coordinates(0) == Approx(0.0).margin(1e-3));
+  // REQUIRE(coordinates(1) == Approx(0.0).margin(1e-7));
+  // REQUIRE(coordinates(2) == Approx(0.0).margin(1e-7));
```
shrit, line 515: I have commented this one, if it is no longer required I will remove it

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
rcurtin: @shrit what do you think, would it be better to take any `std::function` here instead of requiring something that has an `Evaluate()` method?

shrit: @rcurtin Maybe I can add another constructor that takes an `std::function`, at the same time the user might do anything in `std::function` and not only `Evaluate`
I am not sure I can replace everything by `std::function`, since the `network.Evaluate` need to be called at the end of each epoch. right? I am not sure

rcurtin: Yeah, adding another constructor could be a nice thing to do, so that the technique is more general.  I guess, my hope is not to link the mlpack and ensmallen codebases too closely here, so that people can use this callback in non-mlpack situations.  If the user passes a lambda too, that actually makes the implementation easier, since `EarlyStopAtMinLoss` doesn't have to then track the validation set and responses (it will be implicitly held in the given lambda/`std::function`).

rcurtin:
@shrit can I ask what the status of this is?  When I look at the diff, it seems like the code is specific to mlpack neural networks.  The constructor you added is nice and I think that is the best interface we can give, but a couple comments:

 - I think now we can remove the `AnnType&` from the template parameters.
 - The constructor should probably accept a `std::function<double(const MatType&)>` (I think I am writing that signature right); basically we should be able to call the `std::function` with the given coordinates (of type `const MatType&`) and receive back a `double`.
 - Inside of the class, I think all we need to do is hold that `std::function<>`.
 - By default, we can have behavior such that if the `std::function<>` was not specified, we simply use `function.Evaluate()` like the current code does.  (Maybe we need a `bool` member to specify whether or not we have a `std::function`.)

Let me know what you think; I hope the comments here are helpful.  It would be really great to incorporate this into a new ensmallen release. :)

shrit:
@rcurtin I have applied the changes, your comments were very helpful, Thanks.
I have also updated the tests. However, I think they need to be checked a little bit.

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -28,11 +33,33 @@ class EarlyStopAtMinLoss
...
+   *
+   * @param func, callback to return immediate loss calculated bt the networl
+   * @param patienceIn The number of epochs to wait after the minimum loss has
+   *    been reached or no improvement has been made (Default: 10).
+   */
+  EarlyStopAtMinLoss<MatType>(
+      std::function<double(const MatType&)> func,
+      const size_t patienceIn = 10,
+      std::ostream& output = arma::get_cout_stream())
+    : callback_used(true),
```
rcurtin, line 57:
```suggestion
      std::ostream& output = arma::get_cout_stream()) :
      callback_used(true), 
```
A tiny style fix. :)

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -28,11 +33,33 @@ class EarlyStopAtMinLoss
...
+   * @param patienceIn The number of epochs to wait after the minimum loss has
+   *    been reached or no improvement has been made (Default: 10).
+   */
+  EarlyStopAtMinLoss<MatType>(
+      std::function<double(const MatType&)> func,
+      const size_t patienceIn = 10,
+      std::ostream& output = arma::get_cout_stream())
+    : callback_used(true), 
+      patience(patienceIn), bestObjective(std::numeric_limits<double>::max()),
+      steps(0), output(output), localFunc(func)
```
rcurtin, line 59:
```suggestion
      patience(patienceIn),
      bestObjective(std::numeric_limits<double>::max()),
      steps(0),
      output(output),
      localFunc(func)
```
I think the code is clearer when these are on different lines. :+1:

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -43,13 +70,19 @@ class EarlyStopAtMinLoss
    * @param epoch The index of the current epoch.
    * @param objective Objective value of the current point.
    */
-  template<typename OptimizerType, typename FunctionType, typename MatType>
+  template<typename OptimizerType, typename FunctionType>
   bool EndEpoch(OptimizerType& /* optimizer */,
                 FunctionType& /* function */,
-                const MatType& /* coordinates */,
+                const MatType&  coordinates,
```
rcurtin, line 76:
```suggestion
                const MatType& coordinates,
```

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -68,6 +101,9 @@ class EarlyStopAtMinLoss
   }
 
  private:
+  //! False if the first constructor is called, true is the user passed a lambda. 
+  bool callback_used;
```
rcurtin, line 105:
```suggestion
  bool callbackUsed;
```
We should avoid underscores in the variable names (as per the mlpack style guide). :+1:

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -43,13 +70,19 @@ class EarlyStopAtMinLoss
...
+  template<typename OptimizerType, typename FunctionType>
   bool EndEpoch(OptimizerType& /* optimizer */,
                 FunctionType& /* function */,
-                const MatType& /* coordinates */,
+                const MatType&  coordinates,
                 const size_t /* epoch */,
-                const double objective)
+                double objective)
   {
+    if (callback_used)
```
rcurtin, line 80:
```suggestion
    if (callbackUsed)
```

include/ensmallen_bits/function/add_evaluate.hpp
```diff
@@ -60,7 +60,7 @@ class AddEvaluate<FunctionType,
     return static_cast<FunctionType*>(
         static_cast<Function<FunctionType,
                              MatType, GradType>*>(this))->Evaluate(coordinates);
-  }
+   }
```
rcurtin, line 63:
```suggestion
  }
```
I think that this was right before.

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -28,11 +33,33 @@ class EarlyStopAtMinLoss
    * @param patienceIn The number of epochs to wait after the minimum loss has
    *    been reached or no improvement has been made (Default: 10).
    */
-  EarlyStopAtMinLoss(const size_t patienceIn = 10) :
-      patience(patienceIn),
+  EarlyStopAtMinLoss<MatType>(
+      const size_t patienceIn = 10,
+      std::ostream& output = arma::get_cout_stream())
```
rcurtin, line 38: I'm not sure about the use of an `ostream` here.  This means that `EarlyStopAtMinLoss` behaves in two really different ways: (1) don't pass in a callback, and nothing is printed; or (2) pass in a callback, that is used to compute the objective, and its value is printed all the time.  I think that maybe we should remove `output` from here, and perhaps later consider some way that we can print the validation loss in, e.g., `ProgressBar` or something like that.

include/ensmallen_bits/problems/logistic_regression_function_validation_function.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+namespace ens {
+namespace test {
+
+using namespace ens::test;
+
+class LogisticRegressionValidationFunction
+{
+ public:
+  LogisticRegressionValidationFunction(LogisticRegression<>& lrIn,
```
rcurtin, line 26: I'm confused; this class doesn't exist in ensmallen.  I don't think that this will work.  Probably you need to hold a `LogisticRegressionFunction` instead, and then implement the accuracy computation by hand in `Evaluate()`.

tests/callbacks_test.cpp
rcurtin: I think that we need to keep this test---I don't see a reason to remove it.  Maybe I overlooked something?

tests/callbacks_test.cpp
```diff
@@ -152,6 +152,32 @@ void CallbacksFullFunctionTest(OptimizerType& optimizer,
...
+  });
+
+  optimizer.Optimize(lr, coordinates, cb);
+}
+
+TEST_CASE("EarlyStopAtMinLossCallbackTest", "[CallbacksTest]")
+{
+  SMORMS3 smorms3;
+  EarlyStopCallbacksFunctionTest(smorms3);
+}
```
rcurtin, line 180: It would probably be good if we could also add a test that makes sure that the given lambda will terminate the optimization.  I would suggest this... use, e.g., the Rosenbrock function with a custom lambda callback that terminates whenever any of the coordinates has an absolute value of less than 10.  Then once the optimization terminates, make sure that we are nowhere near the minimum, and that at least one of the coordinates has an absolute value of less than 10.  Let me know if I can clarify that idea---it's not the only possible idea, of course.

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
shrit: Exactly, this is what is happening, If we do not passe lambda nothing will be printed.
It is true that this is confusing. I can print it locally from inside the lambda, in this case, there is no need to have `output` in this class.

include/ensmallen_bits/problems/logistic_regression_function_validation_function.hpp
shrit: Actually this class is a wrapper for `LogisticRegressionFunction` in order to provide an `Evaluate` function. However, this class is no longer required since we are using lambda now.

tests/callbacks_test.cpp
shrit: if you have a code sample of the idea, it would be greate, I am truly not expert in tests :+1:.

rcurtin: Ok, here is a rough outline.  It is not perfect.  In fact it may not even compile.  But hopefully it will be enough to get the idea across.

```
// Maybe the name could be shorter...
TEST_CASE("EarlyStopAtMinLossWithCustomLambdaTest", "[CallbacksTest]")
{
  // Use the 50-dimensional Rosenbrock function.
  GeneralizedRosenbrockFunction f(50);
  // Start at some really large point.
  f.fill(100.0);

  EarlyStopAtMinLoss<arma::mat> cb(
      [&](const arma::mat& coordinates)
      {
        // Terminate if any coordinate has a value less than 10.
        return arma::any(arma::abs(coordinates) < 10.0);
      });

  SMORMS3 smorms3;
  smorm3.Optimize(f, coordinates, cb);

  // Make sure that we did not get to the optimum.
  for (size_t i = 0; i < coordinates.n_elem; ++i)
    REQUIRE(std::abs(coordinates[i]) >= 3.0);
}
```
Anyway, I think that will work. :+1:

tests/callbacks_test.cpp
```diff
@@ -180,6 +180,29 @@ TEST_CASE("EarlyStopAtMinLossLambdaCallbackTest", "[CallbacksTest]")
...
+  // Start at some really large point.
+  arma::mat coordinates = f.GetInitialPoint();
+  coordinates.fill(100.0);
+
+  EarlyStopAtMinLoss<arma::mat> cb(
+      [&](const arma::mat& coordinates)
+      {
+        // Terminate if any coordinate has a value less than 10.
+        return arma::abs(coordinates).min();
+      });
```
shrit, line 196: @rcurtin The comparison is done in the `EndEpoch` function inside the  `EarlyStopCallback`. We only need the least coordinates to examin if it is less than `10`

shrit:
@rcurtin if this is passing the test I think we can merge, and eventually do a release

tests/callbacks_test.cpp
rcurtin: Oh, right, actually that changes how we need this to work then.  Basically the callback will stop when this lambda starts going *up* in value.  So, the expression we need here is a little different---we need something that's large (and falling, potentially) when we are far away from 10*the unit ball, and once we get inside it needs to go back up.  We can do that like this:

```
  double minVal = arma::abs(coordinates).min();
  return (minVal < 10.0) ? DBL_MAX : minVal;
```

I think that that will work, want to give it a try?

rcurtin:
Lots of static code analysis issues---I think that this is because we've just set up that job, and since you modified those files it's giving all kinds of errors from the already existing code too. :)  Are you willing to go through them?  I can also just provide a diff that you could apply (or even push directly to this branch); just let me know. :+1:

tests/callbacks_test.cpp
shrit: Agreed

shrit:
@rcurtin I have resolved the static error in the `early_stop_at_min_loss.hpp` file. However, all remaining errors are in the `callback_test.cpp`.
I am not able to understand the error in these cases:
http://ci.mlpack.org/job/pull-requests-ensmallen-static-code-analysis/lastBuild/cppcheck/new/fileName.-1705841338/source.c674556d-c112-4bf9-bbcb-13a56d40d5f1/#144
If you can show me what is causing this error I will fix all of them, they are all the same I think

rcurtin:
Yeah, agreed, those issues don't make any sense to me.  Nothing seems to be wrong.  If you can update `HISTORY.md` then Travis will be happy. :)

zoq:
False warning ignored in: https://github.com/mlpack/jenkins-conf/commit/c3ab7b9ecc9fbd3bb711973e78047a6e04cd1203.

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -28,11 +33,34 @@ class EarlyStopAtMinLoss
    * @param patienceIn The number of epochs to wait after the minimum loss has
    *    been reached or no improvement has been made (Default: 10).
    */
-  EarlyStopAtMinLoss(const size_t patienceIn = 10) :
-      patience(patienceIn),
+  EarlyStopAtMinLoss<MatType>(
+      const size_t patienceIn = 10)
+    : callbackUsed(false),
```
zoq, line 38:
```suggestion
  EarlyStopAtMinLoss<MatType>(const size_t patienceIn = 10) :
      callbackUsed(false), 
```

This should fit in a single line.

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -28,11 +33,34 @@ class EarlyStopAtMinLoss
...
-      patience(patienceIn),
+  EarlyStopAtMinLoss<MatType>(
+      const size_t patienceIn = 10)
+    : callbackUsed(false), 
+      patience(patienceIn), 
       bestObjective(std::numeric_limits<double>::max()),
       steps(0)
-  { /* Nothing to do here */ }
+  { /* Nothing to do here */
+  }
```
zoq, line 43:
```suggestion
  { /* Nothing to do here */ }
```

ust to be consistent with the rest of the code style.

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -12,12 +13,16 @@
 #ifndef ENSMALLEN_CALLBACKS_EARLY_STOP_AT_MIN_LOSS_HPP
 #define ENSMALLEN_CALLBACKS_EARLY_STOP_AT_MIN_LOSS_HPP
 
+#include <functional>
+#include <cmath>
```
zoq, line 17:
```suggestion
```

Already part of `ensmallen.hpp`.

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -76,6 +112,9 @@ class EarlyStopAtMinLoss
 
   //! Locally-stored number of steps since the loss improved.
   size_t steps;
+
+  //! function to call at the end of the epoch
```
zoq, line 116:
```suggestion
  //! Function to call at the end of the epoch.
```

Start with a capital letter and use a `.` at the end.

tests/callbacks_test.cpp
```diff
@@ -8,6 +8,8 @@
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
 
+#include <limits>
```
zoq, line 11: Hm, `ensmallen.hpp` includes `climits`, what about we remove `limits` here and put that into `ensmallen.hpp` instead and remove the `climits` header.

shrit: Agreed :+1:

doc/callbacks.md
```diff
@@ -104,6 +108,27 @@ RosenbrockFunction f;
...
+      [&](const arma::mat& coordinates)
+      {
+        // Terminate if any coordinate has a value less than 10.
+        double minValue = arma::abs(coordinates).min();
+        return (minValue < 10.0) ? 
+          std::numeric_limits<double>::max() : minValue;
+      });
+
+  SMORMS3 smorms3;
+  smorms3.Optimize(f, coordinates, cb);
```
rcurtin, line 130: The primary use case of this lambda is actually for a validation set.  So perhaps a better example might be this:

```suggestion
// Generate random training data and labels.
arma::mat trainingData(5, 100, arma::fill::randu);
arma::Row<size_t> trainingLabels =
    arma::randi<arma::Row<size_t>>(100, arma::distr_param(0, 1));
// Generate a validation set.
arma::mat validationData(5, 100, arma::fill::randu);
arma::Row<size_t> validationLabels =
    arma::randi<arma::Row<size_t>>(100, arma::distr_param(0, 1));

// Create a LogisticRegressionFunction for both the training and validation data.
LogisticRegressionFunction lrfTrain(trainingData, trainingLabels);
LogisticRegressionFunction lrfValidation(validationData, validationLabels);

// Create a callback that will terminate when the validation loss starts to
// increase.
EarlyStopAtMinLoss<arma::mat> cb(
    [&](const arma::mat& coordinates)
    {
      // You could also, e.g., print the validation loss here to watch it converge.
      return lrfValidation.Evaluate(coordinates);
    });
    
arma::mat coordinates = lrfTrain.GetInitialPoint();
SMORMS3 smorms3;
smorms3.Optimize(lrfTrain, coordinates, cb);
```

Maybe that example is a little long, but I tested it and it does work.

include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
```diff
@@ -12,12 +13,15 @@
...
 
+#include <functional>
+
 namespace ens {
 
 /**
  * Early stopping to terminate the optimization process early if the loss stops
  * decreasing.
  */
+template<typename MatType = arma::mat>
```
rcurtin, line 24: Crap, actually, I just realized, to make this change would force a major version bump:

https://github.com/mlpack/ensmallen/blob/master/UPDATING.txt

I think we can fix this by doing the following:

 - Rename this class to `EarlyStopAtMinLossType`.
 - Create a typedef so that `EarlyStopAtMinLoss` is `EarlyStopAtMinLossType<arma::mat>`.
 - Update the documentation (I'll provide a suggestion).
 - Add a comment here suggesting that we should change `EarlyStopAtMinLossType` to `EarlyStopAtMinLoss` for ensmallen 3.10.0.

doc/callbacks.md
```diff
@@ -84,12 +84,16 @@ has been made.
...
+ * `EarlyStopAtMinLoss(`_`func`_`)`
+ * `EarlyStopAtMinLoss(`_`func`_`,`_`patience`_`)`
 
 #### Attributes
 
 | **type** | **name** | **description** | **default** |
 |----------|----------|-----------------|-------------|
 | `size_t` | **`patience`** | The number of epochs to wait after the minimum loss has been reached. | `10` |
+| `std::function<double(const MatType&)>` | **`func`** | A callback to return immediate loss evaluated by the function. | |
+
```
rcurtin, line 96:
```suggestion
| `std::function<double(const arma::mat&)>` | **`func`** | A callback to return immediate loss evaluated by the function. | |

Note that for the `func` argument above, if a
[different matrix type](#alternate-matrix-types) is desired, instead of using
the class `EarlyStopAtMinLoss`, the class `EarlyStopAtMinLossType<MatType>`
should be used.
```

(This goes with [this comment](https://github.com/mlpack/ensmallen/pull/165/files#r450571058).)

If you want to rephrase my comment go ahead---this is just a suggestion. :+1:

rcurtin:
Everything looks great @shrit.  I just left a couple final comments (I promise, no more! :)); basically I realized that there are some reverse compatibility issues that we do need to handle.  Hopefully my suggestions are helpful---if you want to try it differently go ahead, the ideas I posted are just ideas. :)

tests/callbacks_test.cpp
```diff
@@ -481,7 +533,7 @@ TEST_CASE("EarlyStopAtMinLossCallbackTest", "[CallbacksTest]")
   s.ExactObjective() = true;
 
   // The optimization process should return in one second.
-  const double result = s.Optimize(f, coordinates, EarlyStopAtMinLoss(100));
+  const double result = s.Optimize(f, coordinates, EarlyStopAtMinLoss<arma::mat>(100));
```
rcurtin, line 536:
```suggestion
  const double result = s.Optimize(f, coordinates, EarlyStopAtMinLoss(100));
```
(This is for reverse compatibility; once `EarlyStopAtMinLoss` is a type without template parameters, this change will be needed.)

doc/callbacks.md
shrit: Of course, I agree we need to provide use case, you do not have to worry the above example looks great, I will just update the example repository once this gets merged

shrit:
@rcurtin Perfect, you are welcome, I am very excited about this pull request.

rcurtin:
Thanks @shrit!  I'll try to do a release with this support tomorrow. :+1:

zoq:
Maybe it makes sense to wait for https://github.com/mlpack/ensmallen/pull/149 before we do a new release, from my side this is pretty close to getting merged.

rcurtin:
The only thing is, the more releases, the closer I can get us to fully automatic releases.  So maybe I can do one today (once I finish debugging the mlpack-bot bit I'm working on now), and then one again a couple days later?

zoq:
I didn't thought about that, makes sense to me.

PR #167, opened 2020-02-17
Fix memory leak

rcurtin:
While using `valgrind` I noticed that the `instDecayPolicy` and `instUpdatePolicy` used by SGD actually aren't properly deallocated.  So this adds destructors to any optimizer that uses an `Any` internally, and that destructor calls `Clear()` to correctly free any allocated memory.

There are a few PRs now that have been merged since the last release, so I think we can do another release after this is merged.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,8 +2,9 @@
 ###### ????-??-??
   * Prevent spurious compiler warnings
     ([#161](https://github.com/mlpack/ensmallen/pull/161)).
-  
-  * ...
+
+  * Fix minor memory leaks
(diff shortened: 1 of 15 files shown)


birm:
I wonder what it would take to add the memory checker from mlpack to ensmallen too.

zoq:
> I wonder what it would take to add the [jenkins memory checker](http://ci.mlpack.org/job/pull-requests%20mlpack%20memory/) from mlpack to ensmallen too.

Should be simple, maybe we could even run against the complete test suite instead of single test, as we do for mlpack.

PR #169, opened 2020-02-19
proposed note on cmake requirements

conradsnicta:


README.md
```diff
@@ -16,6 +16,15 @@ gradient-free optimizers, and constrained optimization.
 * OpenBLAS or Intel MKL or LAPACK (see Armadillo site for details)
 
 
+### Installation
+
+ensmallen can be installed with CMake 3.3 or later.
+If CMake is not already available on your system, it can be obtained from https://cmake.org
```
rcurtin:
```suggestion
If CMake is not already available on your system, it can be obtained from https://cmake.org.
```
Sorry for the pedantry. :)

README.md
```diff
@@ -16,6 +16,15 @@ gradient-free optimizers, and constrained optimization.
...
+
+If you are using an older system such as RHEL 7 or CentOS 7,
+an updated version of CMake is also available via the EPEL repository.
```
rcurtin:
```suggestion
an updated version of CMake is also available via the EPEL repository via the `cmake3` package.
```
Thoughts?

README.md
```diff
@@ -16,6 +16,15 @@ gradient-free optimizers, and constrained optimization.
 * OpenBLAS or Intel MKL or LAPACK (see Armadillo site for details)
 
 
+### Installation
```
rcurtin: It seems to me like we can leave this as part of the `Requirements` section, so maybe this line isn't needed?

README.md
conradsnicta: Yeah, I left out the full stop at the end of the sentence, so the copy and paste of the address is easy.  This mainly applies to people reading the README with a text editor, or a viewer like "less".

I'm all for strictness where it's required, though allow for flexibility in corner cases like this.

README.md
conradsnicta: Looks good.

README.md
conradsnicta: Is CMake strictly required though?  If I'm not mistaken, ensmallen can be used without installation.

A user could just simply copy the files from `ensmallen/include` to wherever they like, or even just use the `ensmallen/include` directory directly during compilation.  Example:
```
  g++ prog.cpp -o prog -O2 -I ~/git/ensmallen/include -larmadillo
```

rcurtin: Oh, huh, good point.  It is required to build the tests, but you're right, they could do exactly what you specified.  I retract my comment then. :)

README.md
rcurtin: Fair point.  That missing period will bug me though. :)

README.md
conradsnicta: ha :) does this imply that the american date format (MM/DD/YY) bugs you as well?

README.md
conradsnicta: ha :) does this imply that the american date format (MM/DD/YY) bugs you as well?

favre49:
It seems I forgot to put the change in HISTORY.md as well. I can create another PR for this, or we can include it in this. Either works.

rcurtin:
> It seems I forgot to put the change in HISTORY.md as well. I can create another PR for this, or we can include it in this. Either works.

I meant to do this during merge but I forgot.  I'll open a PR in a second since I already have the modifications handy.

README.md
rcurtin: It does!  It doesn't sort right, and it also doesn't make sense.  I generally go with YYYY/MM/DD as a compromise.  At least, I think it's a compromise?  I have no idea.

Anyway I did figure out how to get my brain to rest easy on this one; let me know what you think:

```
If CMake is not already available on your system, http://cmake.org has instructions for how to download and install it.
```

I can make that change during merge.

conradsnicta: looks good.

PR #170, opened 2020-02-19
Update HISTORY.

rcurtin:
Simple history update that I forgot to commit during the #152 merge.

PR #171, opened 2020-02-19
Require history entry

coatless:
From #170, we add a small git diff script during the `before_install` to verify if a new `HISTORY.md` item was added compared to master.


coatless:
With any luck... The PR should error immediately :) 

(I didn't add a `HISTORY.md` commit.)

coatless:
Travis threw an error in ~1 min 14 seconds: 

Screen Shot 2020-02-19 at 4 12 57 PM

Next, adding a commit.

rcurtin:
Nice, thank you!  If only I had waited another five minutes before releasing 2.11.3. :)  I guess this one doesn't make too much of a difference for release anyway though.

coatless:
So, the PR passed nicely on Travis. The delay in testing was due to too many simultaneous jobs on the organization. (GitHub Actions would help with this!) 

Screen Shot 2020-02-19 at 5 31 01 PM

scripts/history-update-check.sh
```diff
@@ -0,0 +1,21 @@
...
+# This should be run from the root of the repository.
+set -e
+
+! git diff --exit-code master -- HISTORY.md > /dev/null
+
+if [ $? ]; then
+    echo "History log was updated ..."
+else
+    echo "History log is missing an entry for this PR ..."
+    echo "Please update the history log to run CI ..."
```
birm: I'd change to a slightly more descriptive return message such as "Please describe your changes in HISTORY.md" Otherwise, this is great! I love how simple it is.

coatless: @birm can you make the phrase a suggestion using:

````
```suggestion
    echo "Please update the history log to run CI ..."
```
````

Comment (birm) [reply scripts/history-update-check.sh]: Oh, that's how you do that! In any case, it looks like you did it.

**Comment by favre49** [comment]
Is this always required though? Sometimes there are PRs related to

**Comment by favre49** [comment]
Sorry, pressed the wrong buttons :/

**Comment by coatless** [comment]
@favre49 this is only triggered on PRs. Direct commits into master are not checked.

File: .travis.yml [hunk 2]
```diff
@@ -7,6 +7,9 @@ env:
   - ARMADILLO=minimum
 
 before_install:
+  - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
```
zoq: Personally I would put this to the end, after the build and test step; let's say I have a minor spelling fix, that would probably not end up in the history, but I still like to check that I haven't introduced anything that breaks the build.

coatless: @zoq the reason I put it first instead of last is two-fold:

1. Less likely to see it at the "end" of a raw log from travis 
   - ~3k lines each section is collapsed though.
2. The total time spent on CI is significant.
   - Likely to correct immediate if flagged earlier vs. later in the CI process. 
   - In particular, CI fails in ~2 minutes vs. ~14 min 49 sec.

Happy to change check position depending on what the consensus is.

zoq: I see what you mean, do you think we should create another job on jenkins `sanity check` or something, that way get a result immediately and we don't have to scroll through a bunch of console output. If someone commits a history commit right after he notice that the jenkins job fails the travis job should be automatically canceled.

coatless: @zoq I'm not familiar with Jenkins configuration. So, I'm deferring to your judgement and, potentially, a helping hand in configuring Jenkins. 

Though, to be honest, I'm frequently checking Travis over Jenkins...

birm: You could accomplish this by adding another line to the env section, and checking if _that_ is set. That way, we'd get both versions of armadillo and the sanity check.
I can't seem to suggest on multiple lines, but something like:
```env:
  - ARMADILLO=latest
  - ARMADILLO=minimum
 - SANITY=history

before_install:
  - if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [ "$SANITY" != "history" ]; then
```

rcurtin:
(I'm going to hijack this PR to also test another semi-related CI check---the valgrind checks for Jenkins.  So please excuse the opening and closing... :))

.travis.yml
zoq: That sounds like a great idea to me.

coatless: @zoq done. See: 09291ce

@birm thanks!

birm: Ah, it looks like you'd want to exit 0 on an ok check too, since it's trying to test everything right now.

zoq:
`"Enable sanity check with jenkins for history entry (h/t @birm)"` - Just to be clear we are still using travis but it's now part of an independent build.

.travis.yml
```diff
@@ -5,8 +5,12 @@ language: cpp
 env:
   - ARMADILLO=latest
   - ARMADILLO=minimum
+  - SANITY=history
```
zoq: We have to use `SANITY=history ARMADILLO=latest` here, to tell travis to test against the latest version.

birm: In that case, we'd be doing a lot of duplicate testing. The sanity check, in my opinion, should perform that test.

coatless: I agree about the duplicity of tests. 

As this is now an environment variable, we could potentially _stop_ the CI from running after the history check ran with `exit 0`.  Thus, we wouldn't need to set `ARMADILLO`.

Thoughts?

zoq: Fine with me, no need to run the same configuration twice.

.travis.yml
```diff
@@ -5,8 +5,12 @@ language: cpp
...
 
 before_install:
+  - if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [ "$SANITY" == "history" ]; then
+      sh ./scripts/history-update-check.sh;
+    fi
```
coatless:
```suggestion
      exit 0
    fi
```

.travis.yml
coatless: @zoq addressed by e28c0c0e4d8f08052d79f7de4bfb53fe7fce07f5

zoq: Nice, looks like the other runs don't know the `SANITY` env and fail because of that.

coatless: @zoq resolved with 882ecf3122d4d0e926f265fabf78733f1a2da4ba and 321f7201d6375f76605d910798841d1a25083b73. 😄

rcurtin:
Thanks @coatless! :rocket:

zoq:
@coatless looks like this fails if it's not a PR, I can fix that one, just let me know.

coatless:
@zoq Argh, needed a double `if`. Sorry about that. 

I think this would work.

```bash
if [ ! -z ${SANITY_HISTORY+x} ] && [ "$SANITY_HISTORY" == "true" ]; then
      if  [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
        sh ./scripts/history-update-check.sh;
      fi
      exit 0;
    fi
```

PR #172, opened 2020-02-28
not-so-quick fix sanity check for outside of PRs

birm:
This follows #171 and should hopefully fix the oversight for non pr-based tests

In doing so, I've made some other travis changes, and I'm not sure if they're desired.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,9 @@ jobs:
       script:
         - sudo apt-get update
         - sudo apt-get install -y --allow-unauthenticated libopenblas-dev liblapack-dev g++ xz-utils
-        - curl $ARMA_URL `curl $ARMA_URL -- | grep '.tar.xz' | sed 's/^.*<a href="\(armadillo-[0-9]*.[0-9]*.[0-9]*.tar.xz\)".*$/\1/' | tail -1` | tar xvJ && cd armadillo*;
-        - mkdir build && cd build && cmake -DCMAKE_CXX_FLAGS="-Werror" -DCMAKE_C_FLAGS="-Werror" .. && make -j2
+        - curl $ARMA_URL -- | grep '.tar.xz' | sed 's/^.*<a href="\(armadillo-[0-9]*.[0-9]*.[0-9]*.tar.xz\)".*$/\1/' | tail -1` | tar xvJ && cd armadillo*;
         - cmake . && make && sudo make install && cd ..
+        - mkdir build && cd build && cmake -DCMAKE_CXX_FLAGS="-Werror" -DCMAKE_C_FLAGS="-Werror" .. && make -j2
     - stage: ctest
       script:
         - CTEST_OUTPUT_ON_FAILURE=1 travis_wait 30 ctest -j2
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,12 @@ env:
 
 stages:
   - name: sanity
-    if: env(TRAVIS_PULL_REQUEST) != true AND env(false) = true
+    if: env(TRAVIS_PULL_REQUEST) != false AND env(SANITY_HISTORY) = true
   - name: cmake
     if: env(SANITY_HISTORY) != true
   - name: build
     if: env(SANITY_HISTORY) != true
-  - name: test
+  - name: ctest
     if: env(SANITY_HISTORY) != true
 
 before_install:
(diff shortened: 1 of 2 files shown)


birm:
If someone other than me merges this, please squash!

Stages didn't work how I expected them to.

Anyway, after ruining my fork's master branch (for now), it looks like the sanity check is indeed not checked on master.
https://travis-ci.org/birm/ensmallen/builds/656138653?utm_medium=notification&utm_source=github_status

I'm now trying to see if I can get it to sanity check _after_ the test, but I may/should sleep before then

HISTORY.md
```diff
@@ -3,7 +3,10 @@
 
   * Require new HISTORY.md entry for each PR.
     ([#171](https://github.com/mlpack/ensmallen/pull/171)).
-    
+
+  * Bugfix for HISTORY.md check
+    ([#172](https://github.com/mlpack/ensmallen/pull/171)).
```
zoq: Maybe just add this one to the existing entry: `([#171](https://github.com/mlpack/ensmallen/pull/171),[#172](https://github.com/mlpack/ensmallen/pull/171))` to keep `HISTORY.md` clean.

birm: Good point; thanks!

.travis.yml
```diff
@@ -1,33 +1,36 @@
...
   - cmake . && make && sudo make install && cd ..
-
-install:
   - mkdir build && cd build && cmake -DCMAKE_CXX_FLAGS="-Werror" -DCMAKE_C_FLAGS="-Werror" .. && make -j2
-
-script:
   - CTEST_OUTPUT_ON_FAILURE=1 travis_wait 30 ctest -j2
 
+
+
```
rcurtin: Any particular need to add these empty lines? :)

HISTORY.md
```diff
@@ -2,8 +2,8 @@
 ###### 2020-??-??
 
   * Require new HISTORY.md entry for each PR.
-    ([#171](https://github.com/mlpack/ensmallen/pull/171)).
-    
+    ([#171](https://github.com/mlpack/ensmallen/pull/171), [#172](https://github.com/mlpack/ensmallen/pull/172)).
```
rcurtin: Ah, this is over 80 characters; can you wrap the line? :+1:

coatless:
@birm any reason for the step approach? This seems to infringe on the results of the latest build. 

I thought the goal was to preserve existing tests + add the history check. Changing the `if` statements should be enough. c.f. 

https://github.com/mlpack/ensmallen/pull/171#issuecomment-592116801

rcurtin:
> @birm any reason for the step approach? This seems to infringe on the results of the latest build.
> 
> I thought the goal was to preserve existing tests + add the history check. Changing the `if` statements should be enough. c.f.
> 
> [#171 (comment)](https://github.com/mlpack/ensmallen/pull/171#issuecomment-592116801)

Ah, yeah, maybe that would be easier to go with (I overlooked that comment, sorry).  Personally there's no preference from my side, I think it's great to see the support either way. :)

birm:
The reasons I decided to use stages were mostly about filtering out the sanity task entirely, and making it clearer that it was a separate kind of thing in contrast to the build & ctest job.
My ulterior motive is that I think this may make it easier/clearer to have travis do other such parallel tasks such as this in the future.
That said, I'm not intensely committed to this concept.

zoq:
I think, if we are going to add more tasks in the future it makes sense, I guess we could put: https://github.com/mlpack/ensmallen/pull/116 in that other stage as well? But @coatless is right, the fix he proposed would fix the issue as well.

coatless:
@birm so, you basically get with filtering:

Screen Shot 2020-02-28 at 4 17 53 PM

Though, I'm confused.... Under this setup, does the `ARMADILLO=latest` build fail if a `HISTORY.md` entry isn't present?

coatless:
For the record, @birm solution is much better for long-term if using Travis. Mine is a very narrow short-term fix.

birm:
@coatless No, it won't fail the job, one stage only does the sanity check, so it it just made it easier to filter by having just two jobs. (I had some unexplained problems with three jobs and stages, but I don't think it's about the number). should I try to change it for readability?

zoq:
Personally, I like the split.

zoq:
@birm @coatless Looks like we missed something: https://github.com/mlpack/ensmallen/pull/174

PR #174, opened 2020-03-01
Example doc fix + add/use default argument.

zoq:
There are some examples in the documentation that can't be copy/pasted, which is either because we didn't use the correct syntax or because some of the parameters are missing.


zoq:
The history wasn't modified, intentionally, to see if it works.

zoq:
Yes it should, @birm is working on it here: https://github.com/mlpack/ensmallen/pull/175

Issue #150, opened 2019-12-24
Better output when convergence check succeeds


rcurtin:
#### Issue description

When using callbacks like `ens::ProgessBar` or similar, and the optimization terminates partway through an epoch due to a convergence check, the output can be confusing:

```
Epoch 1/29[=========================================================================================>..........] 89% - ETA: 0s - loss: 1115.75
```

but then output stops.

See the discussion at https://github.com/mlpack/mlpack/issues/2073 for many more details and how to reproduce.

#### Expected behavior

We should see if we can modify the `ens::ProgressBar` callback (and perhaps others) to give better output in these situations.  Perhaps something like this would be an improvement:

```
Epoch 1/29[=========================================================================================>..........] 89% - ETA: 0s - loss: 1115.75
Optimization converged! (value 1.33e-6, tolerance 1e-5)
```

#### Actual behavior

Instead the output just terminates:

```
Epoch 1/29[=========================================================================================>..........] 89% - ETA: 0s - loss: 1115.75
```

This would be a good issue if you are not familiar with how ensmallen's callbacks work and would like to be.  I don't have a direct route to a solution here, so my suggestion would be to investigate the current callback code, reproduce the issue, then think about the cleanest way to print convergence information in the `ens::ProgressBar` callback.  Once that's done, see if similar changes might be useful for some of the other callbacks that print information (see http://ensmallen.org/docs.html#callback-documentation for more information on the callbacks that are available).

vikram2000b:
I would like to take up this issue.

rcurtin:
Hi @vikram2000b, you're more than welcome to.  When you have a PR ready, someone will review it when they're able to. :+1:

dB2510:
Can take up this issue as there is no PR yet? 
I am a beginner and I want to explore how this works.

dB2510:
Can I take up this issue as there is no PR yet?
I am a beginner and I want to explore how this works.

zoq:
@dB2510 I think that is reasonable, @vikram2000b are you fine with that?

dB2510:
@zoq Any idea how can I reproduce the issue mentioned here https://github.com/mlpack/mlpack/issues/2073 ?

rcurtin:
@dB2510 there is some code in that issue that you can directly compile and use in a program to reproduce the issue.  You will probably have to find some datasets to use to reproduce the issue, but that shouldn't be hard---you can probably use some of the datasets in `src/mlpack/tests/data/` or something like that.

[...]

PR #180, opened 2020-03-15
Better output if the optimization process gets terminated during the execution of an epoch.

gaurav-singh1998:
- Issue reference #150 
Earlier if the entire dataset was not passed to the optimizer the output used to be like
```
Epoch 1/54
1563/1875 [===================================================================================>................] 83% - ETA: 0s - loss: 249.826
```

After the feature which has been implemented in this pull request the output is like
```
Epoch 1/54
1562/1875 [===================================================================================>................] 83% - ETA: 0s - loss: 249.985
1563/1875 [===================================================================================>................] 83% - ETA: 0s - loss: 249.826
Optimization terminated because of the entire dataset not being passed to the optimizer.
```

As it can be seen that there is an improvement in the output shown but if there is any way if the progress bar is shown only once rather than being shown twice as it can be seen from the above-pasted output kindly suggest it. Thanks.

### DIFF
--- a/include/ensmallen_bits/callbacks/progress_bar.hpp
+++ b/include/ensmallen_bits/callbacks/progress_bar.hpp
@@ -170,12 +170,12 @@ class ProgressBar
     {
       if (progress == (size_t)((double)(optimizer.MaxIterations()) /
                                 (double)(function.NumFunctions()) * 100) &&
-                      (int)(step * optimizer.BatchSize() -
+                      (int)(steps * optimizer.BatchSize() -
                       optimizer.MaxIterations()) >= 0)
       {
         output << "\n" << "Optimization terminated because of the entire "
                        << "dataset not being passed to the optimizer."
-                       << "\r";
+                       << "\n" << "\r";
       }
     }
 
--- a/include/ensmallen_bits/callbacks/progress_bar.hpp
+++ b/include/ensmallen_bits/callbacks/progress_bar.hpp
@@ -183,7 +183,7 @@ class ProgressBar
           }
           else if (i == progress)
           {
-            output << ">";
+            output << "|";
           }
           else
           {
--- a/include/ensmallen_bits/callbacks/progress_bar.hpp
+++ b/include/ensmallen_bits/callbacks/progress_bar.hpp
@@ -132,8 +132,8 @@ class ProgressBar
    * @param objective Objective value of the current point.
    */
   template<typename OptimizerType, typename FunctionType, typename MatType>
-  void StepTaken(OptimizerType& /* optimizer */,
-                 FunctionType& /* function */,
+  void StepTaken(OptimizerType& optimizer,
+                 FunctionType& function,
                  const MatType& /* coordinates */)
   {
     if (newEpoch)
@@ -164,8 +164,22 @@ class ProgressBar
     output << "] " << progress << "% - ETA: " << (size_t) stepTimer.toc() *
         (epochSize - step + 1) % 60 << "s - loss: " <<
         objective / (double) step <<  "\r";
-    output.flush();
 
+    if (optimizer.MaxIterations() < function.NumFunctions() &&
+        optimizer.MaxIterations() != 0)
+    {
+      if (progress == (size_t)((double)(optimizer.MaxIterations()) /
+                                (double)(function.NumFunctions()) * 100) &&
+                      (double)(step*optimizer.BatchSize() -
+                      optimizer.MaxIterations()) >= 0.0)
+      {
+        output << "\n" << "Optimization terminated because of the entire "
+                       << "dataset not being passed to the optimizer."
+                       << "\r";
+      }
+    }
+
+    output.flush();
     stepTimer.tic();
   }
 
--- a/include/ensmallen_bits/callbacks/progress_bar.hpp
+++ b/include/ensmallen_bits/callbacks/progress_bar.hpp
@@ -170,7 +170,7 @@ class ProgressBar
     {
       if (progress == (size_t)((double)(optimizer.MaxIterations()) /
                                 (double)(function.NumFunctions()) * 100) &&
-                      (int)(step*optimizer.BatchSize() -
+                      (int)(step * optimizer.BatchSize() -
                       optimizer.MaxIterations()) >= 0)
       {
         output << "\n" << "Optimization terminated because of the entire "
--- a/include/ensmallen_bits/callbacks/progress_bar.hpp
+++ b/include/ensmallen_bits/callbacks/progress_bar.hpp
@@ -170,8 +170,8 @@ class ProgressBar
     {
       if (progress == (size_t)((double)(optimizer.MaxIterations()) /
                                 (double)(function.NumFunctions()) * 100) &&
-                      (double)(step*optimizer.BatchSize() -
-                      optimizer.MaxIterations()) >= 0.0)
+                      (int)(step*optimizer.BatchSize() -
+                      optimizer.MaxIterations()) >= 0)
       {
         output << "\n" << "Optimization terminated because of the entire "
                        << "dataset not being passed to the optimizer."
--- a/include/ensmallen_bits/callbacks/progress_bar.hpp
+++ b/include/ensmallen_bits/callbacks/progress_bar.hpp
@@ -173,9 +173,30 @@ class ProgressBar
                       (int)(steps * optimizer.BatchSize() -
                       optimizer.MaxIterations()) >= 0)
       {
-        output << "\n" << "Optimization terminated because of the entire "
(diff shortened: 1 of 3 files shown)


shrit:
@gaurav-singh1998 Just a small question, Did you keep the macro`ENS_PRINT_INFO ON` when it printed the progress bar twice? if yes, Did both of them were progressing forward at the same time? or you had only the second one printed out when the first one finished?

gaurav-singh1998:
Hi, @shrit no I didn't keep the macro `ENS_PRINT_INFO ON` while I tested this but when I did I got the result as,
```
Epoch 1/1
156/1875 [========>...........................................................................................] 8% - ETA: 15s - loss: 2049.38
157/1875 [========>...........................................................................................] 8% - ETA: 10s - loss: 2039.62
SGD: maximum iterations (10000) reached; terminating optimization.ssed to the optimizer.
```
As it can be seen that [this](https://github.com/mlpack/ensmallen/blob/master/include/ensmallen_bits/sgd/sgd_impl.hpp#L205) and the line `Optimization terminated because of the entire dataset not being passed to the optimizer.` do not get flushed properly on the screen so to remedy this I changed 
```
output << "\n" << "Optimization terminated because of the entire "
                       << "dataset not being passed to the optimizer."
                       << "\r";
```
to
```
output << "\n" << "Optimization terminated because of the entire "
                       << "dataset not being passed to the optimizer."
                       << "\n" << "\r";
```
and the output I got is
```
Epoch 1/1
156/1875 [========>...........................................................................................] 8% - ETA: 33s - loss: 2049.38
Optimization terminated because of the entire dataset not being passed to the optimizer.
157/1875 [========>...........................................................................................] 8% - ETA: 26s - loss: 2039.62
Optimization terminated because of the entire dataset not being passed to the optimizer.
SGD: maximum iterations (10000) reached; terminating optimization.
```

shrit:
@gaurav-singh1998, Actually I would prefer adding a line here explaining the message of SGD that is printed from a callback. Because I would not like to have all the output of the library for only one line especially if I am using it inside other libraries and other software. Imagine in a worse case, I might not have access to the stdout in an embedded system. What do you think?

gaurav-singh1998:
Oh okay! thanks for clarifying that @shrit. So we need to find a way so that the line gets printed only once.

include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -164,8 +164,22 @@ class ProgressBar
...
         (epochSize - step + 1) % 60 << "s - loss: " <<
         objective / (double) step <<  "\r";
-    output.flush();
 
+    if (optimizer.MaxIterations() < function.NumFunctions() &&
+        optimizer.MaxIterations() != 0)
+    {
+      if (progress == (size_t)((double)(optimizer.MaxIterations()) /
+                                (double)(function.NumFunctions()) * 100) &&
+                      (int)(steps * optimizer.BatchSize() -
```
gaurav-singh1998: Hi, @shrit I made a minor change in this line changing `step` to `steps` and got the expected output i.e.
```
Epoch 1/6
157/1875 [========>...........................................................................................] 8% - ETA: 0s - loss: 2039.62
Optimization terminated because of the entire dataset not being passed to the optimizer.
```
and now also if `ENS_PRINT_INFO` is turned on then the output is,
```
157/1875 [========>...........................................................................................] 8% - ETA: 0s - loss: 2039.62
Optimization terminated because of the entire dataset not being passed to the optimizer.
SGD: maximum iterations (10000) reached; terminating optimization.
```
Let me know what you think of this approach. Thanks.

gaurav-singh1998:
Hi, @shrit have a look at this whenever you get the chance. Thanks.

rcurtin:
Hey @gaurav-singh1998, thanks for doing this, I think it will be a lot clearer to users what's going on.  But I wonder if we can modify the way the progress bar looks so that it's a bit more obvious what happened, maybe using a different character than `=`.  Here's a few ideas:  (I don't think they're the right width, but you get the idea)

```
Epoch 1/6
157/1875 [========------------------------------------------------------------] 8% - ETA: 0s - loss: 2039.62
Optimization terminated because of the entire dataset not being passed to the optimizer.
```

```
Epoch 1/6
157/1875 [========|__________________________________________] 8% - ETA: 0s - loss: 2039.62
Optimization terminated because of the entire dataset not being passed to the optimizer.
```
```
Epoch 1/6
157/1875 [========|xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx] 8% - ETA: 0s - loss: 2039.62
Optimization terminated because of the entire dataset not being passed to the optimizer.
```

I'm not sure which one of those is best.  I might lean towards the `x`?  What do others think?

Also, we could probably improve the message.  It's true that the optimization is indeed terminated because the entire dataset did not get passed to the optimizer, but a user will probably want to know *why*.  For now we could perhaps use a simpler, more generic message like `Optimization finished before end of epoch.`---the term `terminated` could imply that it was a failure or that something's wrong, and that may not be the case.

For a longer-term solution, we could modify the `EndOptimization` callback to also take a `const std::string& reason` parameter, that would contain a little text on why the optimization has ended.  There might need to be a little bit more change to properly pass a reason for termination from anywhere, I'm not sure.  So maybe that's not the best way to do it.  But maybe I sparked some ideas, I don't know. :)

gaurav-singh1998:
Hi, @rcurtin how to keep this open?

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -21,7 +21,7 @@
 namespace ens {
 
 /**
- * CMA-ES - Covariance Matrix Adaptation Evolution Strategy is s a stochastic
+ * CMA-ES - Covariance Matrix Adaptation Evolution Strategy is a stochastic
```
rcurtin, line 24: Nice catch, thanks! :+1:

include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -164,8 +164,43 @@ class ProgressBar
...
+            output << "x";
+          }
+        }
+
+        output << "] " << progress << "% - ETA: " << (size_t) stepTimer.toc() *
+            (epochSize - step + 1) % 60 << "s - loss: " <<
+            objective / (double) step <<  "\r";
+        output << "\n" << "Optimization finished before the end of an epoch "
+                       << "because of the entire dataset not being passed to "
+                       << "the optimizer." << "\n" << "\r";
```
rcurtin, line 199: Do you think that we could improve this error message?  To a user I don't think it's clear what's going on.  I think maybe something more like this could be clearer:

```
Optimization terminated; maximum iterations reached before the end of an epoch.
```

Also, why the `\r` at the end of the line?  `\n` or `std::endl` should be sufficient.  I don't imagine that we are printing anything *after* termination such that the line needs to be rolled back.

include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -164,8 +164,43 @@ class ProgressBar
     output << "] " << progress << "% - ETA: " << (size_t) stepTimer.toc() *
         (epochSize - step + 1) % 60 << "s - loss: " <<
         objective / (double) step <<  "\r";
-    output.flush();
 
+    if (optimizer.MaxIterations() < function.NumFunctions() &&
```
rcurtin, line 168: I don't think that this is the right way to check this condition.  This will fail if the termination happens during the second epoch.  Here's some example code:

```
#include <ensmallen.hpp>

int main()
{
  arma::mat data(10, 1000, arma::fill::randu);
  arma::Row<size_t> responses(1000);
  for (size_t i = 0; i < 500; ++i)
    responses[i] = 0;
  for (size_t i = 500; i < 1000; ++i)
    responses[i] = 1;

  ens::test::LogisticRegressionFunction<> lrf(data, responses, 0.1);

  ens::Adam adam;
  adam.BatchSize() = 1;
  adam.MaxIterations() = 1500;
  arma::mat coordinates = arma::randu<arma::mat>(1, 11);
  adam.Optimize(lrf, coordinates, ens::ProgressBar());
}
```

If you try running that on this branch, you'll see:

```
$ ./test && echo ""
Epoch 1/2
1000/1000 [====================================================================================================] 100% - 0s 0ms/step - loss: 1.23509
Epoch 2/2
500/1000 [==================================================>.................................................] 50% - ETA: 0s - loss: 0.836281
$
```

(The `&& echo ""` is necessary because the last `\r` is printed unnecessarily.)

I wonder if we need to do this in a separate way, where a final call to this callback happens when the method terminates.  Perhaps in the destructor of `ProgressBar`?  I'm not sure if that could work, but it might.

Issue #178, opened 2020-03-13
Bugs in the progress bar output.


gaurav-singh1998:
#### Issue description
There are certain bugs in the progress bar callback output which needs attention.

#### Your environment

 * version of ensmallen: 2.11.4
 * operating system: Ubuntu 18.04
 * compiler: gcc 7.4.0
 * version of Armadillo: 9.800.3

#### Steps to reproduce

I ran a mlpack ffn model for three epochs with the progress bar callback incorporated on a dummy dataset the output I got was,
```
Epoch 1/192
1875/1875 [====================================================================================================] 100% - 42s 54ms/step - loss: 208.982
Epoch 2/192
1875/1875 [====================================================================================================] 100% - 44s 55ms/step - loss: 1.47048
Epoch 3/192
1875/1875 [====================================================================================================] 100% - 38s 52ms/step - loss: 0.523849
```
And when I the same model with `MaxIterations()` parameter of the optimizer set to `0` i.e. with no limits on the number of epochs the output I got was,
```
Epoch 1/0
1875/1875 [====================================================================================================] 100% - 37s 52ms/step - loss: 208.982
Epoch 2/0
1875/1875 [====================================================================================================] 100% - 43s 55ms/step - loss: 1.47048
```

#### Expected behavior

- With three epochs the expected epoch counter should be shown as 1/3, 2/3 and 3/3 and not 1/192, 2/192, 3/192

- With no limit on the number of epochs, the epoch counter should be shown, in my opinion, as 1/(no limit), 2/(no limit) and so on to be more intuitive.

- Also, the `ETA` value shouldn't be stuck on `0` when the epoch is executing. 

#### Actual behavior

From the output which has been pasted above it can be seen that there are bugs in the output which needs attention. Thank you

PR #181, opened 2020-03-15
Fixing ProgressBar()

gaurav-singh1998:
- Issue reference #178 

When a mlpack ffn model with `ProgressBar()` callback function is executed,

- If the model is run for 3 epochs the output now is,

```
Epoch 1/3
1875/1875 [====================================================================================================] 100% - 9s 37ms/step - loss: 208.982
Epoch 2/3
1875/1875 [====================================================================================================] 100% - 12s 38ms/step - loss: 1.47048
Epoch 3/3
1875/1875 [====================================================================================================] 100% - 13s 39ms/step - loss: 0.523849
```

- If the same model is made to run with `MaxIterations()` parameter of the `optimizer` set to `0` the output now is,

```
Epoch 1
1875/1875 [====================================================================================================] 100% - 9s 37ms/step - loss: 208.982
Epoch 2
1875/1875 [====================================================================================================] 100% - 9s 37ms/step - loss: 1.47048
```

- Also, the `ETA` value isn't stuck on `0` during the epoch execution here is a screenshot of output in the middle of an epoch execution,
```
157/1875 [========>...........................................................................................] 8% - ETA: 10s - loss: 2039.62
```


HISTORY.md
```diff
@@ -1,35 +1,13 @@
...
-
-  * Revamp CMake configuration
-    ([#152](https://github.com/mlpack/ensmallen/pull/152)).
+  
+  * ...
 
 ### ensmallen 2.11.2: "The Poster Session Is Full"
 ###### 2020-01-16
-  * Allow callback instantiation for SGD based optimizer
+  * Allow callback instantiation for SGD based optimizer
```
zoq: I think this was okay, without the space at the end.

gaurav-singh1998:
Hi, @zoq to check for `MaxIterations()==0` I have added this test,
```
/**
 * Make sure the ProgressBar callback will show the progress on the specified
 * output stream if the MaxIterations parameter of the optimizer is 0.
 */
TEST_CASE("ProgressBarCallbackTest", "[CallbacksTest]")
{
  SGDTestFunction f;

  StandardSGD s(0.0003, 1, 0, 1e-9, true);

  std::stringstream stream;
  s.Optimize(f, coordinates, ProgressBar(10, stream));

  REQUIRE(stream.str().length() > 0);
}
```
There is already a test which is given below present in `callbacks_test.cpp` which I think should suffice for `MaxIterations()!=0`
```
/**
 * Make sure the ProgressBar callback will show the progress on the specified
 * output stream.
 */
TEST_CASE("ProgressBarCallbackTest", "[CallbacksTest]")
{
  SGDTestFunction f;

  StandardSGD s(0.0003, 1, 10, 1e-9, true);

  std::stringstream stream;
  s.Optimize(f, coordinates, ProgressBar(10, stream));

  REQUIRE(stream.str().length() > 0);
}
```
Kindly let me know if I am wrong somewhere

HISTORY.md
gaurav-singh1998: Hi, @zoq I messed up the `HISTORY.md` in the previous commit which is resolved now. Kindly have a look at it. Thanks.

zoq:
Since the output for each case is slightly different I think it makes sense to check that the output is correct.

HISTORY.md
```diff
@@ -1,5 +1,12 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
+ * Correction in the formulation of sigma in CMA-ES
+    ([#183](https://github.com/mlpack/ensmallen/pull/183)).
+
```
rcurtin, line 5:
```suggestion
```

I think that's a part of a different PR. :)

HISTORY.md
```diff
@@ -1,5 +1,12 @@
...
+ * Fixing ProgressBar
```
rcurtin, line 6: Can we be a little more explicit about what got fixed here?  When people read the changelogs, we can probably generally assume they don't read much further.  I might suggest, e.g.,

```
 * Fix total number of epochs and time estimation for ProgressBar callback
    ([#181](https://github.com/mlpack/ensmallen/pull/181)).
```

include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -132,15 +124,24 @@ class ProgressBar
...
+        output << "Epoch " << epoch << "\n";
+        output.flush();
+        newEpoch = false;
+      }
+      else
+      {
+        output << "Epoch " << epoch << "/" << epochs << "\n";
+        output.flush();
+        newEpoch = false;
+      }
```
rcurtin, line 144: Here's a simpler way to write this; what do you think?

```suggestion
      output << "Epoch " << epoch;
      if (epochs > 0)
        output << "/" << epochs << std::endl;
      output.flush();
      newEpoch = false;
```

The functionality should be the same, so, your choice here. :+1:  The suggestion above would also allow us to keep `optimizer` commented out in the function definition.

include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -165,8 +166,6 @@ class ProgressBar
         (epochSize - step + 1) % 60 << "s - loss: " <<
         objective / (double) step <<  "\r";
     output.flush();
-
-    stepTimer.tic();
```
rcurtin, line 169: I don't think that this is the right solution.  I looked more closely, and I think the reason that the progress bar prediction output is 0 is that `stepTimer.toc()`, which returns a `double`, is being casted immediately to `size_t`.  Often, a single iteration will take far less than 1 second, meaning that `(size_t) stepTimer.toc()` takes a value of `0`, thus leading to a prediction of `0`.  If we avoid casting `stepTimer.toc()` immediately, that should be sufficient (it seems to work in my local tests).

The solution here will actually just cause the timer to accumulate, leading to extremely large and incorrect estimates for the rest of the epoch.

Also, it may be better to use an exponentially weighted average of recent computation times instead of an instantaneous guess, but we can leave that for a future PR or issue. :)

HISTORY.md
gaurav-singh1998: Hi, @rcurtin I have changed this accordingly. Kindly have a look. Thanks

include/ensmallen_bits/callbacks/progress_bar.hpp
gaurav-singh1998: Kindly have a look @rcurtin I have incorporated these changes. Thanks.

gaurav-singh1998:
Hi, @zoq please can you elaborate on how the tests should be written I am confused. Sorry I didn't get the gist of what you were saying and sorry for responding so late on this.

zoq:
No worries, right now you only check if the output isn't empty, but that would be true for `MaxIterations == 0` and `MaxIterations != 0`, so you would have to check if the output is correct. E.g. by checking if there is `Epoch 1/x` or `Epoch 1` in the output.

include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -161,8 +155,8 @@ class ProgressBar
       }
     }
 
-    output << "] " << progress << "% - ETA: " << (size_t) stepTimer.toc() *
-        (epochSize - step + 1) % 60 << "s - loss: " <<
+    output << "] " << progress << "% - ETA: " <<  fmod(stepTimer.toc() *
+        (double)(epochSize - step + 1), 60) << "s - loss: " <<
```
rcurtin, line 159: `fmod()` will return a `double`; we probably want to use the integer-valued `%`.  Maybe, e.g., `size_t(stepTimer.toc() * (epochSize - step + 1)) % 60`?  (You might want to double check if that works right; I didn't test it.)

include/ensmallen_bits/callbacks/progress_bar.hpp
gaurav-singh1998: Hi, @rcurtin I incorporated these changes suggested by you in the previous commit by didn't check it for `MaxIterations==0` case. It was not showing the desired output then but I made the necessary changes in this commit and now it is working fine. Kindly have a look.

include/ensmallen_bits/callbacks/progress_bar.hpp
gaurav-singh1998: Hi, @rcurtin I made the necessary changes kindly have a look at it now. Thanks.

gaurav-singh1998:
Hi, @rcurtin @zoq I have added the required tests kindly have a look at it now. Thanks.

tests/callbacks_test.cpp
```diff
@@ -562,3 +562,66 @@ TEST_CASE("TimerStopCallbackTest", "[CallbacksTest]")
...
+/**
+ * Make sure the ProgressBar callback will show the progress on the specified
+ * output stream if the MaxIterations parameter of the optimizer is 0.
+ */
+TEST_CASE("ProgressBarCallbackNoMaxIterationsTest", "[CallbacksTest]")
+{
+  SGDTestFunction f;
+  arma::mat coordinates = f.GetInitialPoint();
+
+  StandardSGD s(0.0003, 1, 0, 1e-9, true);
```
zoq, line 575: If we set `MaxIterations() = 0` we should use a higher threshold, to decrease the test runtime, looks like with that change we added about 20 minutes to the test/build time.

gaurav-singh1998: Hi, @zoq I increased the `tolerance` value to `DBL_MAX` should I increase the step size too? Thanks.

tests/callbacks_test.cpp
```diff
@@ -572,7 +572,7 @@ TEST_CASE("ProgressBarCallbackNoMaxIterationsTest", "[CallbacksTest]")
   SGDTestFunction f;
   arma::mat coordinates = f.GetInitialPoint();
 
-  StandardSGD s(0.0003, 1, 0, INT_MAX, true);
+  StandardSGD s(0.0003, 1, 0, DBL_MAX, true);
```
zoq, line 575: Was about to comment on this one, `DBL_MAX` makes more sense here.

gaurav-singh1998: Hi  @zoq I have changed `INT_MAX` to `DBL_MAX`. Kindly have a look at this whenever you have a chance.

gaurav-singh1998: Yes, I realised my mistake after making the commit but changed immediately afterwards. So, @zoq @rcurtin is this now ready to get merged since all the necessary changes have been made and the necessary tests have been added? Thanks

include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -161,8 +157,8 @@ class ProgressBar
       }
     }
 
-    output << "] " << progress << "% - ETA: " << (size_t) stepTimer.toc() *
-        (epochSize - step + 1) % 60 << "s - loss: " <<
+    output << "] " << progress << "% - ETA: " <<  (size_t)(stepTimer.toc() *
```
rcurtin, line 160:
```suggestion
    output << "] " << progress << "% - ETA: " << (size_t) (stepTimer.toc() *
```

tests/callbacks_test.cpp
rcurtin: What's the goal here?  This sets a tolerance of `DBL_MAX`, which means that this will probably converge within a single iteration.  I suppose that's still a valid test though, since it checks if we get any output at all, and that's all we're aiming to do in the tests.

tests/callbacks_test.cpp
```diff
@@ -562,3 +562,66 @@ TEST_CASE("TimerStopCallbackTest", "[CallbacksTest]")
...
+
+  std::stringstream stream;
+  s.Optimize(f, coordinates, ProgressBar(10, stream));
+  bool success;
+  if (stream.str().find("Epoch 1") != std::string::npos)
+  {
+    success = true;
+  }
+
+  REQUIRE(success == true);
```
rcurtin, line 603:
```suggestion
  REQUIRE(stream.str().find("Epoch 1") != std::string::npos);
```
Maybe this is cleaner?

Also we could add a simple test to ensure that the `/` isn't printed:

```
REQUIRE(stream.str().find("Epoch 1/") == std::string::npos);
```

tests/callbacks_test.cpp
```diff
@@ -562,3 +562,66 @@ TEST_CASE("TimerStopCallbackTest", "[CallbacksTest]")
...
+
+  std::stringstream stream;
+  s.Optimize(f, coordinates, ProgressBar(10, stream));
+  bool success;
+  if (stream.str().find("Epoch 1/1") != std::string::npos)
+  {
+    success = true;
+  }
+
+  REQUIRE(success == true);
```
rcurtin, line 626:
```suggestion
  REQUIRE(stream.str().find("Epoch 1/1") != std::string::npos);
```

gaurav-singh1998:
Hi, @zoq kindly have a look at this whenever you have a chance. The tests are pending for a long time. Thanks.

zoq:
Hm, looks like I can't comment on another comment, anyway:

> What's the goal here? This sets a tolerance of DBL_MAX, which means that this will probably converge within a single iteration. I suppose that's still a valid test though, since it checks if we get any output at all, and that's all we're aiming to do in the tests.

Right, this is a workaround to only run for a single iteration without setting `MaxIterations() = 1`, because the output is going to be the same if we do that, but in this case we have to check if the output is different.

zoq:
> Hi, @zoq kindly have a look at this whenever you have a chance. The tests are pending for a long time. Thanks.

Just looked the travis build, the timings are looking good to me.

zoq:
Thanks again for the contribution and for keeping up with the comments.

rcurtin:
Thanks @gaurav-singh1998!  I'll release a new version of ensmallen now.

PR #183, opened 2020-03-21
Correction in the computation of sigma for the next population in CMA-ES.

gaurav-singh1998:
I found a minor bug in the computation of sigma for the next population and fixed it.


include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -216,8 +216,7 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
     }
 
     const ElemType psNorm = arma::norm(ps[idx1]);
-    sigma(idx1) = sigma(idx0) * std::pow(
-        std::exp(cs / ds * psNorm / enn - 1), 0.3);
+    sigma(idx1) = sigma(idx0) * std::exp(cs / ds * psNorm / enn - 1);
```
favre49: Should this not be 
```suggestion
    sigma(idx1) = sigma(idx0) * std::exp(cs / ds *( psNorm / enn - 1));
```

gaurav-singh1998: Oh! thanks for pointing that out @favre49

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -216,8 +216,7 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
     }
 
     const ElemType psNorm = arma::norm(ps[idx1]);
-    sigma(idx1) = sigma(idx0) * std::pow(
-        std::exp(cs / ds * psNorm / enn - 1), 0.3);
+    sigma(idx1) = sigma(idx0) * std::exp(cs / ds *( psNorm / enn - 1));
```
favre49:
```suggestion
    sigma(idx1) = sigma(idx0) * std::exp(cs / ds * ( psNorm / enn - 1));
```

favre49:
I'll wait on the tests before approval. Good catch!

gaurav-singh1998:
Hi, @favre49 why the `momentum_sgd_test.cpp` is failing although it has nothing do with this PR?

gaurav-singh1998:
Hi, @favre49 also in the implementation I noticed that for `sigma` [here](https://github.com/mlpack/ensmallen/blob/master/include/ensmallen_bits/cmaes/cmaes_impl.hpp#L80) and for `mPositon` [here](https://github.com/mlpack/ensmallen/blob/master/include/ensmallen_bits/cmaes/cmaes_impl.hpp#L100) a vector of size 3 has been allocated but why do we need a vector of size 3 when only 2 values are needed i.e. for the current population and the next population?

favre49:
You're right, I don't think it needs to be of size 3. Do you mind changing that?

gaurav-singh1998:
Okay, @favre49 I will make those changes in this PR itself. Thanks.

zoq:
Can you update the `HISTORY.md` file.

gaurav-singh1998:
Hi, @zoq it is done. Thanks.

favre49:
Alright, I think we can merge this now. Thanks!

PR #186, opened 2020-03-27
Use new ensmallen logo

rcurtin:
This PR updates the README so that it uses the new ensmallen logo.  NumFOCUS helped us design this by providing us with an excellent graphic designer, who came up with a logo that roughly represents the task of numerical optimization.  She did way better than I could ever hope to do. :)

I've already put the logo on the website, and this puts it in the README.  Maybe it could be cleaned up, but making things look nice is not my strongest ability.  This is at least a start, though. :)

(Also, I've changed the text to use `numerical optimization` instead of `mathematical optimization`.  That's the more standard term that is probably more recognizable for many.)

Example of what it will look like: https://github.com/rcurtin/ensmallen/tree/readme-new-logo


favre49:
I'll go ahead and merge this then.

say4n:
The new logo is beautiful! 
👍

PR #185, opened 2020-03-25
remove sdp deprecated methods

jwdinius:
fixes #184

PR #192, opened 2020-03-29
Update release scripts

rcurtin:
I noticed in #185 that @favre49 wasn't able to push any extra commits during merge, and I ended up with a better understanding of Github branch protection permissions.  Anyway, what it means is that folks in the Contributors team can't push directly to `master` (and now, neither can I, I shouldn't have been able to to begin with given my propensity for typing before thinking :smile:), and so the release scripts have to be updated.

Here's the new idea for the workflow:

1. Anyone (actually even including folks not on the Contributors team) can run the `scripts/ensmallen-release.sh` script, like this:

```
$ scripts/ensmallen-release.sh rcurtin 2 13 0 "New Release Name"
```

(The Github username is necessary so that it's known where the fork is.)  This script will create a new branch in the given user's fork (i.e., `rcurtin/ensmallen:release-2.13.0` in the example), and automatically open a PR for that release.  The PR will actually include one *extra* commit to create a new HISTORY block.  These PRs will look exactly like this: #191 and will be tagged `t: release`.

2. Once that PR is approved and merged, @mlpack-bot will tag the release and automatically push a new release to Github.  @mlpack-bot will also run the `update-website-after-release.sh` script to push any changes to the website (which is then rebuilt and deployed by Jenkins).

Anyway, this is a much easier and more automatic process than before.  I wonder if I should make a *second* bot so we can quickly and automatically get two reviews for these automatic PRs, but maybe having a human do sanity checks here isn't a bad thing.  It shouldn't take more than 30 seconds to review anyway. :)  (I do think that I'll have mlpack-bot automatically approve immediately if the PR was opened by someone on the Contributors team.)

I still have some amount of homework to do: I need to implement the mlpack-bot code.  But since I don't want to release over and over and over again while I debug, I'll just do it iteratively, and each time it fails I'll just do the `git tag` and `update-website-after-release.sh` myself, which is easy enough.

### DIFF
--- a/scripts/ensmallen-release.sh
+++ b/scripts/ensmallen-release.sh
@@ -63,7 +63,6 @@ fi
 
 # Make sure 'gh' is installed.
 hub_output="`which hub`" || true;
-echo "hub output is $hub_output";
 if [ "a$hub_output" == "a" ]; then
   echo "The Hub command-line tool must be installed for this script to run" \
       "successfully.";
--- a/scripts/ensmallen-release.sh
+++ b/scripts/ensmallen-release.sh
@@ -3,66 +3,160 @@
 # Release a new version of ensmallen.
 #
 # Arguments:
-#   $ ensmallen-release.sh <major> <minor> <patch> [<name>]
+#   $ ensmallen-release.sh <github username> <major> <minor> <patch> [<name>]
 #
 # This should be run from the root of the repository.
-#
-# Make sure to update HISTORY.md manually first!
 set -e
 
-if [ "$#" -lt 3 ]; then
-  echo "At least three arguments required!";
-  echo "$ ensmallen-release.sh <major> <minor> <patch> [<name>]";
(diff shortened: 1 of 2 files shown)


birm:
I've been occasionally updating [homebrew-core's ensmallen formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ensmallen.rb) and it sure would be nice if we had a way to automate that too.

coatless:
Maybe a GitHub action?

https://github.com/marketplace/actions/bump-homebrew-formula

rcurtin:
Hmm, I wonder if we can leave mlpack-bot out of the loop here and instead use Github Actions to do the release.  I'll look into it, but I'm unclear on whether the time I've already put into learning octokit via probot would let me put together mlpack-bot to do that faster. :smile:

rcurtin:
Ok, I'll merge this now and we can see what happens next release... :)

PR #193, opened 2020-04-04
Fix for CMA-ES inconsistencies.

gaurav-singh1998:
- Issue reference #70 
This pull request aims to resolve inconsistencies that currently exist in CMA-ES implementation. I have added the boundary constraints as per [this](https://github.com/CMA-ES/c-cmaes/blob/master/src/boundary_transformation.c) code snippet which is a part of CMA-ES implementation in C by N. Hansen. I will make the appropriate changes in documentation also after a member's approval of every change I made. Thank you.


gaurav-singh1998:
Hi, @zoq kindly have a look at this whenever you get a chance. Thanks.

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -75,7 +76,8 @@ class CMAES
         const size_t batchSize = 32,
         const size_t maxIterations = 1000,
         const double tolerance = 1e-5,
-        const SelectionPolicyType& selectionPolicy = SelectionPolicyType());
+        const SelectionPolicyType& selectionPolicy = SelectionPolicyType(),
+        const double initialSigma = 0.6);
```
zoq, line 80: Hm, do you think we could just use `sigma` instead of `initialSigma` here?

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -132,14 +134,19 @@ class CMAES
   //! Modify the selection policy.
   SelectionPolicyType& SelectionPolicy() { return selectionPolicy; }
 
+  //! Get the initial step size.
+  double InitialSigma() const { return initialSigma; }
```
zoq, line 138: See comment above.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +325,89 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
   return overallObjective;
 }
 
+// Transforms the candidate into the given bounds.
+template<typename SelectionPolicyType>
+template<typename ElemType, typename BaseMatType>
```
zoq, line 330: Instead of using another template parameter here, we used:

```c+=
// Convenience typedef.
typedef typename BaseMatType::elem_type ElemType;
```

in the past.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +325,89 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+void CMAES<SelectionPolicyType>::BoundaryTransform(BaseMatType& matrix)
+{
+  double diff = (upperBound - lowerBound) / 2.0;
```
zoq, line 333: Looks like we could use `const double` here, for most of the parameter.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +325,89 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+  double xlow = lowerBound - 2 * al - diff;
+  double xup = upperBound + 2 * au + diff;
+  double r = 2 * (upperBound - lowerBound + al + au);
+
+  for (size_t col=0; col < matrix.n_cols; col++)
+  {
+    for (size_t row=0; row < matrix.n_rows; row++)
+    {
+      ElemType y = matrix(row, col);
+      //boundary transformation shift into feasible preimage.
```
zoq, line 345:
```suggestion
      // Boundary transformation shift into feasible pre-image.
```

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +325,89 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+      else if (y > upperBound - au)
+      {
+        y = (ElemType)(upperBound - (y - (upperBound + au)) *
+            (y - (upperBound + au)) / 4.0 / au);
+      }
+
+      matrix(row, col) = y;
+    }
+  }
+
```
zoq, line 377:
```suggestion
```

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +325,89 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+void CMAES<SelectionPolicyType>::BoundaryTransform(BaseMatType& matrix)
+{
+  double diff = (upperBound - lowerBound) / 2.0;
+  double al = std::min(diff, (1 + std::abs(lowerBound)) / 20.0);
+  double au = std::min(diff, (1 + std::abs(upperBound)) / 20.0);
+  double xlow = lowerBound - 2 * al - diff;
+  double xup = upperBound + 2 * au + diff;
+  double r = 2 * (upperBound - lowerBound + al + au);
+
+  for (size_t col=0; col < matrix.n_cols; col++)
```
zoq, line 340:
```suggestion
  for (size_t col = 0; col < matrix.n_cols; col++)
```

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +325,89 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+  {
+    for (size_t row=0; row < matrix.n_rows; row++)
```
zoq, line 342:
```suggestion
    for (size_t row = 0; row < matrix.n_rows; row++)
```

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +325,89 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+      }
+      if (y < lowerBound - al)
+      {
+        y += (ElemType)(2 * (lowerBound - al - y));
+      }
+      if (y > upperBound + au)
+      {
+        y -= (ElemType)(2 * (y - upperBound - au));
+      }
+      //boundary transformation
```
zoq, line 362:
```suggestion
      // Boundary transformation.
```

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +325,89 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+      }
+      if (y > xup)
+      {
+        y -= (ElemType)(r * (1 + (y - xup) / r));
+      }
+      if (y < lowerBound - al)
+      {
+        y += (ElemType)(2 * (lowerBound - al - y));
+      }
+      if (y > upperBound + au)
```
zoq, line 358: Should this be `else if`?

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -97,19 +99,19 @@ class CMAES
                                        MatType& iterate,
                                        CallbackTypes&&... callbacks);
 
-  //! Get the step size.
+  //! Get the population size.
```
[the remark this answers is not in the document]
gaurav-singh1998: Thanks to @favre49

include/ensmallen_bits/cmaes/cmaes.hpp
gaurav-singh1998: I used this parameter to tackle point number 2 in #70. If I only used sigma as you suggested here then it would have the same name as [this](https://github.com/mlpack/ensmallen/blob/master/include/ensmallen_bits/cmaes/cmaes_impl.hpp#L80) `arma` vector here and will cause an error. Please correct me if I am wrong somewhere.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
gaurav-singh1998: But if we don't use `ElemType` as the template parameter here won't it give `not declared in this scope` error. Please correct me if I am wrong.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
gaurav-singh1998: It is done.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
gaurav-singh1998: It is done.

gaurav-singh1998:
As per [this](https://github.com/mlpack/ensmallen/pull/88#discussion_r268454414) discussion, I have implemented the constructor in this pull request but wanted to ask how should mark this constructor deprecated. Should I use `Info` here?

favre49:
You can use `ens_deprecated` to mark a constructor as deprecated. Make sure to add it to the comments as well. See the diff of #185 to see how the deprecated constructors were marked.

favre49:
Does it still qualify as breaking reverse compatibility though? Since you made `initialSigma` the last member of the constructor, I don't think it does (though the behavior if the function does change slightly).

gaurav-singh1998:
Oh! okay, thanks for the clarification @favre49

include/ensmallen_bits/cmaes/cmaes_impl.hpp
zoq: If you put:

```c++
// Convenience typedef.
typedef typename BaseMatType::elem_type ElemType;
```

right at the top it should work, unless I missed something.

gaurav-singh1998:
Hi, @zoq I have made the relevant changes in the code, `HISTORY.md` and in the documentation as well. Kindly have a look at it whenever you have a chance. Thanks.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +325,90 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+
+  for (size_t col = 0; col < matrix.n_cols; col++)
+  {
+    for (size_t row = 0; row < matrix.n_rows; row++)
+    {
+      ElemType y = matrix(row, col);
+      // Boundary transformation shift into feasible pre-image.
+      if (y < xlow)
+      {
+        y += (ElemType)(r * (1 + (xlow - y) / r));
```
zoq, line 349: Shouldn't this be` y += (ElemType)(r * (1 + (int)(xlow - y) / r));` also is the cast to `ElemType` needed, not sure I can see the reason to do that?

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +325,90 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+  for (size_t col = 0; col < matrix.n_cols; col++)
+  {
+    for (size_t row = 0; row < matrix.n_rows; row++)
+    {
+      ElemType y = matrix(row, col);
+      
+      if (y < lowerBound + al)
+      {
+        y =  (ElemType)(lowerBound - al) + 2 *
+              std::pow(std::abs(al * (y - lowerBound)),0.5);
```
zoq, line 399: Any reason to avoid `sqrt`?

include/ensmallen_bits/cmaes/cmaes_impl.hpp
gaurav-singh1998: Hi, @zoq sorry for replying so late. I did it to ensure that every element of the matrix are of same types if not required I'll remove it. Thanks.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
gaurav-singh1998: Sorry, @zoq for this mistake I have rectified it. Have a look at it now.

gaurav-singh1998:
@zoq, in [this](https://github.com/gaurav-singh1998/ensmallen/blob/cmaes-fix/include/ensmallen_bits/cmaes/cmaes_impl.hpp#L45) line `ens::Warn` is used which will only throw a warning if upper bound and lower bound value is same but can you point to something specific like `Log::Fatal` of mlpack which will terminate the program because I think it would be more appropriate for this case. Thank you.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
zoq: Thanks!

zoq:
You can use assert, here is an example: https://github.com/mlpack/ensmallen/blob/45efb3e5d03babd9ef04a0b189adb8b1a0c939e6/include/ensmallen_bits/pso/init_policies/default_init.hpp#L91 let me know if that doesn't work.

gaurav-singh1998:
@zoq, I have made the relevant changes kindly have a look at it whenever you get a chance. Thanks.

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -66,6 +66,7 @@ class CMAES
    * @param maxIterations Maximum number of iterations allowed (0 means no
    *     limit).
    * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param initialSigma The initial step size
```
favre49, line 69:
```suggestion
   * @param initialSigma The initial step size.
```

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -29,15 +30,20 @@ CMAES<SelectionPolicyType>::CMAES(const size_t lambda,
...
     batchSize(batchSize),
     maxIterations(maxIterations),
     tolerance(tolerance),
-    selectionPolicy(selectionPolicy)
-{ /* Nothing to do. */ }
+    selectionPolicy(selectionPolicy),
+    initialSigma(initialSigma)
+{
+  assert(this->lowerBound != this->upperBound && "The values of "
+        "lowerbound and upperbound must be different.");
```
favre49, line 45: Either camelcase the variable names or say `lower bound`.

tests/cmaes_test.cpp
```diff
@@ -50,7 +50,7 @@ TEST_CASE("CMAESLogisticRegressionTest", "[CMAESTest]")
         responses, testResponses, shuffledResponses);
```
favre49, line 50: Should we add a test that makes sure CMAES keeps the answers within the given bounds?

gaurav-singh1998: Hi, @favre49 sorry for being late on this. Can you please elaborate on the test for this should be written? Should I consider an armadillo matrix which is out of bounds and then apply the transformation to see if the values are within the specified range? Sorry, that I didn't get the gist of what you were saying.

favre49: I meant we should add a test that ensures the optimized coordinates are within the given bounds [-20,20]. Something like `REQUIRE(coordinates[i] <=20 && coordinates[i] >= -20)`

gaurav-singh1998: Hi, @favre49 the test has been added kindly have a look whenever you get a chance. Thanks.

gaurav-singh1998:
Hello, @zoq kindly have a look at this whenever you get a chance. I have made all the necessary changes required. Thanks.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +324,90 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+// Transforms the candidate into the given bounds.
+template<typename SelectionPolicyType>
+template<typename BaseMatType>
+void CMAES<SelectionPolicyType>::BoundaryTransform(BaseMatType& matrix)
+{
+  typedef typename BaseMatType::elem_type ElemType;
+  const double diff = (upperBound - lowerBound) / 2.0;
+  const double al = std::min(diff, (1 + std::abs(lowerBound)) / 20.0);
+  const double au = std::min(diff, (1 + std::abs(upperBound)) / 20.0);
+  const double xlow = lowerBound - 2 * al - diff;
```
zoq, line 336: Do you think we should make t possible to define bounds for all coordinates similar to what has been done for the PSO method: 

https://github.com/mlpack/ensmallen/blob/70a2be8646034300cf884649f7de8f5cc9eff75d/include/ensmallen_bits/pso/pso.hpp#L91-L92

It uses a 1x1 matrix as a default, so the current implementation but you use a matrix of the same size as the coordinates as well. Let me know what you think. We can also implement the change another time and merge this as is.

gaurav-singh1998: Hi @zoq, I implemented this boundary handing function after referring the N Hansen's implementation and in my opinion, I think we should merge this as it is and if there are any issues in the future I will try to change this accordingly. Thanks.

zoq: Okay, fine for me.

gaurav-singh1998:
Hi, @zoq Are there any more changes required for this?

zoq:
No I think this is fine, I just have to double check the windows issue, but for some reason I can't access the page right now.

zoq:
Looks like the decomposition failed on the windows build:

```
[00:11:38] 1/1 Test #1: ensmallen_tests ..................***Failed  283.19 sec
[00:11:38] ensmallen version: 2.12.1 (Stir Crazy)
[00:11:38] armadillo version: 8.400.0 (Entropy Bandit)
[00:11:38] 
[00:11:38] warning: chol(): decomposition failed
[00:11:38] 
[00:11:38] warning: chol(): decomposition failed
[00:11:38] 
[00:11:38] warning: chol(): decomposition failed
[00:11:38] 
[00:11:38] warning: chol(): decomposition failed
[00:11:38] 
[00:11:38] ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[00:11:38] ensmallen_tests.exe is a Catch v2.4.1 host application.
[00:11:38] Run with -? for options
[00:11:38] 
[00:11:38] -------------------------------------------------------------------------------
[00:11:38] BoundaryConditionTestFunction
[00:11:38] -------------------------------------------------------------------------------
[00:11:38] C:\projects\ensmallen\tests\cmaes_test.cpp(180)
[00:11:38] ...............................................................................
[00:11:38] 
[00:11:38] C:\projects\ensmallen\tests\cmaes_test.cpp(194): FAILED:
[00:11:38]   REQUIRE( success == true )
[00:11:38] with expansion:
[00:11:38]   false == true
[00:11:38] 
[00:11:38] ===============================================================================
[00:11:38] test cases:   265 |   264 passed | 1 failed
[00:11:38] assertions: 12149 | 12148 passed | 1 failed
```

gaurav-singh1998:
Hi, @zoq please keep this open I want to work on it whenever I get a chance. Thanks

include/ensmallen_bits/cmaes/cmaes_impl.hpp
zoq: If you don't mind let's remove the extra cast.

zoq:

tests/cmaes_test.cpp
```diff
@@ -172,3 +172,26 @@ TEST_CASE("ApproxCMAESLogisticRegressionFMatTest", "[CMAESTest]")
...
   REQUIRE(success == true);
 }
+
+/**
+ * Test to check if the boundary transformation
+ * works as expected.
+ */
+TEST_CASE("BoundaryConditionTestFunction", "[CMAESTest]")
+{
+  SGDTestFunction f;
```
zoq, line 182: Do you mind to change the object function here, to something more complex, the SGD function has only 3 parameters.

zoq: @gaurav-singh1998 let me know if you need any help with this.

gaurav-singh1998: Hi @zoq, sorry for being so late on this(life got messed up). I want to take this up again. Can you provide some help as to what objective functions I need to use and what other changes do I need to make? Thanks.

zoq: @gaurav-singh1998 No worries I hope everything is going well. Let's test this one on the `LogisticRegression` function, here is an example - https://github.com/mlpack/ensmallen/blob/60a99f780ba24fb0403579534bc53db9e09f54c9/tests/cmaes_test.cpp#L40-L66.

gaurav-singh1998:
Hi, @zoq made the relevant changes in the test function. Kindly have a look at it now.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -99,13 +105,16 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
 
   std::vector<BaseMatType> mPosition(2, BaseMatType(iterate.n_rows,
       iterate.n_cols));
-  mPosition[0] = lowerBound + arma::randu<BaseMatType>(
-      iterate.n_rows, iterate.n_cols) * (upperBound - lowerBound);
+  BaseMatType initialVal;
+  initialVal.randu(iterate.n_rows, iterate.n_cols);
+  initialVal += (BaseMatType)(iterateIn);
+  mPosition[0] = initialVal;
```
zoq, line 111: I think it's still valid to apply the lower/upper bound here.

gaurav-singh1998: Hi, @zoq so should I remove the boundary transformation from [here](https://github.com/mlpack/ensmallen/pull/193/files#diff-98ddaa22db898dc556617924f1ad158ab9603064325059c236d604daeba55b56R117) and implement it like it was before like [here](https://github.com/mlpack/ensmallen/blob/master/include/ensmallen_bits/cmaes/cmaes_impl.hpp#L102). Thanks

zoq:
> Hi, @zoq made the relevant changes in the test function. Kindly have a look at it now.

Sorry for the slow response, I'll review the code over the weekend.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -313,6 +324,90 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
   return overallObjective;
 }
 
+// Transforms the candidate into the given bounds.
+template<typename SelectionPolicyType>
+template<typename BaseMatType>
+void CMAES<SelectionPolicyType>::BoundaryTransform(BaseMatType& matrix)
```
zoq, line 330: So I dived into the code and I think it makes sense to introduce another policy e.g. `TransformationPolicy` so we can enable/disable the boundary constraint. Let me know what you think, I can push the changes to you branch, if you like?

gaurav-singh1998: Hi, @zoq thanks for the review. I will implement the `TransformationPolicy` this week.

zoq: So basically it's similar to how the update policies are implemented https://github.com/mlpack/ensmallen/tree/master/include/ensmallen_bits/sgd/update_policies.

We basically add a new template type called `TransformationPolicy` or something like that and define an interface. I think we just need two functions at this point `Transform` and `Inverse`. So the class policy class for the `BoundaryBoxConstraint` could look like:

```c++
class BoundaryBoxConstraint
{
 public:
  void Transform(...)
  {
  }

  void Inverse(...)
  {
  }
}
```

we also a policy class that does nothing:

```c++
class EmptyTransformation
{
 public:
  void Transform(...)
  {
  }

  void Inverse(...)
  {
  }
}
```


in the `CMAES` class we can than use an instantiation of the `TransformationPolicy` and call the `Transform` and `Inverse` methods. Hopefully what I wrote above makes sense, happy to clarify anything.

PR #195, opened 2020-04-15
Update HISTORY.md

say4n:
Add missing end quote. :)


zoq:
Nice catch, I wonder if we have to update the release script: https://github.com/mlpack/ensmallen/blob/master/scripts/ensmallen-release.sh as well?

say4n:
Marcus, [this](https://github.com/mlpack/ensmallen/blob/26153353216c346f256ab918da6292156e466258/scripts/ensmallen-release.sh#L135) line in the release script adds the end quote as expected. So, no update needed for that. :)

zoq:
Easy merge, thanks again!

PR #198, opened 2020-05-04
Fix cmake package export.

Lasall:
Just some minor fixes for cmake package export:

 * Fix spelling mistake introduced in 369046f
 * Don't export compiler warning flags to interface.


zoq:
@Lasall thanks for putting this together, can you update https://github.com/mlpack/ensmallen/blob/master/HISTORY.md, to reflect the change, the rest looks good to me.

Issue #199, opened 2020-05-13
Constrained function optimization failed


JJR1234:
#### Issue description
I'm trying to optimize a simple quadratic function (x-beta)^2 with simple constraints: beta(0) > 0. And, I used the ens::AugLagrangian optimizer for that, but the results produce NaN. 

#### Your environment

 * version of ensmallen: ensmallen 2.12.1
 * operating system: Mac
 * compiler: Xcode/c++
 * version of Armadillo: Armadillo 9.870.2
 * any other environment information you think is relevant:

#### Steps to reproduce
#include <ensmallen.hpp>

class TestConstraint
{
public:
    // Construct the object with the given the design 
    // matrix and responses.
    TestConstraint(const arma::mat& X) :
    X(X) { }
    
    // Return the objective function for model parameters beta.
    double Evaluate(const arma::mat& beta)
    {
        return arma::accu(arma::square(X-beta));
    }
    
    // Compute the gradient for model parameters beta
    void Gradient(const arma::mat& beta, arma::mat& g)
    {
        g = -2 *(X - beta);
    }
    
    size_t NumConstraints() {
        return 1;
    }
    
    double EvaluateConstraint(const size_t i, const arma::mat& beta){
        if(beta(i,0)>0){
            return 0;
        }else{
            return DBL_MAX;
        }
    }
    
    void GradientConstraint(const size_t i, const arma::mat& beta, 
                            arma::mat& g){
        arma::mat g_tmp(beta.n_rows,beta.n_cols, arma::fill::zeros);
        g = g_tmp;
    }
    
    
private:
    // The design matrix.
    const arma::mat& X;
};

int main() {
    arma::mat x(2,2,arma::fill::ones);
    TestConstraint test(x);
    
    ens::AugLagrangian optfun;
    
    arma::mat beta(x.n_rows, x.n_cols, arma::fill::ones);
    
    // Run the optimization
    optfun.Optimize(test, beta);
    std::cout << beta ;
    
}

#### Expected behavior

The results should return a matrix of all ones.

#### Actual behavior

The retuned results are NaN. As I changed the input x, the results sometime is correct while sometimes are NaN. If the first element of x is negative (where the constraint condition is), the results are always NaN.

rcurtin:
[...]

PR #201, opened 2020-06-23
Fix constrained optimizer documentation

rcurtin:
This solves #199. 

Our documentation for constrained functions is actually incorrect.  The documentation implies that both "hard" and "soft" constraints are supported but this is not true.  In fact, the augmented Lagrangian optimizer (the only constrained optimizer we currently have) requires "soft" (differentiable) constraints.  Although I think it would be useful to provide support for optimizers that can take in arbitrary constraints, at the moment we don't have any, so I thought it was best to just simplify the documentation.

Also I fixed an issue where the L-BFGS optimizer will return NaNs when the starting point gives a gradient that is 0.  Essentially, for some reason, we previously did not allow termination on the first iteration due to the gradient norm being too small.  That's likely just a code artifact; I can't find any good reason to avoid that.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,12 @@
  * Fix CMake package export
     ([#198](https://github.com/mlpack/ensmallen/pull/198)).
 
+ * Clarify and fix documentation for constrained optimizers
+    ([#201](https://github.com/mlpack/ensmallen/pull/201)).
+
+ * Fix L-BFGS convergence when starting from a minimum
+    ([#201](https://github.com/mlpack/ensmallen/pull/201)).
+
 ### ensmallen 2.12.1: "Stir Crazy"
 ###### 2020-04-20
  * Fix total number of epochs and time estimation for ProgressBar callback
--- a/doc/function_types.md
+++ b/doc/function_types.md
@@ -847,17 +847,18 @@ A constrained function is an objective function `f(x)` that is also subject to
 some constraints on `x`.  (For instance, perhaps a constraint could be that `x`
 is a positive semidefinite matrix.)  ensmallen is able to handle differentiable
 objective functions of this type---so, `f'(x)` must also be computable.  Given
-some set of constraints c_0(x), ... c_M(x), we can re-express our constrained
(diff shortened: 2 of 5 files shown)


birm:
It looks like we need a corresponding mlpack pr to fix some dependent tests too?

zoq:
> It looks like we need a corresponding mlpack pr to fix some dependent tests too?

This shouldn't break anything or require any update.

rcurtin:
No, actually, it looks like mlpack's NCA implementation may depend on L-BFGS not terminating in the first iteration.  (I am not clear on why the memory checks or static code analysis jobs are failing...)

zoq:
> I am not clear on why the memory checks or static code analysis jobs are failing...

I guess for the static analysis job we have to update the job config same we did for the mlpack one, will look into it now.

zoq:

zoq:

zoq:

zoq:
The static code analysis check works now.

rcurtin:
Now that mlpack/mlpack#2479 is merged, this should build okay... let's find out...

rcurtin:

rcurtin:

rcurtin:

rcurtin:

rcurtin:
@zoq the memory checks build times out, even after 6 hours, I think because it's trying to run *all* of the tests---do you think we should adapt the script in `jenkins-conf/memory/parse-tests.py` to the ensmallen codebase?  I am not familiar with that script, but I'm happy to give it a shot if you like. :+1:

rcurtin:
Anyone willing to review this one?  I believe everything is working here and we could merge it. :+1:

zoq:
> @zoq the memory checks build times out, even after 6 hours, I think because it's trying to run _all_ of the tests---do you think we should adapt the script in `jenkins-conf/memory/parse-tests.py` to the ensmallen codebase? I am not familiar with that script, but I'm happy to give it a shot if you like. 👍

I missed this one, I'll take a look at the memory script later, should be easy, but don't think we have to wait for the memory job to pass.

zoq:
Just resolved the merge conflict.

PR #202, opened 2020-07-13
PVS - 521

zoq:
The static analysis test reports: `Such expressions using the  ',' operator are dangerous. Make sure the expression is correct.` (https://www.viva64.com/en/w/v521/) for every test, I'm not sure yet what the issue is, but I don't want to add another entry to the Jenkins file to ignore the issue, so let's use the PR to figure out what the trigger is. Starting with a new version of catch2.

### DIFF
--- a/tests/catch.hpp
+++ b/tests/catch.hpp
@@ -1,12 +1,12 @@
 /*
- *  Catch v2.4.1
- *  Generated: 2018-09-28 15:50:15.645795
+ *  Catch v2.13.0
+ *  Generated: 2020-07-12 20:07:49.015950
  *  ----------------------------------------------------------
  *  This file has been merged from multiple headers. Please don't edit it directly
- *  Copyright (c) 2018 Two Blue Cubes Ltd. All rights reserved.
+ *  Copyright (c) 2020 Two Blue Cubes Ltd. All rights reserved.
  *
  *  Distributed under the Boost Software License, Version 1.0. (See accompanying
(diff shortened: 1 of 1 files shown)


say4n:
Marcus build [#545](http://ci.mlpack.org/job/pull-requests-ensmallen-static-code-analysis/545/) passed but build [#542](http://ci.mlpack.org/job/pull-requests-ensmallen-static-code-analysis/542/) failed the static analysis test. 

I removed some unused variables from `nsga2_test.cpp` in between these two builds, could it be related? We had two `obj` variables of type `std::vector<arma::Col<double>>` earlier(1da4b449cdbaf44d98f259d845e36ecaf0646ab9), but upon changing to type `double` in the [present state](206145bd59dc2d0b5fe2b87c99b95328759ca209), the compiler did [complain](https://travis-ci.org/github/mlpack/ensmallen/builds/711115021).

Aside: Also, the compiler should have complained about unused variables in 1da4b449cdbaf44d98f259d845e36ecaf0646ab9 because of the`-Werror` in the Travis CI builds, [but it didn't](https://travis-ci.org/github/mlpack/ensmallen/jobs/709780380).

zoq:
Yes, so the issues is Catch2 related, the framework we use to write tests. It's a false warning, so we went ahead and filter out that message: https://github.com/mlpack/jenkins-conf/blob/master/static/clean_cppxml.py#L16.

Eventually this will be resolved with a update on the PVS side.

say4n:
Ah, gotcha. :)

PR #210, opened 2020-07-29
Adding MOEA/D as a multi-objective optimizer to ensmallen.

UtR491:
Hi! This PR implements [MOEA/D](https://www.researchgate.net/publication/3418989_MOEAD_A_Multiobjective_Evolutionary_Algorithm_Based_on_Decomposition) multi objective optimiser, PR #149 has been used as a reference.
TODO:

- [x] adding callbacks.

- [x] ~adding documentation.~ editing function_types pending.

- [x] looing for some more efficient implementations in some parts of the code.

- [ ] adding more decomposition approaches other than Tchebycheff approach.

- ~There is one particular difference from PR #149 however, `lowerBound` and `upperBound` in that PR are per objective but in this one are per variable of the variable space.~
- There is one more point to note, I printed the coordinates obtained from Front() for Fonseca Flemming Function and Schaffer Function and found that they converged to a particular point which changed on changing `populationSize` variable(and thus number of generations and number of weights). I also checked this for NSGA2 implementation and the same thing happened, though the points both of them converged to were different for the same problem(I didn't tinker a lot with NSGA2's test parameters).
- ~There are some places where I had to wrap a single element as a vector before calling EvaluateObjectives, I also had to overload the EvaluateObjectives function to take vectors as arguments as well. I was a bit lazy in finding a better approach, if someone could suggest something, it would be great.~ Modified the major ones. If I overlooked some please point them out.
- I also noticed a problem on changing mutationStrength from 1e-4 to 0.6, the test failed(it also did for NSGA2)~, I think this is a problem but I am not sure what is causing this. Crossover and Mutate functions are random, this might be a reason, the paper says to apply improvement heurestics, but I am not sure what that means~.
Thanks!


say4n:
> There is one particular difference from PR #149 however, lowerBound and upperBound in that PR are per objective but in this one are per variable of the variable space.

Erm, I have corrected my comment on our [earlier thread](https://github.com/mlpack/ensmallen/pull/149#discussion_r462231769), I was indeed using per variable bounds. I had missed one bound for Fonseca Flemming. You can see the latest commit here: 42ee14f08e06cb6703e246d27d59cd275a2c7449.

UtR491:
The checks are failing because the tests use problems added in pr #149. Running `valgrind --tool=memcheck --leak-check=full --track-origins=yes --num-callers=40 --error-exitcode=1 --xml=yes --xml-file=reports/tests/ensmallen.xml --log-file=reports/tests/ensmallen.memcheck --verbose  ./ensmallen_tests [MOEADTest]`
locally shows no errors.

UtR491:
Hi @zoq, now since NSGA2 is merged can you review this one too please. Static Code Analysis says `Not all members of a class are initialized inside the constructor. Consider inspecting: numObjectives.` I don't see what the problem exactly is. Thanks!

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -0,0 +1,234 @@
...
+     *
+     * @tparam ArbitraryFunctionType std::tuple of multiple function types.
+     * @tparam MatType Type of matrix to optimize.
+     * @param population The elite population.
+     * @param objectives The set of objectives.
+     * @param calculatedObjectives Vector to store calculated objectives.
+     */    
+    template<std::size_t I = 0,
+    typename MatType,
+             typename ...ArbitraryFunctionType>
```
say4n, line 172: Please indent the method declarations as per the [style guide](https://github.com/mlpack/mlpack/wiki/DesignGuidelines#method-declarations).

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -0,0 +1,234 @@
...
+               typename std::enable_if<I == sizeof...(ArbitraryFunctionType),
+               void>::type EvaluateObjectives(std::vector<MatType>&,
+                   std::tuple<ArbitraryFunctionType...>&,
+                   arma::mat&);
+
+    template<std::size_t I = 0,
+      typename MatType,
```
say4n, line 179: Please indent the method declarations as per the style guide.

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -0,0 +1,234 @@
...
+      typename ...ArbitraryFunctionType>
+        typename std::enable_if<I < sizeof...(ArbitraryFunctionType), void>::type
+        EvaluateObjectives(std::vector<MatType>& population,
+            std::tuple<ArbitraryFunctionType...>& objectives,
+            arma::mat& calculatedObjectives);
+    template<std::size_t I = 0,
```
say4n, line 185: Please add a new line before the method declaration and indent the method declarations as per the style guide.

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -0,0 +1,234 @@
...
+      typename MatType,
+      typename ...ArbitraryFunctionType>
+        typename std::enable_if<I == sizeof...(ArbitraryFunctionType), void>::type
+        EvaluateObjectives(std::vector<MatType>&,
+            std::tuple<ArbitraryFunctionType...>&,
+            std::vector<arma::vec>&);
+
+    template<std::size_t I = 0,
```
say4n, line 193: Please indent the method declarations as per the style guide.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,356 @@
...
+
+  for (arma::Col<ElemType> objective: externalPopulationFValue)
+    if (arma::accu(objective) < performance)
+      performance = arma::accu(objective);
+
+  return performance;
+}
+
+//! Perform mutation of the candidate.
+  template<typename MatType>
```
say4n, line 238: Remove the indentation. :)

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,356 @@
...
+  std::vector<MatType> externalPopulation;
+  std::vector<arma::vec> externalPopulationFValue;
+
+  // Weight vectors, where each one of them represents a decomposition.
+  std::vector<arma::vec> weights(populationSize);
+  for (size_t i = 0; i < populationSize; i++)
+    weights[i] = arma::vec(numObjectives, arma::fill::randu);
+
+  // 1.2 Storing the indices of nearest neighbours of each weight vector.
+  arma::Mat<size_t> B(populationSize, neighbourhoodSize);
```
say4n, line 75: We could perhaps rename `B` to something more descriptive like, say, `indices`. Also, if you are unsure about the naming conventions, you can read them [here](https://github.com/mlpack/mlpack/wiki/DesignGuidelines#naming-conventions)! :)

UtR491: Thanks for reviewing it so quickly. Actually I used `B` because it was used in the paper. However, now I have changed it to `weightNeighbourIndices` and also fixed the indentation problems. Also, what might be the reason behind appveyor failing? I tried to read the console output but couldn't figure anything out.

say4n: Ah, yes I thought so too! We should go with descriptive names whenever possible. :)

say4n: About the AppVeyor build, I see [C4514](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4514?view=vs-2019).

UtR491: I am still unable to figure out the problem. Can you please elaborate. I see that many functions in moead give C4514 but there are other optimisers giving the same warning. Am I missing something?

UtR491: Hi @say4n, when I run `./ensmallen_test -d yes`, `MOEADSchafferN1Test` fails (The front contains only one value -0.0900101). But when I run `./ensmallen [MOEADTest] -d yes`, `MOEADSchafferN1Test` passes (The front has many values ranging from 0.45 to 0.53). Maybe this is the reason why appveyor build is failing because on comparing the console output in this PR to that in #211, I found it is exactly the same till the point the tests start running. While implementing the code I never ran `./ensmallen_test` (because it takes some time) only checked my code with `./ensmallen_test [MOEADTest]`, because I did not know this could happen. What might be the reason?

say4n: I went through the AppVeyor logs again, at a glance, I don't see the `ctest` command (which runs the unit tests) being executed. Let me get back to you in a bit.

say4n: Ah, from what I can tell, it looks like your build is not reaching the test stage. It is terminating before executing [these](https://github.com/mlpack/ensmallen/blob/c73c92a2f57fa256d2db50b314731635147bdd9d/.appveyor.yml#L55). Now, as to why that is happening, I have no clue! 😰

say4n: What you could do is set up AppVeyor on your own fork and enter the build VM to triage the issue. Instructions [here](https://www.appveyor.com/docs/how-to/ssh-to-build-worker/). :)

UtR491: > Hi @say4n, when I run `./ensmallen_test -d yes`, `MOEADSchafferN1Test` fails (The front contains only one value -0.0900101). But when I run `./ensmallen [MOEADTest] -d yes`, `MOEADSchafferN1Test` passes (The front has many values ranging from 0.45 to 0.53). Maybe this is the reason why appveyor build is failing because on comparing the console output in this PR to that in #211, I found it is exactly the same till the point the tests start running. While implementing the code I never ran `./ensmallen_test` (because it takes some time) only checked my code with `./ensmallen_test [MOEADTest]`, because I did not know this could happen. What might be the reason?

And what might be the reason behind this?

say4n: Ah, I'll have to see. :)

UtR491: Hi @say4n, sorry for reaching back so late. I will try to figure out what happened by accessing the build VM. Since I have not done this before, I have a question. The docs you shared say 
> Add the following Bash command at the place where SSH access should be enabled:
> `curl -sflL 'https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-ssh.sh' | bash -e -`

Should I add this line just before the build seems to terminate or at the very beginning just after build_script?

UtR491: Will this work as expected?
```
environment:
  ARMADILLO_DOWNLOAD: "http://ftp.fau.de/macports/distfiles/armadillo/armadillo-8.400.0.tar.xz"
  BLAS_LIBRARY: "%APPVEYOR_BUILD_FOLDER%/OpenBLAS.0.2.14.1/lib/native/lib/x64/libopenblas.dll.a"
  BLAS_LIBRARY_DLL: "%APPVEYOR_BUILD_FOLDER%/OpenBLAS.0.2.14.1/lib/native/lib/x64/libopenblas.dll"
  APPVEYOR_SSH_KEY: <your ssh public key>

  matrix:
    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
      VSVER: Visual Studio 14 2015 Win64
      MSBUILD: C:\Program Files (x86)\MSBuild\14.0\bin\MSBuild.exe

    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
      VSVER: Visual Studio 15 2017 Win64
      MSBUILD: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\MSBuild.exe

    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
      VSVER: Visual Studio 16 2019
      MSBUILD: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin\MSBuild.exe

configuration: Release

install:
  - ps: nuget install OpenBLAS -o "${env:APPVEYOR_BUILD_FOLDER}"

build_script:
  # First, download and build Armadillo.
  - cd ..
  - appveyor DownloadFile %ARMADILLO_DOWNLOAD% -FileName armadillo.tar.xz
  - 7z x armadillo.tar.xz -so -txz | 7z x -si -ttar > nul
  - cd armadillo-8.400.0 && mkdir build && cd build
  - >
    cmake -G "%VSVER%"
    -DBLAS_LIBRARY:FILEPATH=%BLAS_LIBRARY%
    -DLAPACK_LIBRARY:FILEPATH=%BLAS_LIBRARY%
    -DCMAKE_PREFIX:FILEPATH="%APPVEYOR_BUILD_FOLDER%/armadillo"
    -DBUILD_SHARED_LIBS=OFF
    -DCMAKE_BUILD_TYPE=Release ..
  - >
    "%MSBUILD%" "armadillo.sln"
    /m /verbosity:quiet /p:Configuration=Release;Platform=x64
  - cd ../..

  # Now build ensmallen.
  - cd ensmallen && mkdir build && cd build
  - >
    cmake -G "%VSVER%"
    -DARMADILLO_INCLUDE_DIR=%APPVEYOR_BUILD_FOLDER%/../armadillo-8.400.0/include/
    -DARMADILLO_LIBRARIES=%BLAS_LIBRARY%
    -DLAPACK_LIBRARY=%BLAS_LIBRARY%
    -DBLAS_LIBRARY=%BLAS_LIBRARY%
    -DCMAKE_BUILD_TYPE=Release ..
  - >
    "%MSBUILD%" "ensmallen.sln"
    /m /verbosity:minimal /nologo /p:BuildInParallel=true
  - sh: curl -sflL 'https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-ssh.sh' | bash -e -
  # Run tests after copying libraries.
  - ps: cp C:\projects\ensmallen\OpenBLAS.0.2.14.1\lib\native\bin\x64\*.* C:\projects\ensmallen\build\
  - ctest -C Release -V --output-on-failure .
```

say4n: You should ideally set it to the place just before the failing command, try and experiment with it. You can even ssh in the `init` phase and follow the logs locally and break in when required.

You also have the `APPVEYOR_SSH_BLOCK` flag, you can use it too to pause the build till you have succefully ssh-ed and the continue by deleting the lock file.

UtR491: So the first of the three builds failed. How will I be able to access the VM, I can't seem to figure it on my own. Do I need some virtualization software or a Windows machine?

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Constructor for the MOEA/D optimizer.
+   *
+   * The default values provided here are not necessarily suitable for a
+   * given function. Therefore, it is highly recommended to adjust the
+   * parameters according to the problem.
+   *
+   * @param populationSize, the number of elements in the population.
+   * @param crossoverProb The probability that a crossover will occur.
+   * @param mutationProb The probability that a mutation will occur.
```
zoq, line 57: Looks like `mutationStrength` is missing here.

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -0,0 +1,236 @@
...
+class MOEAD {
+ public:
+  /**
+   * Constructor for the MOEA/D optimizer.
+   *
+   * The default values provided here are not necessarily suitable for a
+   * given function. Therefore, it is highly recommended to adjust the
+   * parameters according to the problem.
+   *
+   * @param populationSize, the number of elements in the population.
```
zoq, line 55:
```suggestion
   * @param populationSize The number of elements in the population.
```

The description should start with a capital letter, can to adjust the style for the other descriptions as well?

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,358 @@
...
+      distances(j) = 0;
+      for (size_t w = 0; w < numObjectives ; w++)
+      {
+        distances(j) += std::pow(weights[i](w) - weights[j](w), 2);
+      }
+      distances(j) = std::sqrt(distances(j));
+    }
+    arma::uvec sortedIndices = arma::stable_sort_index(distances, "descend");
+    for (size_t iter = 1; iter <= neighbourhoodSize; iter++)
+      weightNeighbourIndices(i, iter-1) = sortedIndices(iter);
```
zoq, line 91:
```suggestion
      weightNeighbourIndices(i, iter - 1) = sortedIndices(iter);
```

Missing space between operations.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  // 1.3 Random generation of the initial population.
+  std::vector<MatType> population(populationSize);
+  for (size_t i = 0; i < populationSize; i++)
+    population[i] = arma::randu<MatType>(iterate.n_rows, iterate.n_cols)
+        - 0.5 + iterate;
```
zoq, line 98: Can you use `{}` for multiple lines?

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,358 @@
...
+
+  // 1.3 F-value initialisation for the population.
+  std::vector<arma::vec> FValue(populationSize);
+  for (size_t i = 0; i < populationSize; i++)
+    FValue[i].resize(numObjectives);
+  EvaluateObjectives(population, objectives, FValue);
+
+  // 1.4 Initialize the ideal point z.
+  arma::vec idealPoint(numObjectives);
+  idealPoint.fill(-1/.0);
```
zoq, line 108: Not sure I get the idea of using`-1/.0`  here, can you elaborate on that?

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,358 @@
...
+  // 2 The main loop.
+  for (size_t i = 0; i < populationSize && !terminate; i++)
+  {
+    terminate |= Callback::StepTaken(*this, objectives, iterate, callbacks...);
+
+    // To generate random numbers.
+    std::default_random_engine generator;
+
+    // Distribution to choose two random parents for mutation.
+    std::uniform_int_distribution<size_t> distribution(0, populationSize-1);
```
zoq, line 120: What about using `arma::randi` that way it would be easier for a user to control the random number generator.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,358 @@
...
+    for (size_t idx = 0;idx < numObjectives;idx++)
+    {
+      idealPoint(idx) = std::max(idealPoint(idx),
+          evaluatedCandidate[0][idx]);
+    }
+
+    // 2.4 Update of the neighbouring solutions.
+    for (size_t idx = 0;idx < neighbourhoodSize;idx++)
+    {
+      if(DecomposedSingleObjective(weights[weightNeighbourIndices(i, idx)],
```
zoq, line 173:
```suggestion
      if (DecomposedSingleObjective(weights[weightNeighbourIndices(i, idx)],
```

Minor style issue.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,358 @@
...
+      bool flag = 0;
+      for (size_t idx = 0; idx < externalPopulation.size(); idx++)
+      {
+        if (Dominates(first[0], evaluatedCandidate[0]))
+        {
+          flag = 1;
+          break;
+        }
+      }
+      if(flag == 0)
```
zoq, line 214:
```suggestion
      if (flag == 0)
```

Minor style issue.

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: Update of `idealPoint` happens as follows,

idealPoint(j) = max(idealPoint(j), F<sub>j</sub>(candidate)), where F<sub>j</sub> is the j<sup>th</sup> objective.

So initially all points in idealPoint are set to -INF, so after the first iteration, idealPoint will basically be holding the values of the first candidate we add to externalPopulation. The paper said to initialize idealPoint with a problem specific method. I was not sure what it meant so I went with this approach.

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: Yes, I will replace it with `arma::randi` but what does the user having control over the random number generator exactly mean?

include/ensmallen_bits/moead/moead_impl.hpp
zoq: So might be better to initialize with `std::numeric_limits<ElemType>::max()`?

include/ensmallen_bits/moead/moead_impl.hpp
zoq: Like I can use `arma_rng::set_seed(value)` to set the seed, in case I want to start with a specifc seed.

UtR491: When I replaced `distribution` with `arma::randi`, SchafferN1Test failed, what should I do? To make sure that the implementation has no problem and the tests passing is not a fluke, I tested it with another test function, [Kursawe Function](https://en.wikipedia.org/wiki/File:Kursawe_function.pdf). It gave points with values around -19.45 and 0.44, which seems about correct (but points around this one are not plotted in the graph I saw on Wikipedia, so I can't really say). I also checked  for [Binh and Korn Function](https://en.wikipedia.org/wiki/File:Binh_and_Korn_function.pdf) which gave correct answer around 0.04 and 48. I feel that either there is some problem in the implementation or MOEA/D does not work well when there is a discontinuous front, I am looking for any errors in the implementation.

include/ensmallen_bits/moead/moead_impl.hpp
say4n: Ah, the build should pause when you set the `APPVEYOR_SSH_BLOCK` to true. It should only resume once you've ssh-ed into the machine and you've deleted the lock file that is created as per the docs [here](https://www.appveyor.com/docs/how-to/ssh-to-build-worker/). For linux you can just type in ssh `<username>@<ip-address>` to ssh to the server, on Windows, you can use something like putty.

say4n: Glad you figured it out!
(although, wrong username tag 😛)

UtR491: Locally the tests pass but they are still failing (on two builds now).

say4n: Hm, you could try tweaking the hyper-parameters. :)

UtR491: The hyper-parameter values I had set (N=1000, T=200) were very different from the ones recommended in the paper (N=150, T=10). Let's hope it passes this time.

include/ensmallen_bits/moead/moead_impl.hpp
zoq: Maybe we slightly increase the population size to make the test more stable, if that makes sense?

UtR491: Actually I had initially kept population size to be 2000 and on @say4n's suggestion, I decided to look at the paper and found the recommended value to be around 150. That was the reason I changed it. I have found [another MOEA/D implementation](https://github.com/esa/pagmo/blob/master/src/algorithm/moea_d.cpp) and I am comparing it to mine. Will keep you updated.

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: First of the three appveyor builds passed in my repository!! Hopefully it passes here too. There are some undocumented variables. I'll make the changes once(if) the builds pass here.

UtR491: The first appveyor builld here failed because of NSGA2Test and not MOEADTest. Can you please take a look?

include/ensmallen_bits/moead/moead_impl.hpp
zoq: I guess you test against the same problem/function?

UtR491: Can you please elaborate? If you mean I should test for the same problems the other implementation has tested for, I am trying to comprehend their testing methodology. It is not very straight-forward.

UtR491: I have modified the SchafferN1Test a little bit. If a negative value is encountered, it checks if it's absolute value is less than the absolute value of the minimum positive value encountered, if it is, it does not report an error. Is this allowed? I tried finding errors in the implementation but could not, so this was the only way out.

UtR491:
@mlpack-jenkins  test this please

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: Hi @zoq, can you review this please? I am very excited to get this PR merged. Thanks!

zoq: Absolutely, I'll set this on my list for later today.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,423 @@
...
+  // Controls early termination of the optimization process.
+  bool terminate = false;
+
+  // 1.1 The external population, non-dominated solutions.
+  std::vector<MatType> externalPopulation;
+  std::vector<arma::vec> externalPopulationFValue;
+  std::vector<size_t> shuffle(populationSize);
+
+  // Weight vectors, where each one of them represents a decomposition.
+  std::vector<arma::vec> weights(populationSize);
```
zoq, line 103: If we already know the size of the weights and don't change it, we should use `arma::Mat< ElemType>` instead.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,423 @@
...
+  // Number of objective functions. Represented by m in the paper.
+  numObjectives = sizeof...(ArbitraryFunctionType);
+
+  // Controls early termination of the optimization process.
+  bool terminate = false;
+
+  // 1.1 The external population, non-dominated solutions.
+  std::vector<MatType> externalPopulation;
+  std::vector<arma::vec> externalPopulationFValue;
+  std::vector<size_t> shuffle(populationSize);
```
zoq, line 100: I think we can use the `shuffel` method thats part of the function implementation, see https://github.com/mlpack/ensmallen/blob/60a99f780ba24fb0403579534bc53db9e09f54c9/include/ensmallen_bits/sgd/sgd_impl.hpp#L200-L201 for an example.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,423 @@
...
+    // To temporarily store the distance between weights(i) and each other weights.
+    arma::vec distances(populationSize);
+    for (size_t j = 0; j < populationSize; j++)
+    {
+      distances(j) = 0;
+      for (size_t w = 0; w < numObjectives ; w++)
+      {
+        distances(j) += std::pow(weights[i](w) - weights[j](w), 2);
+      }
+      distances(j) = std::sqrt(distances(j));
```
zoq, line 123: Once `weights` is of type `arma::Mat` we can simplify this using `arma::pow, arma::accu` and `.each_col()`, I can elaborate on the strategy here, if needed.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,423 @@
...
+  // 1.3 F-value initialisation for the population.
+  std::vector<arma::vec> FValue(populationSize);
+  for (size_t i = 0; i < populationSize; i++)
+    FValue[i].resize(numObjectives);
+  EvaluateObjectives(population, objectives, FValue);
+
+  // 1.4 Initialize the ideal point z.
+  std::vector<arma::vec> idealPoint(1);
+  idealPoint[0].resize(numObjectives);
+  std::vector<MatType> iterateWrapper(1);
```
zoq, line 147: Maybe I missed something, but it looks like we don't change the size of `iterateWrapper` and could directly use `MatType`. Same for `idealPoint`.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,423 @@
...
+  {
+    std::random_shuffle(shuffle.begin(), shuffle.end());
+    for (size_t i : shuffle)
+    {
+      terminate |= Callback::StepTaken(*this, objectives, iterate, callbacks...);
+
+      // 2.1 Randomly select two indices in weightNeighbourIndices(i) and use them
+      // to make a child.
+      size_t k = weightNeighbourIndices(i, arma::randi(arma::distr_param(0,  neighbourhoodSize-1))),
+             l = weightNeighbourIndices(i, arma::randi(arma::distr_param(0,  neighbourhoodSize-1)));
```
zoq, line 166: We should make sure `k != l`.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,423 @@
...
+    if (arma::accu(objective) < performance)
+      performance = arma::accu(objective);
+  }
+
+  return performance;
+}
+
+//! Perform mutation of the candidate.
+template<typename MatType>
+inline void MOEAD::Mutate(MatType& child,
```
zoq, line 282: Is this the Differential Evolution (DE) operator and polynomial mutation procedure as in Li and Zhang (2009)?

UtR491: Yes. Actually I used the implementation I had mentioned earlier to implement it, so I forgot to cite the paper.

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: I had to wrap iterate and idealPoint in single element vectors because I was passing them to `EvaluateObjectives `which accepted a vector as arguments. I could alternatively overload the function to take single elements as arguments.

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: Should I add a Shuffle function to problem or use arma::shuffle and arma::linspace as used in the Shuffle function above?

UtR491:
@zoq I was going through the gsoc ideas list and was thinking can't we employ parallelization to parts of ensmallen as well? I saw some tutorials on using openMP and thought there are some parts in the codebase (for e.g. evolutionary algorithms) which could benefit from it, since they are kind of doing same task over and over again (i.e. SPMD). I am still on very new to parallelization, maybe this is not possible, just thought of pitching this to you.

include/ensmallen_bits/moead/moead_impl.hpp
zoq: Personally I would rely on the shuffel function that is part of the problem, for example:

https://github.com/mlpack/mlpack/blob/e11c3e320c4008a3c6ed16abeef69d53c0584366/src/mlpack/methods/ann/ffn.hpp#L282

implement the `Shuffel` function

https://github.com/mlpack/mlpack/blob/master/src/mlpack/methods/logistic_regression/logistic_regression_function.hpp#L57

as well, basically all methods in mlpack, so I think it makes sense to reuse that functionality.

zoq:
> @zoq I was going through the gsoc ideas list and was thinking can't we employ parallelization to parts of ensmallen as well? I saw some tutorials on using openMP and thought there are some parts in the codebase (for e.g. evolutionary algorithms) which could benefit from it, since they are kind of doing same task over and over again (i.e. SPMD). I am still on very new to parallelization, maybe this is not possible, just thought of pitching this to you.

Absolutely, we already have one optimizer that uses OpenMP:

https://github.com/mlpack/ensmallen/blob/d15772b26810861dee5bc528bf79fc10b7db0dbe/include/ensmallen_bits/parallel_sgd/parallel_sgd_impl.hpp#L146-L153

we just have to be careful in our decision what part would benefit from parallelization using OpenMP and which doesn't, but we can always test things out and time them.

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: @zoq, since there is more than one objective, and we have access to the object of each objective via the tuple, I have implemented Shuffle() inside the first problem. Is this fine? I was wondering, isn't Shuffle() in SGDTestFunction used to change the order in which the three components of the function are added, i.e. wasn't it affecting the problem? The shuffle in our case is affecting the optimizer itself and not the problem, so I am still not sure if this is correct.

UtR491: @zoq, can you review this please? I thought I would update the docs once everything is finalised.

zoq: `SGDTestFunction` that is correct, but I'm almost sure that's what was implemented but was part of the optimizer; will double check to make sure this is still correct.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,418 @@
...
+  arma::Mat<ElemType> weights(numObjectives, populationSize, arma::fill::randu);
+
+  // 1.2 Storing the indices of nearest neighbours of each weight vector.
+  arma::Mat<size_t> weightNeighbourIndices(populationSize, neighbourhoodSize);
+  for (size_t i = 0; i < populationSize; i++)
+  {
+    // To temporarily store the distance between weights(i) and each other weights.
+    arma::vec distances(populationSize);
+    for (size_t j = 0; j < populationSize; j++)
+      distances(j) = std::sqrt(arma::accu(arma::pow(weights.col(i)-weights.each_col(), 2)));
```
zoq, line 112:
```suggestion
      distances(j) = std::sqrt(arma::accu(arma::pow(weights.col(i) - weights.each_col(), 2)));
```

Minor style fix, missing space between minus operator.

include/ensmallen_bits/moead/moead_impl.hpp
zoq: We could just update `EvaluateObjectives` to take `arma::mat` instead of `std::vector<arma::vec>` as input, it's an internal optimizer function, so there is no problem and `EvaluateObjectives` currently unpacks the `std::vector` and calls `Evaluate` on that function, so we can just use `Evaluate(.col())` instead.

UtR491: I thought of comparing the performance value of the fronts obtained from MOEA/D and NSGA2, the values for the former were in the correct range but those for the latter were all 0! The least possible values can be 0.98 for Fonseca Fleming function and 2 for Schaffer N1 function.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,383 @@
...
+  // Nothing to do here.
+}
+
+//! Evaluate the objectives for the entire population.
+template<std::size_t I,
+  typename MatType,
+  typename ...ArbitraryFunctionType>
+  typename std::enable_if<I < sizeof...(ArbitraryFunctionType), void>::type
+MOEAD::EvaluateObjectives(
+    std::vector<MatType>& population,
```
zoq, line 370: So it looks like it's save to just use `MatType` here instead of passing a vector, since the population we pass has only a single element. What I'm trying to do here is keep the code as simple as possible, which involves removing unnecessary containers.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,383 @@
...
+      }
+
+      // 2.4 Update of the neighbouring solutions.
+      for (size_t idx = 0;idx < neighbourhoodSize;idx++)
+      {
+        if (DecomposedSingleObjective(weights.col(weightNeighbourIndices(i, idx)),
+              idealPoint.col(0), evaluatedCandidate.col(0))
+            <= DecomposedSingleObjective(
+              weights.col(weightNeighbourIndices(i,idx)),
+              idealPoint.col(0), FValue.col(weightNeighbourIndices(i, idx))))
```
zoq, line 200: Minor style issue, the lines should align with the beginning of `DecomposedSingleObjective`.

include/ensmallen_bits/moead/moead_impl.hpp
zoq: Ahh, nevermind.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,383 @@
...
+  for (i = 0; i < numObjectives; i++)
+  {
+    if ( first(i) > second(i) )
+    {
+      return false;
+    }
+    if( first(i) < second(i) )
+      flag = 1;
+  }
+  if ( i == numObjectives && flag == 1 )
```
zoq, line 341: We can remove the `i == numObjectives` condition here, because it's always true, which also allows us to write `for (size_t i = 0; i < numObjectives; ++i)`.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,383 @@
...
+  {
+    return true;
+  }
+  else
+  {
+    return false;
+  }
```
zoq, line 348: We could remove the else case and just write `return false;` here make it a little bit shorter.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,383 @@
...
+  return value;
+}
+
+inline bool MOEAD::Dominates(const arma::vec& first,
+    const arma::vec& second)
+{
+  size_t i; int flag = 0;
+  for (i = 0; i < numObjectives; i++)
+  {
+    if ( first(i) > second(i) )
```
zoq, line 334:
```suggestion
    if (first(i) > second(i))
```

Minor style issue.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,383 @@
...
+    {
+      return false;
+    }
+    if( first(i) < second(i) )
```
zoq, line 338:
```suggestion
    }

    if(first(i) < second(i))
```

Minor style issue.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,383 @@
...
+    const arma::vec& evaluatedCandidate)
+{
+  double value = weights(0)*std::abs(evaluatedCandidate(0)
+      - idealPoint(0));
+  for (size_t i = 1; i < numObjectives; i++)
+  {
+    value = std::min(value, weights(i)*std::abs(
+          evaluatedCandidate(i) - idealPoint(i)));
+  }
+  return value;
```
zoq, line 325: `return arma::min(weights % arma::abs(evaluatedCandidate - idealPoint);` should do the trick here as well.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,383 @@
...
+inline void MOEAD::Mutate(MatType& child,
+    const double& rate,
+    const arma::vec& lowerBound,
+    const arma::vec& upperBound)
+{
+  size_t numVariables = lowerBound.n_elem;
+  double rnd, delta1, delta2, mutationPower, deltaq;
+  double y, yl, yu, val, xy;
+
+  for(size_t j=0; j < numVariables; j++)
```
zoq, line 280:
```suggestion
  for(size_t  j =0; j < numVariables; ++j)
```

There are other places where we should put spaces around an operation as  well.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,383 @@
...
+        externalPopulationFValue.push_back(evaluatedCandidate.col(0));
+      }
+    }
+  }
+  bestFront = std::move(externalPopulation);
+
+  Callback::EndOptimization(*this, objectives, iterate, callbacks...);
+
+  ElemType performance = std::numeric_limits<ElemType>::max();
+
```
zoq, line 259:
```suggestion
```

I would remove the line here, because I see this as a block where we compute the final objective.

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: Not sure why, but replacing the for each loop in the end with a normal for loop resulted in expected values of performance (in NSGA2).

UtR491:
@mlpack-jenkins test this please.

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: Can I implement other decomposition approaches later? This one has been going on for quite some time now.:sweat_smile:

UtR491: I don't know why this build is failing, the one on my branch [passed](https://ci.appveyor.com/project/UtR491/ensmallen).

UtR491: Hi @zoq, can you take some time out to review this, please? I thought the appveyor build failing was just a one-off, but it failed twice in a row, I tried to figure out why but could not. Also, [it passed](https://ci.appveyor.com/project/UtR491/ensmallen) on my branch so I am not sure what the problem is.:confused:

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -0,0 +1,285 @@
...
+ * volume={13},
+ * number={2},
+ * pages={284-302},}
+ * @endcode
+ *
+ * MOEA/D can optimize arbitrary multi-objective functions. For more details,
+ * see the documentation on function types included with this distribution or
+ * on the ensmallen website.
+ */
+
```
zoq, line 54: We can remove the empty line here.

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -0,0 +1,285 @@
...
+   */
+    MOEAD(const size_t populationSize = 100,
+        const size_t numGeneration = 100,
+        const double crossoverProb = 0.6,
+        const double mutationProb = 0.3,
+        const double mutationStrength = 1e-3,
+        const size_t neighbourhoodSize = 50,
+        const double distributionIndex = 0.5,
+        const double lowerBound = 0,
+        const double upperBound = 1);
```
zoq, line 117: The arguments should align with the first one.

include/ensmallen_bits/moead/moead_impl.hpp
zoq: Hm looks like the force-push removed the build log?

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+               weights.col(weightNeighbourIndices(i,idx)),
+               idealPoint.col(0), FValue.col(weightNeighbourIndices(i, idx))))
+        {
+          population.at(weightNeighbourIndices(i, idx)) = candidate[0];
+          FValue.col(weightNeighbourIndices(i, idx)) = evaluatedCandidate.col(0);
+        }
+      }
+
+      // 2.5 Updating External Population.
+      if ( !externalPopulation.empty() )
```
zoq, line 206:
```suggestion
      if (!externalPopulation.empty())
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+  {
+    double determiner = arma::randu();
+    if(determiner <= rate)
+    {
+      y = child[j];
+      yl = lowerBound(j);
+      yu = upperBound(j);
+      delta1 = (y-yl)/(yu-yl);
+      delta2 = (yu-y)/(yu-yl);    
+      rnd=arma::randu();
```
zoq, line 287:
```suggestion
      rnd = arma::randu();
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+      mutationPower=1/distributionIndex;
```
zoq, line 288:
```suggestion
      mutationPower= 1 / distributionIndex;
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+      if (rnd <= 0.5)
+      {
+        xy = 1.0-delta1;
+        val = 2.0 * rnd + (1.0 - 2.0 * rnd) *
+            (std::pow(xy, (distributionIndex + 1.0)));
+        deltaq =  std::pow(val,mutationPower) - 1.0;
```
zoq, line 294:
```suggestion
        deltaq = std::pow(val, mutationPower) - 1.0;
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+      }
+      else
+      {
+        xy = 1.0-delta2;
```
zoq, line 298:
```suggestion
        xy = 1.0 - delta2;
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+      y = child[j];
+      yl = lowerBound(j);
+      yu = upperBound(j);
+      delta1 = (y-yl)/(yu-yl);
+      delta2 = (yu-y)/(yu-yl);    
+      rnd=arma::randu();
+      mutationPower=1/distributionIndex;
+      if (rnd <= 0.5)
+      {
+        xy = 1.0-delta1;
```
zoq, line 291:
```suggestion
        xy = 1.0 - delta1;
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+        val = 2.0 * rnd + (1.0 - 2.0 * rnd) *
+            (std::pow(xy, (distributionIndex + 1.0)));
+        deltaq =  std::pow(val,mutationPower) - 1.0;
+      }
+      else
+      {
+        xy = 1.0-delta2;
+        val = 2.0 * (1.0 - rnd) + 2.0 * (rnd-0.5) *
```
zoq, line 299:
```suggestion
        val = 2.0 * (1.0 - rnd) + 2.0 * (rnd - 0.5) *
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+  for(size_t j = 0; j < numVariables; ++j)
+  {
+    double determiner = arma::randu();
+    if(determiner <= rate)
+    {
+      y = child[j];
+      yl = lowerBound(j);
+      yu = upperBound(j);
+      delta1 = (y-yl)/(yu-yl);
+      delta2 = (yu-y)/(yu-yl);
```
zoq, line 286:
```suggestion
      delta1 = (y - yl) / (yu - yl);
      delta2 = (yu - y) / (yu - yl);    
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+inline void MOEAD::Mutate(MatType& child,
+    const double& rate,
+    const arma::vec& lowerBound,
+    const arma::vec& upperBound)
+{
+  size_t numVariables = lowerBound.n_elem;
+  double rnd, delta1, delta2, mutationPower, deltaq;
+  double y, yl, yu, val, xy;
+
+  for(size_t j = 0; j < numVariables; ++j)
```
zoq, line 277:
```suggestion
  for (size_t j = 0; j < numVariables; ++j)
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+  {
+    double determiner = arma::randu();
+    if(determiner <= rate)
```
zoq, line 280:
```suggestion
    if (determiner <= rate)
```

tests/moead_test.cpp
```diff
@@ -0,0 +1,180 @@
...
+    for (arma::mat solution: bestFronts)
+    {
+      double val = arma::as_scalar(solution);
+      minimumPositive = std::min(minimumPositive, val>=0?val:1000);
+      if ((val < 0.00 && std::abs(val) >= minimumPositive) || val > 2.0)
+      {
+        allInRange = false;
+        break;
+      }
+    }
```
zoq, line 178: Looks the indentation is off my two spaces.

tests/moead_test.cpp
```diff
@@ -0,0 +1,180 @@
...
+    for (arma::mat solution: bestFronts)
+    {
+      const double val = arma::as_scalar(solution);
+      minimumPositive = std::min(minimumPositive, val>=0?val:1000);
+      if ((val < 0.00 && std::abs(val) >= minimumPositive) || val > 2.0)
+      {
+        allInRange = false;
+        break;
+      }
+    }
```
zoq, line 142: Looks the indentation is off my two spaces.

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: Yes. Actually I just dropped the commits that caused the failure. Maybe, it was some timeout issue because both the times, it failed after running for 90 minutes.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+//! Perform mutation of the candidate.
+template<typename MatType>
+inline void MOEAD::Mutate(MatType& child,
+    const double& rate,
+    const arma::vec& lowerBound,
+    const arma::vec& upperBound)
+{
+  size_t numVariables = lowerBound.n_elem;
+  double rnd, delta1, delta2, mutationPower, deltaq;
+  double y, yl, yu, val, xy;
```
say4n, line 275: I was wondering if we could have a bit more descriptive names, if these are the symbols used in the paper, it is fine! :)

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -0,0 +1,284 @@
...
+namespace ens {
+
+/**
+ * This class implements the MOEA/D algorithm with Differential Evolution
+ * crossover. Step numbers used in different parts of the implementation
+ * correspond to the step number used in the original algorithm by the author.
+ *
+ * For more information, see the following:
+ * @code
+ * @article{article,
```
say4n, line 29: The citations should be kept simple according to the [style guide](https://github.com/mlpack/mlpack/wiki/DesignGuidelines#citations), we don't need a bunch of fields, the author, year, pages, title, journal/conference should be sufficient!

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: Actually I did not see the paper first hand. I used [this](https://github.com/esa/pagmo/blob/master/src/algorithm/moea_d.cpp) implementation as a reference. I have made them more descriptive.

say4n: Ah, I see.

zoq: Ahh, that's why I had some issues reconstructing the mutation step from the paper.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+  typename ...ArbitraryFunctionType>
+  typename std::enable_if<I < sizeof...(ArbitraryFunctionType), void>::type
+MOEAD::EvaluateObjectives(
+    std::vector<MatType>& population,
+    std::tuple<ArbitraryFunctionType...>& objectives,
+    arma::mat& calculatedObjectives)
+{
+  for (size_t i = 0; i < population.size(); ++i)  {
+    calculatedObjectives(I, i) = std::get<I>(objectives).Evaluate(population[i]);
+    EvaluateObjectives<I+1, MatType, ArbitraryFunctionType...>(population, objectives,
```
zoq, line 363:
```suggestion
    EvaluateObjectives<I + 1, MatType, ArbitraryFunctionType...>(population, objectives,
```

include/ensmallen_bits/problems/schaffer_function_n1.hpp
```diff
@@ -72,6 +72,12 @@ class SchafferFunctionN1
     {
         return std::pow(coords[0], 2);
     }
+    arma::Col<size_t> Shuffle(size_t& populationSize)
+    {
+      return arma::shuffle(arma::linspace<arma::Col<size_t> >(0,
+            populationSize -1));
```
zoq, line 78:
```suggestion
          populationSize -1));
```

Should be tabbed twice (4 spaces) - https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping

tests/moead_test.cpp
```diff
@@ -0,0 +1,180 @@
...
+  opt.Optimize(objectives, coords);
+  std::vector<arma::mat> bestFronts = opt.Front();
+
+  bool allInRange = true;
+  double minimumPositive = 1000;
+
+  for (arma::mat solution: bestFronts)
+  {
+    const double val = arma::as_scalar(solution);
+    minimumPositive = std::min(minimumPositive, val>=0?val:1000);
```
zoq, line 136:
```suggestion
    minimumPositive = std::min(minimumPositive, val>=0 ? val : 1000);
```

tests/moead_test.cpp
```diff
@@ -0,0 +1,180 @@
...
+  opt.Optimize(objectives, coords);
+  std::vector<arma::mat> bestFronts = opt.Front();
+
+  bool allInRange = true;
+  double minimumPositive = 1000;
+
+  for (arma::mat solution: bestFronts)
+  {
+    double val = arma::as_scalar(solution);
+    minimumPositive = std::min(minimumPositive, val>=0?val:1000);
```
zoq, line 172:
```suggestion
    minimumPositive = std::min(minimumPositive, val>=0 ? val : 1000);
```

include/ensmallen_bits/problems/fonseca_fleming_function.hpp
```diff
@@ -77,6 +78,12 @@ class FonsecaFlemingFunction
              -pow(static_cast<double>(coords[2]) - 1.0 / sqrt(3.0), 2.0)
         );
     }
+    arma::Col<size_t> Shuffle(size_t& populationSize)
+    {
+      return arma::shuffle(arma::linspace<arma::Col<size_t> >(0,
+            populationSize -1));
```
zoq, line 84:
```suggestion
          populationSize -1));
```

Should be tabbed twice (4 spaces) - https://github.com/mlpack/mlpack/wiki/DesignGuidelines#line-length-and-wrapping

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+template<std::size_t I,
+  typename MatType,
+  typename ...ArbitraryFunctionType>
+  typename std::enable_if<I < sizeof...(ArbitraryFunctionType), void>::type
+MOEAD::EvaluateObjectives(
+    std::vector<MatType>& population,
+    std::tuple<ArbitraryFunctionType...>& objectives,
+    arma::mat& calculatedObjectives)
+{
+  for (size_t i = 0; i < population.size(); ++i)  {
```
zoq, line 361:
```suggestion
  for (size_t i = 0; i < population.size(); ++i)
  {
```

The opening `{` should be on a new line.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+  arma::mat idealPoint(numObjectives, 1);
+  std::vector<MatType> iterateWrapper(1);
+  iterateWrapper[0].resize(iterate.n_rows, iterate.n_cols);
+  iterateWrapper[0]=iterate;
+  EvaluateObjectives(iterateWrapper, objectives, idealPoint);
+
+  terminate |= Callback::BeginOptimization(*this, objectives, iterate, callbacks...);
+
+  // 2 The main loop.
+  for(size_t g = 0; g < numGeneration; ++g)
```
zoq, line 138:
```suggestion
  for (size_t g = 0; g < numGeneration; ++g)
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+    {
+      terminate |= Callback::StepTaken(*this, objectives, iterate, callbacks...);
+
+      // 2.1 Randomly select two indices in weightNeighbourIndices(i) and use them
+      // to make a child.
+      size_t k = weightNeighbourIndices(i, arma::randi(arma::distr_param(0,  neighbourhoodSize-1))),
+             l = weightNeighbourIndices(i, arma::randi(arma::distr_param(0,  neighbourhoodSize-1)));
+      if(k == l)
+      {
+        if(k == neighbourhoodSize-1)
```
zoq, line 151:
```suggestion
        if(k == neighbourhoodSize - 1)
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+  {
+    shuffle = std::get<0>(objectives).Shuffle(populationSize);
+    for (size_t i : shuffle)
+    {
+      terminate |= Callback::StepTaken(*this, objectives, iterate, callbacks...);
+
+      // 2.1 Randomly select two indices in weightNeighbourIndices(i) and use them
+      // to make a child.
+      size_t k = weightNeighbourIndices(i, arma::randi(arma::distr_param(0,  neighbourhoodSize-1))),
+             l = weightNeighbourIndices(i, arma::randi(arma::distr_param(0,  neighbourhoodSize-1)));
```
zoq, line 148:
```suggestion
      size_t k = weightNeighbourIndices(i, arma::randi(arma::distr_param(0,  neighbourhoodSize - 1))),
             l = weightNeighbourIndices(i, arma::randi(arma::distr_param(0,  neighbourhoodSize - 1)));
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+  // Convenience typedefs.
+  typedef typename MatType::elem_type ElemType;
+
+  if(lowerBound.size() != upperBound.size())
+    throw std::logic_error("MOEAD::Optimize(): size of lowerBound and upperBound "
+        "must be equal.");
+
+  if(lowerBound.size() != iterate.n_elem)
+    throw std::logic_error("MOEAD::Optimize(): there should be a lower bound and "
+        "an upper bound for each variable in the initial point.");
```
zoq, line 87: Do you mind to use `{}` here to improve readability.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -0,0 +1,369 @@
...
+
+  // Check if lower bound is a vector of a single dimension.
+  if (upperBound.n_rows == 1)
+    upperBound = upperBound(0, 0) * arma::ones(iterate.n_rows, iterate.n_cols);
+  // Convenience typedefs.
+  typedef typename MatType::elem_type ElemType;
+
+  if(lowerBound.size() != upperBound.size())
+    throw std::logic_error("MOEAD::Optimize(): size of lowerBound and upperBound "
+        "must be equal.");
```
zoq, line 83: Do you mind to use `{}` here to improve readability.

zoq (commented):
Some minor style comments, I will have to reread the paper, and check the mutation step.

include/ensmallen_bits/moead/moead_impl.hpp
UtR491: Yes. Will do.

UtR491: So what I found out is, `Mutate` implements polynomial mutation and the crossover step is just plain crossover, where each gene of the variable is chosen from one of the two randomly selected members. This isn't DE crossover at all. I should have gone through the paper earlier but could not get my hands on it. So should I implement the crossover step as DE crossover or just remove it from the citation and docs? Sorry for the mess up.:bow:

rcurtin:
Is this PR just waiting on review or is there anything left to be done?  I can read the paper and provide a review if needed. :+1:

say4n:
Ryan as far as I can recall, it is waiting on a review. :)

UtR491:
Actually, there were some major changes to be made but I couldn't find time to do them. I think I should close this for now and reopen it when I have come up with something correct.

PR #211, opened 2020-08-02
DemonSGD & DemonAdam

zoq:
DemonSGD & DemonAdam - https://arxiv.org/abs/1910.04952


zoq:
> Strange I tested the tests in a loop with different random seeds without any issue.

Okay so the issue is that with the cast here: https://github.com/mlpack/ensmallen/pull/211/commits/a57ff2d510e901dfa89cbce17caf1985cc0a80b8 the decay was always set to zero.

zoq:

zoq:

zoq:

zoq:

zoq:

zoq:

zoq:

rcurtin:
Let's keep this open---I have been meaning to review it but haven't quite gotten to it yet. :+1:

doc/optimizers.md
```diff
@@ -202,7 +202,7 @@ with _`UpdateRule`_` = AdamUpdate`.
 | `double` | **`beta1`** | Exponential decay rate for the first moment estimates. | `0.9` |
 | `double` | **`beta2`** | Exponential decay rate for the weighted infinity norm estimates. | `0.999` |
 | `double` | **`eps`** | Value used to initialize the mean squared gradient parameter. | `1e-8` |
-| `size_t` | **`max_iterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+| `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
```
rcurtin, line 205: Nice catch! :+1:

doc/optimizers.md
```diff
@@ -744,6 +744,119 @@ optimizer.Optimize(f, coordinates);
  * [Differential Evolution in Wikipedia](https://en.wikipedia.org/wiki/Differential_Evolution)
  * [Arbitrary functions](#arbitrary-functions)
 
+## DemonAdam
+
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+DemonAdam is an Adam based optimzer. DemonAdam is motivated by decaying the
```
rcurtin, line 751:
```suggestion
DemonAdam is an Adam based optimizer. DemonAdam is motivated by decaying the
```

doc/optimizers.md
```diff
@@ -744,6 +744,119 @@ optimizer.Optimize(f, coordinates);
...
+ * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
+ * [SGD](#standard-sgd)
+ * [Decaying momentum helps neural network training](https://arxiv.org/abs/1910.04952)
+ * [Differentiable separable functions](#differentiable-separable-functions)
+
+## DemonSGD
+
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+DemonSGD is an SGD based optimzer. DemonSGD is motivated by decaying the total
```
rcurtin, line 811:
```suggestion
DemonSGD is an SGD based optimizer. DemonSGD is motivated by decaying the total
```

include/ensmallen_bits/demon_adam/demon_adam.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+using DemonAdaMax = DemonAdamType<AdaMaxUpdate>;
+
+using DemonAMSGrad = DemonAdamType<AMSGradUpdate>;
+
+using DemonNadam = DemonAdamType<NadamUpdate>;
+
+using DemonNadaMax = DemonAdamType<NadaMaxUpdate>;
+
+using DemonOptimisticAdam = DemonAdamType<OptimisticAdamUpdate>;
```
rcurtin, line 202: Maybe it's worth at least mentioning these typedefs in the `DemonAdam` and `DemonSGD` documentation?

include/ensmallen_bits/demon_adam/demon_adam_update.hpp
```diff
@@ -0,0 +1,169 @@
...
+      T(momentumIterations),
+      betaInit(momentum),
+      t(0),
+      adamUpdateInst(adamUpdate)
+  {
+    // Make sure the momentum iterations parameter is non-zero.
+    assert(momentumIterations != 0 && "The number of iterations before the "
+        "momentum will decay is zero, make sure the max iterations and "
+        "batch size parameter is set correctly. "
+        "Default: momentumIterations = maxIterations * batchSize.");
```
rcurtin, line 62: Shouldn't this be `maxIterations / batchSize` as the default?  Otherwise the number of decay steps for momentum won't match the number of steps taken during optimization.

include/ensmallen_bits/demon_sgd/demon_sgd_update.hpp
```diff
@@ -0,0 +1,139 @@
...
+     * @param cols Number of columns in the gradient matrix.
+     */
+    Policy(DemonSGDUpdate& parent,
+           const size_t /* rows */,
+           const size_t /* cols */) :
+      parent(parent)
+    { /* Nothing to do here */ }
+
+    /**
+     * Update step for DamonSGD.
```
rcurtin, line 96:
```suggestion
     * Update step for DemonSGD.
```

rcurtin (commented):
Nice, looks good! Only one real comment about the default value for `T`. It's surprising both that the technique seems to yield such nice results, and also that the authors did such comprehensive comparisons... most of these papers they seem to only compare with plain SGD on logistic regression on MNIST...

include/ensmallen_bits/demon_adam/demon_adam.hpp
zoq: Agreed, good idea.

include/ensmallen_bits/demon_adam/demon_adam_update.hpp
zoq: Nice catch!

include/ensmallen_bits/demon_adam/demon_adam.hpp
```diff
@@ -0,0 +1,206 @@
...
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_DEMON_ADAM_DEMON_ADAM_HPP
+#define ENSMALLEN_DEMON_ADAM_DEMON_ADAM_HPP
+
+#include <ensmallen_bits/sgd/sgd.hpp>
+#include "demon_adam_update.hpp"
```
zoq, line 16: I'll update the includes today.

conradsnicta: We may need to sort out which format to use (`<xyz>` vs `"xyz"`). 

Use of `<...>` caused problems within the context of RcppArmadillo (see for example https://github.com/RcppCore/RcppArmadillo/pull/281, https://github.com/RcppCore/RcppArmadillo/pull/315, and related issues). The tldr is that for people that had two versions of Armadillo (one via RcppArmadillo, and one via plain Armadillo), there were clashes between the headers from the two versions.  We settled on `"xyz"` to avoid this.

Within the context of ensmallen, rather than doing something like `#include <ensmallen_bits/sgd/sgd.hpp>`, perhaps we can use `..` to specify one level back in the directory structure, ie. `#include "../sgd/sgd.hpp"`.  Alternatively, rather than having subdirectories, we have a flat layout in `ensmallen_bits`, ie. avoid use of further subdirectories.  While the latter option may seem more messy, it has the advantage of all files being visible at the same time.

rcurtin: My only opinion on the format is that it's consistent. :) It sounds like `""` is the right way to go here. How should we document this to ensure that it doesn't happen again? Style guide? Build check that greps for `#include <ensmallen_bits/`?

zoq: That also mean, we I should revert https://github.com/mlpack/ensmallen/pull/234/files/7776391f212b6c20a80c5a7fdb58785625736451#r810741863 and use `"xyz"` instead and adapt `<ensmallen_bits/sgd/sgd.hpp>` accordingly.

I'd like to keep the sub-directory structure, I think it's cleaner, and I don't mind to go one level back and forth. On the GitLab Armadillo repository page, the browser search function is unusable for me; the interface only lists the files in the viewport, so if I search for a file, I have to scroll down to the end, then I can search for a specific file.

include/ensmallen_bits/demon_adam/demon_adam_update.hpp
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_DEMON_ADAM_DEMON_ADAM_UPDATE_HPP
+#define ENSMALLEN_DEMON_ADAM_DEMON_ADAM_UPDATE_HPP
+
+#include <assert.h>
```
rcurtin, line 15: I imagine this is already included by other headers?  (Also I'm not sure it's needed.  Correct me if I'm wrong.)

rcurtin (approved):
Everything looks good to me here. :+1:

It seems like the sanity checks are incorrect or irrelevant---maybe we should filter out those messages?

include/ensmallen_bits/demon_adam/demon_adam_update.hpp
zoq: No `assert` in https://github.com/mlpack/ensmallen/blob/master/include/ensmallen.hpp.

PR #213, opened 2020-08-03
Optimizer Summary Report

zoq:
Example output:

```
Report
--------------------------------------------------------------------------------

Initial Coordinates:
    6.0000  -45.6000    6.2000

Final coordinates:
   5.9975  -6.1651   0.0030

iter          loss          loss change   |gradient|    step size     total time
0             1.22e+03      0             991           0.0003        5.98e-05
333           465           759           6.55          0.0003        0.00106
666           312           154           2.48          0.0003        0.00203
999           209           103           50.1          0.0003        0.00287
1332          140           68.9          0.00248       0.0003        0.00377
1665          93.9          46.2          0.00248       0.0003        0.00465
1998          63            31            0.00248       0.0003        0.00555
2331          42.2          20.8          0.00248       0.0003        0.00635
2664          28.3          13.9          0.0591        0.0003        0.00711
2997          19            9.33          0.00248       0.0003        0.00787
3330          12.7          6.26          0.0178        0.0003        0.00863

--------------------------------------------------------------------------------

Version:
ensmallen:                    2.11.1 (The Poster Session Is Full)
armadillo:                    9.900.1 (Nocturnal Misbehaviour)

Function:
Num functions:                3
Function:
Coordinates rows:             3
Coordinates columns:          1

Loss:
Initial                       1.22e+03
Final                         12.7
Change                        1.21e+03

Optimizer:
Max iterations:               10000
Reached maximum iterations:   false
Batchsize:                    1
Iterations:                   3333
Final step size:              0.0003
Coordinates max norm:         991
Evaluate calls:               10000
Gradient calls:               10000
Time (in seconds):            0.00864
```


doc/callbacks.md
```diff
@@ -208,6 +208,39 @@ optimizer.Optimize(f, coordinates, ProgressBar());
 
 </details>
 
+### Report 
+
+Callback that prints a optimizer report to stdout or a specified output stream.
```
rcurtin, line 213: Should we add some example report output to the documentation here?

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+   * @param optimizer The optimizer used to update the function.
+   * @param function Function to optimize.
+   * @param coordinates Starting point.
+   */
+  template<typename OptimizerType, typename FunctionType, typename MatType>
+  void EndOptimization(OptimizerType& optimizer,
+                       FunctionType& function,
+                       MatType& coordinates)
+  {
+    output << "Report" << std::endl;
```
rcurtin, line 77:
```suggestion
    output << "Optimization Report" << std::endl;
```
I dunno, up to you if you want to change this.  Maybe it's a little more clear? :)

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+   *
+   * @param function The instantiated function that implements NumFunctions().
+   * @param stream The output stream.
+   */
+  template<typename FunctionType>
+  typename std::enable_if<
+      traits::HasNumFunctionsSignature<FunctionType>::value, void>::type
+  PrintNumFunctions(const FunctionType& function, std::stringstream& stream)
+  {
+    PrettyPrintElement(stream, "Num functions:", 30);
```
rcurtin, line 349:
```suggestion
    PrettyPrintElement(stream, "Number of functions:", 30);
```
I dunno, maybe this makes it too verbose, I'm not sure.  "Number of objectives"?  I'm struggling with the best terminology here.

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+   * @param optimizer The instantiated optimizer that implements
+   *     MaxIterations().
+   * @param stream The output stream.
+   */
+  template<typename OptimizerType>
+  typename std::enable_if<
+      traits::HasMaxIterationsSignature<OptimizerType>::value, void>::type
+  PrintMaxIterations(const OptimizerType& optimizer, std::stringstream& stream)
+  {
+    PrettyPrintElement(stream, "Max iterations:", 30);
```
rcurtin, line 372:
```suggestion
    PrettyPrintElement(stream, "Max. iterations:", 30);
```

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+    stream << optimizer.MaxIterations() << std::endl;
+
+    PrettyPrintElement(stream, "Reached maximum iterations:", 30);
+    stream << std::string(optimizer.MaxIterations() == objectives.size() ?
+        "true" : "false") << std::endl;
```
rcurtin, line 377: Up to you, but if the function is separable, do you think we should also print "Max. epochs" too or something like that?

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+   *
+   * @param optimizer The instantiated optimizer that implements BatchSize().
+   * @param stream The output stream.
+   */
+  template<typename OptimizerType>
+  typename std::enable_if<traits::HasBatchSizeSignature<OptimizerType>::value,
+      void>::type
+  PrintBatchSize(const OptimizerType& optimizer, std::stringstream& stream)
+  {
+    PrettyPrintElement(stream, "Batchsize:", 30);
```
rcurtin, line 397:
```suggestion
    PrettyPrintElement(stream, "Batch size:", 30);
```
:)

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+      void>::type
+  PrintBatchSize(const OptimizerType& /* optimizer */,
+                 std::stringstream& /* stream */) { }
+
+  /**
+   * Output formated data.
+   *
+   * @param out Output stream.
+   * @param data The data to print on the given stream.
+   * @param width The width of the the formated output data.
```
rcurtin, line 412:
```suggestion
   * @param width The width of the the formatted output data.
```

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+  }
+
+  template<typename OptimizerType>
+  typename std::enable_if<!traits::HasBatchSizeSignature<OptimizerType>::value,
+      void>::type
+  PrintBatchSize(const OptimizerType& /* optimizer */,
+                 std::stringstream& /* stream */) { }
+
+  /**
+   * Output formated data.
```
rcurtin, line 408:
```suggestion
   * Output formatted data.
```

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+  void PrettyPrintElement(std::ostream& out,
+                          const T& data,
+                          const size_t width = 14)
+  {
+    out << std::left << std::setw(width) << std::setfill(' ')
+        << std::setprecision(3) << data;
+  }
+
+  /**
+   * Output formated data.
```
rcurtin, line 424:
```suggestion
   * Output formatted data.
```

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+   *
+   * @param data The data to print on the given stream.
+   * @param width The width of the the formated output data.
```
rcurtin, line 427:
```suggestion
   * @param width The width of the the formatted output data.
```

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+   * 1 2 ... 5
+   * 6 7 ... 10
+   * ...
+   * 15 16 ... 18
+   *
+   * @param data The data to print on the given stream in a truncated format.
+   * @param size The number of elements per column/row.
+   */
+  template<typename T>
+  void TruncatePrint(const T& data, const size_t size)
```
rcurtin, line 454: This is a nice function!

@conradnicta do you think that you might want to merge this back upstream into Armadillo?  I'm not sure how exactly it would fit, but it might be useful in addition to `.print()` (or maybe as an extra option to `.print()` or something).  Anyway, no opinion from my side, I just thought it might be useful to bring it to your attention. :)

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+
+    if (!stepsizes.empty())
+    {
+      PrettyPrintElement("Final step size:", 30);
+      output << stepsizes.back() << std::endl;
+    }
+
+    if (hasGradient)
+    {
+      PrettyPrintElement("Coordinates max norm:", 30);
```
rcurtin, line 179:
```suggestion
      PrettyPrintElement("Coordinates max. norm:", 30);
```

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+    * @param objective Objective value of the current point.
+    */
+  template<typename OptimizerType, typename FunctionType, typename MatType>
+  void EvaluateConstraint(OptimizerType& optimizer,
+                          FunctionType& function,
+                          const MatType& coordinates,
+                          const size_t /* constraint */,
+                          const double objective)
+  {
+    Evaluate(optimizer, function, coordinates, objective);
```
rcurtin, line 295: Are you sure this is what we want here?  It looks like this will set `this->objective` to the objective given to `EvaluateConstraint()`, but maybe what we want to be doing instead is adding the `objective` input (which is the objective of the constraint) to the overall objective?  i.e.:

```
  objective += objectiveIn;
```

Correct me if I overlooked something there.

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -0,0 +1,558 @@
...
+   * @param objective Objective value of the current point.
+   */
+  template<typename OptimizerType, typename FunctionType, typename MatType>
+  void EndEpoch(OptimizerType& optimizer,
+                FunctionType& /* function */,
+                const MatType& /* coordinates */,
+                const size_t /* epoch */,
+                const double objective)
+  {
+    // In case StepTake() has been called first we clear the existing data.
```
rcurtin, line 216:
```suggestion
    // In case StepTaken() has been called first we clear the existing data.
```

zoq:
> This looks awesome! All of my comments are super minor.
> 
> I wondered briefly if it would be possible to print "interactively" the progress; i.e., once we are 10% done, print the results for 10% of the way through the optimization immediately, instead of waiting until everything is done. However I am not sure if this is reasonably possible given how the callback is architected, so I think it's fine as-is unless maybe you see an easy way to accomplish that. 👍

I guess there is no standard way that works across all optimizers, for some we could use `StepTaken` method and introduce another parameter that defines after how many iterations the report should be called. For other optimizers the `Gradient/Evalaute` call could be used.

rcurtin:
Yeah---maybe it is not readily possible.  In any case, we could always leave it for later.  This summary report is super useful as-is, I think. :+1:

doc/callbacks.md
rcurtin: Nice, the output looks good.  Thanks for adding that! :+1:

include/ensmallen_bits/callbacks/report.hpp
rcurtin: Looks like this is handled now, so I'll resolve the conversation. :+1:

rcurtin:
Everything looks just about ready here---but there are still a couple of open conversations.  I think they should all be pretty easy to address. :+1:

include/ensmallen_bits/callbacks/report.hpp
zoq: I called it `Number of epochs` - https://github.com/mlpack/ensmallen/pull/213/files#diff-5aa5b7e1523591f922934321e984ab31e0171e931ceee2ca499d993daf1e0a3cR173

But I can rename it.

zoq:
> Everything looks just about ready here---but there are still a couple of open conversations. I think they should all be pretty easy to address. 👍

For some reason I missed some of the comments, thanks for the reminder.

zoq:
I think I got everything now.

include/ensmallen_bits/callbacks/report.hpp
rcurtin: Ah, looks good to me!  Sorry if I missed that in the original commit. :+1:

rcurtin (approved):
Just one more little spelling fix. :)  https://github.com/mlpack/ensmallen/pull/213#discussion_r466678548

Anyway, otherwise this looks great to me!  This is nice support to add. :+1:

zoq:
That is so strange, I thought I hit that button.

include/ensmallen_bits/callbacks/report.hpp
conradsnicta: @rcurtin Sorry, only saw this now.  The @ was missing an 's'.
A version of .print() which prints an abridged matrix would indeed be useful -- similar to Python.  I'll add that on the TODO list for Armadillo 10.2

rcurtin: Oh, you're right, I missed a key.  Sorry about that!

conradsnicta: @rcurtin @zoq I've added `.brief_print()` to Armadillo 10.2: http://arma.sourceforge.net/docs.html#brief_print

zoq: This is nice.

Issue #142, opened 2019-11-20
Tests fail for ensmallen-2.10.4


yurivict:
```
    Start 1: ensmallen_tests
1/1 Test #1: ensmallen_tests ..................***Failed  363.56 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) = 363.57 sec

The following tests FAILED:
	  1 - ensmallen_tests (Failed)
Errors while running CTest
```

OS: FreeBSD

zoq:
Hello @yurivict I couldn't reproduce the issue on FreeBSD 11.2, can you share some more information about your setup, OS, clang, armadillo version. Also does `./ensmallen_tests -s -r console` show any intermediate results?

yurivict:
I added the port for ensmallen yesterday, so if you would just type ```cd /usr/ports/math/ensmallen &&  make test``` it should reproduce the problem.

rcurtin:
If you can run with `CTEST_OUTPUT_ON_FAILURE=1 ctest` that would be really helpful too.  ensmallen has some random tests.  We try to keep the random failure probability really low, but we're not always successful, so if you can show us which test failed we can figure out if it's random or an actual problem.

yurivict:
It passes now when I added ```CTEST_OUTPUT_ON_FAILURE=1```

yurivict:
I added this patch to the FreeBSD port to work around this issue: 
```
--- src/libtfhe/fft_processors/nayuki/fft_processor_nayuki.cpp.orig     2019-10-11 03:07:51 UTC
+++ src/libtfhe/fft_processors/nayuki/fft_processor_nayuki.cpp
@@ -12,7 +12,7 @@ FFT_Processor_nayuki::FFT_Processor_nayuki(const int32
     tables_reverse = fft_init_reverse(_2N);
     omegaxminus1 = (cplx*) malloc(sizeof(cplx) * _2N);
     for (int32_t x=0; x<_2N; x++) {

+       omegaxminus1[x]=std::complex<double>(cos(x*M_PI/N)-1., sin(x*M_PI/N));
        //exp(i.x.pi/N)-1
     }
 }
```

zoq:
Hm, is this the correct patch? This code isn't part of ensmallen, maybe I missed something?

barak:
Even with the "ten billion" issue fixed, I'm getting build failures on some 32-bit architectures. This is Debian version `2.10.4-2` (the `-2` includes the ten billion patch).

[...]

PR #217, opened 2020-08-12
Set AugLagrangian tolerance based on ElemType

rcurtin:
Finally, after far too long, I took some time to look into #142.  I was able to replicate the problem with the `Johnson844LovaszThetaFMatSDP` test.

The augmented Lagrangian optimizer (`AugLagrangian`) keeps a penalty parameter `sigma` internally.  This is stored as a `double`---but `Johnson844LovaszThetaFMatSDP` is a test that specifically uses `arma::fmat` as an objective matrix type (so, e.g., `float` is the element type).

`AugLagrangian` was also hard-coded to terminate on an absolute objective difference of `1e-10`---but for this particular test, the objective function converges to `-14`... and due to the precision afforded by `1e-14`, it's not actually possible to have two different `float`s very near `-14` that are less than `1e-10` apart!  So, I modified the tolerance to be based on the value of `numeric_limits<ElemType>::epsilon()`.

Next, I also set a limit on how large `sigma` can grow.  The optimization will terminate when `sigma` is greater than half the size that can be represented by `ElemType`.  This issue (when paired with the other one) is what caused these errors:

```
Parameter 4 to routine SLASCL was incorrect
Parameter 5 to routine SLASCL was incorrect
```

Those were caused by a too-large `sigma`.

Anyway, to the best of my belief this fixes the issue.  The only thing I am not sure of is why this failure only seemed to occur on an i386 system, instead of also on an x86_64 system.  But, I don't know that I'll dig deeper this time---if it works, it works. :)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,7 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
+ * Adjust tolerance for AugLagrangian convergence based on element type
+   ([#217](https://github.com/mlpack/ensmallen/pull/217)).
 
 ### ensmallen 2.14.0: "No Direction Home"
 ###### 2020-08-10
--- a/include/ensmallen_bits/aug_lagrangian/aug_lagrangian_impl.hpp
+++ b/include/ensmallen_bits/aug_lagrangian/aug_lagrangian_impl.hpp
@@ -108,6 +108,10 @@ AugLagrangian::Optimize(
   // Track the last objective to compare for convergence.
   ElemType lastObjective = function.Evaluate(coordinates);
 
+  // Convergence tolerance---depends on the epsilon of the type we are using for
+  // optimization.
+  ElemType tolerance = 1e3 * std::numeric_limits<ElemType>::epsilon();
+
   // Then, calculate the current penalty.
   ElemType penalty = 0;
   for (size_t i = 0; i < function.NumConstraints(); i++)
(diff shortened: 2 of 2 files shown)


conradsnicta:
> The only thing I am not sure of is why this failure only seemed to occur on an i386 system, instead of also on an x86_64 system

Speculation on a possible cause.  It's been a while since I looked into this, so the exact details are a little vague.  Going from memory, the built-in FPU on the i386 (or on x86-64 ?) may internally use extra precision for calculations, with something resembling behind-the-scenes type promotion. A bunch of calculations can be done in extra precision before the result needs to be written back to memory (and hence the result is forced back into "normal" precision).  

For double precision, the values are effectively represented in 80 bits instead of 64 bits. Maybe something similar is happening with single precision.

gcc has the `-ffloat-store` option, which states:
> -ffloat-store
           Do not store floating-point variables in registers, and inhibit other options that might change whether a floating-point value is taken from a register or memory. This option prevents undesirable excess precision on machines such as the 68000 where the floating registers (of the 68881) keep more precision than a "double" is supposed to have.  Similarly for the x86 architecture. For most programs, the excess precision does only good, but a few programs rely on the precise definition of IEEE floating point.  Use -ffloat-store for such programs, after modifying them to store all pertinent intermediate computations into variables.

An old-ish blog post which attempts to explain the mess:
https://randomascii.wordpress.com/2012/03/21/intermediate-floating-point-precision/

rcurtin:
Thanks @conradsnicta, those are some nice resources.  I think maybe that is what was going on here. :+1:

PR #219, opened 2020-08-18
Release version 2.14.1: "No Direction Home"

rcurtin:
This automatically-generated pull request adds the commits necessary to make the 2.14.1 release.

Once the PR is merged, mlpack-bot will tag the release as HEAD~1 (so that it doesn't include the new HISTORY block) and publish it.

Or, well, hopefully that will happen someday.

When you merge this PR, be sure to merge it using a *rebase*.

### Changelog

 * Fix release script (remove hardcoded information, trim leading whitespaces introduced by `wc -l` in MacOS) ([#216](https://github.com/mlpack/ensmallen/pull/216)).

 * Adjust tolerance for AugLagrangian convergence based on element type ([#217](https://github.com/mlpack/ensmallen/pull/217)).

PR #220, opened 2020-08-18
Fix quotation bug in release script.

rcurtin:
This is a simple change to correctly extract the changelog in the release script.  Previously, it would search for, e.g.,

```
### ensmallen 2.14.1: No Direction Home
```

but it needs to search for

```
### ensmallen 2.14.1: "No Direction Home"
```

You can see an example of the existing script failing in #218 and then the modified script succeeding in #219.


zoq (approved):
Makes sense, looks good to me.

Issue #222, opened 2020-08-22
HISTORY.md version update not included in the tag build


coatless:
Hi all,

I'm working on automating a PR based on an upstream change in `ensmallen`'s codebase (c.f. coatless/rcppensmallen#30). One part that I noticed is the `.tar.gz` made available under the release tag does not include the appropriate `HISTORY.md` version update.

For instance:

https://raw.githubusercontent.com/mlpack/ensmallen/2.14.1/HISTORY.md

Gives:

```md
### ensmallen ?.??.?: "???"
###### ????-??-??
 * Fix release script (remove hardcoded information, trim leading whitespaces
   introduced by `wc -l` in MacOS)
    ([#216](https://github.com/mlpack/ensmallen/pull/216),
     [#220](https://github.com/mlpack/ensmallen/pull/220)).

 * Adjust tolerance for AugLagrangian convergence based on element type
   ([#217](https://github.com/mlpack/ensmallen/pull/217)).
```

Only in a later commit is the information passed on:

https://raw.githubusercontent.com/mlpack/ensmallen/9d3fc459ce34eefd969dc0843d55c619b13e4b6e/HISTORY.md

```md
### ensmallen 2.14.1: "No Direction Home"
###### 2020-08-19
```

It's a minor inconvenience to substitute this in as unreleased version is kept; however, it would be nice for the `.tar.gz` associated with the release tag to have an up-to-date `HISTORY.md` file.

coatless:
@rcurtin I didn't realize but the version information propagation into `ens_version.hpp` is also broke. The tagged build shows 2.14.0 -- not v2.14.1.

Tagged version: https://github.com/mlpack/ensmallen/blob/2.14.1/include/ensmallen_bits/ens_version.hpp#L15-L19

```cpp
#define ENS_VERSION_MAJOR 2
// The minor version is two digits so regular numerical comparisons of versions
// work right.  The first minor version of a release is always 10.
#define ENS_VERSION_MINOR 14
#define ENS_VERSION_PATCH 0
```

Later version: 

https://github.com/mlpack/ensmallen/blob/9d3fc459ce34eefd969dc0843d55c619b13e4b6e/include/ensmallen_bits/ens_version.hpp#L15-L19

rcurtin:
[...]

PR #226, opened 2020-09-06
Incorporate release date with ensmallen version

coatless:
From #222, we discussed adding date information to the header file for when ensmallen was released/updated. This PR includes the version update within `ens_version.hpp` and updates the release script to automatically set these fields.

### DIFF
--- a/include/ensmallen_bits/ens_version.hpp
+++ b/include/ensmallen_bits/ens_version.hpp
@@ -44,7 +44,7 @@ struct version
 
     std::stringstream ss;
     ss << version::major << '.' << version::minor << '.' << version::patch
-       << " (" << nickname << ')\n' 
+       << " (" << nickname << ")\n" 
        << "Released on " << year << '-' << month << '-' << day;
 
     return ss.str();
--- a/include/ensmallen_bits/ens_version.hpp
+++ b/include/ensmallen_bits/ens_version.hpp
@@ -59,5 +59,4 @@ struct version
   }
 };
 
-
 } // namespace ens
--- a/include/ensmallen_bits/ens_version.hpp
+++ b/include/ensmallen_bits/ens_version.hpp
@@ -22,7 +22,7 @@
 // name will typically be a seemingly arbitrary set of words that does not
 // contain the capitalized string "RC".
 #define ENS_VERSION_NAME  "No Direction Home"
-// Incorporate the date the version was released
+// Incorporate the date the version was released.
 #define ENS_VERSION_YEAR  "2020"
 #define ENS_VERSION_MONTH "09"
 #define ENS_VERSION_DAY   "05"
--- a/include/ensmallen_bits/ens_version.hpp
+++ b/include/ensmallen_bits/ens_version.hpp
@@ -41,19 +41,15 @@ struct version
 
     std::stringstream ss;
     ss << version::major << '.' << version::minor << '.' << version::patch
-       << " (" << nickname << ")";
+       << " (" << nickname << ')';
 
     return ss.str();
   }
     
   static inline std::string date()
   {
-    const char* year = ENS_VERSION_YEAR;
-    const char* month = ENS_VERSION_MONTH;
-    const char* day = ENS_VERSION_DAY;
-
     std::stringstream ss;
-    ss << year << '-' << month << '-' << day;
+    ss << ENS_VERSION_YEAR << '-' << ENS_VERSION_MONTH << '-' << ENS_VERSION_DAY;
 
     return ss.str();
   }
--- a/include/ensmallen_bits/ens_version.hpp
+++ b/include/ensmallen_bits/ens_version.hpp
@@ -21,7 +21,11 @@
 // (i.e. the version name will be "RC1", "RC2", etc.).  Otherwise the version
 // name will typically be a seemingly arbitrary set of words that does not
 // contain the capitalized string "RC".
-#define ENS_VERSION_NAME "No Direction Home"
+#define ENS_VERSION_NAME  "No Direction Home"
+// Incorporate the date the version was released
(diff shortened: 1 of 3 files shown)


conradsnicta:
What's the use case for the date?  Having a date appears redundant.  The version number already contains all the necessary information to determine the functionality (ie. API state) and bug fixes present within a given release.  I'm not convinced by the reasoning provided in https://github.com/mlpack/ensmallen/issues/222.


A further problem with adding the date is that it adds more things to keep track of (ie. adds to the maintenance burden), while providing no compelling value.

 If somebody really needs to know the date of a release, they can always look at the [tags](https://github.com/mlpack/ensmallen/tags) or perhaps the date stamps within a release.

coatless:
@conradsnicta thanks for the question. The reason for the date being embedded in `ens_version.hpp` was to simplify extraction. Prior to that, there was a hiccup during release when the tagged version lacked both a version number/date, which was problematic for generating an automatic upstream PR refresh with [`RcppEnsmallen`](https://github.com/coatless/rcppensmallen)'s [`upstream-refresh`](https://github.com/coatless/rcppensmallen/blob/master/.github/workflows/upstream-refresh.yaml) GitHub Action.

Now onto the other point, I do agree except for the fact that `ensmallen` has a real nice release utility @rcurtin worked on: [`scripts/ensmallen-release.sh`](https://github.com/mlpack/ensmallen/blob/60a99f780ba24fb0403579534bc53db9e09f54c9/scripts/ensmallen-release.sh). The version and date information is automatically set and a release candidate PR is created.

include/ensmallen_bits/ens_version.hpp
```diff
@@ -34,10 +38,14 @@ struct version
...
     const char* nickname = ENS_VERSION_NAME;
+    const char* year = ENS_VERSION_YEAR;
+    const char* month = ENS_VERSION_MONTH;
+    const char* day = ENS_VERSION_DAY;
 
     std::stringstream ss;
     ss << version::major << '.' << version::minor << '.' << version::patch
-       << " (" << nickname << ')';
+       << " (" << nickname << ")\n" 
+       << "Released on " << year << '-' << month << '-' << day;
```
rcurtin, line 48: I feel like someone would expect `ens::version::as_string()` to return something very simple without a newline; if we do go with this idea, maybe we should make an additional function `date()` that returns `year << '-' << month << '-' << day`?

rcurtin (commented):
Given the automatic release script, I agree that there's not much maintenance burden for keeping this up-to-date.  Another way to get the date would be something like this:

```
grep '^######' HISTORY.md  | grep -v '????' | head -1 | sed 's/^###### //'
```

and then you have the release date in YYYY-MM-DD form.

Anyway, I'm fairly ambivalent about what gets chosen here---if it really does make your life easier to add it here, great, I don't have a problem with merging it, but if you have another way to get the same information without adding extra complexity (even though we can hide that in the automatic release scripts), I'd probably lean that way.  Let me know what you think. :+1:  (And sorry for the slow response!)

rcurtin:
Hey @coatless, just trying to see this one through---what do you think?  Either approach is fine with me. :+1:

coatless:
@rcurtin sorry for missing this. I've been swamped the last few weeks. I'll aim to add a second function.

coatless:
@rcurtin split the function. Looks like everything is passing.

include/ensmallen_bits/ens_version.hpp
```diff
@@ -37,10 +41,23 @@ struct version
...
+    const char* day = ENS_VERSION_DAY;
+
+    std::stringstream ss;
+    ss << year << '-' << month << '-' << day;
 
     return ss.str();
   }
 };
 
+
```
rcurtin, line 62:
```suggestion
```
No need for an extra line. :)

rcurtin (approved):
Awesome, looks good to me.  Thanks @coatless! :+1:  And no worries about being swamped; I know how busy things can get sometimes...

include/ensmallen_bits/ens_version.hpp
```diff
@@ -21,7 +21,11 @@
 // (i.e. the version name will be "RC1", "RC2", etc.).  Otherwise the version
 // name will typically be a seemingly arbitrary set of words that does not
 // contain the capitalized string "RC".
-#define ENS_VERSION_NAME "No Direction Home"
+#define ENS_VERSION_NAME  "No Direction Home"
+// Incorporate the date the version was released
```
zoq, line 25:
```suggestion
// Incorporate the date the version was released.
```

Pedantic style issue, just to be consistent.

include/ensmallen_bits/ens_version.hpp
```diff
@@ -37,7 +41,19 @@ struct version
...
+  }
+    
+  static inline std::string date()
+  {
+    const char* year = ENS_VERSION_YEAR;
+    const char* month = ENS_VERSION_MONTH;
+    const char* day = ENS_VERSION_DAY;
+
+    std::stringstream ss;
+    ss << year << '-' << month << '-' << day;
```
zoq, line 56: What is the reason to not pass the date into the stream directly:

`ss << ENS_VERSION_YEAR << '-' << ENS_VERSION_MONTH << '-' << ENS_VERSION_DAY;`

coatless: @zoq No reason not to. I opted to mirror the styling of `nickname` if memory serves correctly.

zoq (approved):
This looks great to me, thanks for putting this together.

coatless:
@rcurtin / @zoq if possible -- when this is merged in -- could you use squash and, then, merge? No need for keeping all of these commits in the repo history.

rcurtin:
Sure!  Thanks again for adding this. :+1:

Issue #225, opened 2020-09-03
2.14.1: Tests fail: REQUIRE( coordinates(1) == Approx(0.0f).margin(0.01) )


yurivict:
```
]$ ./work/.build/ensmallen_tests
ensmallen version: 2.14.0 (No Direction Home)
armadillo version: 9.900.2 (Nocturnal Misbehaviour)

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ensmallen_tests is a Catch v2.4.1 host application.
Run with -? for options

-------------------------------------------------------------------------------
SimpleAdaDeltaTestFunctionFMat
-------------------------------------------------------------------------------
/usr/ports/math/ensmallen/work/ensmallen-2.14.1/tests/ada_delta_test.cpp:66
...............................................................................

/usr/ports/math/ensmallen/work/ensmallen-2.14.1/tests/ada_delta_test.cpp:75: FAILED:
  REQUIRE( coordinates(1) == Approx(0.0f).margin(0.01) )
with expansion:
  -0.13114f == Approx( 0.0 )
```

OS: FreeBSD 12.1
clang-9

UtR491:
I think you should take a look at #182.

zoq:
@yurivict any chance you can check if https://github.com/mlpack/ensmallen/issues/182 works in some cases, if that's the case I'm going to bump the number of iterations.

rcurtin:
@zoq I think bumping the iterations is the right thing to do here.  Let me know what you think; if you want, I can put together a PR, up to you. :+1:

rcurtin:
Hey @yurivict, I believe this is fixed in #228.  So hopefully in the next ensmallen release you will not see this problem.  Thanks for the report! :+1:

PR #228, opened 2020-10-04
Fix SimpleSGDTestFunction arma::fmat test failures

### LINKED ISSUE #182: SimpleAdaDeltaTestFunctionFMat Failing
#### Steps to reproduce

Just Run The Ensmallen Test Suite

#### Expected behavior

The Test Should Pass.

#### Actual behavior

Screenshot 2020-03-18 at 7 52 29 PM

I would like to work on this issue, if everyone is experiencing this.

rcurtin:
This PR adapts the strategies used for the two tests (AdaDelta and AdaGrad) that use `SimpleSGDTestFunction` with an `arma::fmat`.  Basically I increased the step size and this seemed to help, but not enough, so I then made the test run a maximum of three times to make sure it converges at least once.  With these changes, now locally I saw no failures of either test with >1000 runs.

This should fix #182 and #225. :+1:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,6 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
+ * Make a few tests more robust.
 
 ### ensmallen 2.14.2: "No Direction Home"
 ###### 2020-08-31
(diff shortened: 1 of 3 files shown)


zoq (approved):
Looks great to me.

birm (approved):

PR #231, opened 2020-10-18
Nelder-Mead simplex algorithm with adaptive parameters

zoq:
Nelder-Mead simplex algorithm with adaptive parameters as proposed in "Implementing the Nelder-Mead simplex algorithmwith adaptive parameters" by F Gao and L. Han. The implementation outperforms the standard Nelder-Mead method   (for high dimensional problems).

### DIFF
--- a/include/ensmallen_bits/nelder_mead/nelder_mead.hpp
+++ b/include/ensmallen_bits/nelder_mead/nelder_mead.hpp
@@ -1,0 +1,234 @@
+/**
+ * @file NelderMead.hpp
+ * @author Marcus Edel
+ *
+ * Definition of an Nelder-Mead with adaptive parameters proposed by F. Gao and
+ * L. Han in "Implementing the Nelder-Mead simplex algorithm with adaptive
+ * parameters".
+ *
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_NELDER_MEAD_HPP
+#define ENSMALLEN_NELDER_MEAD_HPP
+
+#include "affine_simplexer.hpp"
+
+namespace ens {
+
+/**
+ * Nelder-Mead is a direct search method. It keeps track of the function value
+ * at a number of points in the search space and iteratively generates a
+ * sequence of simplices to approximate an optimal point.
+ *
+ * For more information, please refer to:
+ *
+ * @code
+ * @article{Nelder1965ASM,
+ *   author  = {J. Nelder and R. Mead},
+ *   title   = {A Simplex Method for Function Minimization},
+ *   year    = {1965},
+ *   journal = {Comput. J.},
+ *   volume  = {7},
+ *   pages   = {308-313}
+ * }
+ * @endcode
+ *
+ * @code
+ * @article{Fuchang2012,
+ *   author    = {Gao, Fuchang and Han, Lixing},
+ *   title     = {Implementing the Nelder-Mead Simplex Algorithm with Adaptive
+ *                Parameters},
+ *   year      = {2012},
+ *   publisher = {Kluwer Academic Publishers},
+ *   volume    = {51},
+ *   number    = {1},
+ *   journal   = {Comput. Optim. Appl.},
+ *   month     = jan,
+ *   pages     = {259–277}
+ * }
+ * @endcode
+ *
+ * NelderMead can optimize arbitrary functions.  For more details, see the
+ * documentation on function types included with this distribution or on the
+ * ensmallen website.
+ *
+ * @tparam SimplexerType Simplex policy used by Nelder Mead to construct the
+ * initial simplex.
+ */
+template<typename SimplexerType = AffineSimplexer>
+class NelderMeadType
+{
+ public:
+  /**
+   * Construct the NelderMead optimizer with the default adaptive parameters
+   * scheme as used in "Implementing the Nelder-Mead simplex algorithm with
+   * adaptive parameters" by F. Gao and L. Han (2010). These are based on the
+   * dimensionality of the problem, and are given by
+   *
+   * alpha = 1
+   * beta = 1 + 2 / n
+   * gamma = 0.75 - 0.5 / n
+   * delta = 1 - 1 / n
+   *
+   * where n is the dimensionality of the problem.
+   *
+   * The defaults here are not necessarily good for the given problem, so it is
+   * suggested that the values used be tailored to the task at hand.  The
+   * maximum number of iterations refers to the maximum number of points that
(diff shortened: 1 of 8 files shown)


include/ensmallen_bits/nelder_mead/nelder_mead.hpp
```diff
@@ -0,0 +1,234 @@
+/**
+ * @file NelderMead.hpp
```
zoq, line 2:
```suggestion
 * @file nelder_mead.hpp
```

include/ensmallen_bits/nelder_mead/nelder_mead_impl.hpp
```diff
@@ -0,0 +1,241 @@
...
+    beta = 1.0 + 2.0 / (double) dim;
+    gamma = 0.75 - 0.5 / (double) dim;
+    delta = 1.0 - 1.0 / (double) dim;
+  }
+
+  // Construct the initial simplex.
+  MatType simplex;
+  simplexer.Simplex(simplex, iterate, function);
+
+  arma::rowvec fx(dim + 1);
```
rcurtin, line 101: It could be more efficient to use a max-heap here instead of a vector.

include/ensmallen_bits/nelder_mead/nelder_mead.hpp
```diff
@@ -0,0 +1,234 @@
...
+   * here are not necessarily good for the given problem, so it is suggested
+   * that the values used be tailored to the task at hand.  The maximum number
+   * of iterations refers to the maximum number of points that are processed
+   * (i.e., one iteration equals one point; one iteration does not equal one
+   * pass over the dataset).
+   *
+   * @param alpha The reflection parameter.
+   * @param beta The expansion parameter.
+   * @param gamma The contraction parameter.
+   * @param delta The shrink step parameter.
```
rcurtin, line 103: It might be pointing out that the ANMS paper uses different notation here than the Wikipedia description of Nelder-Mead; on Wikipedia, gamma is the expansion parameter, and in the ANMS paper, beta is...  I don't think we need to change the implementation, but we should probably point that out. :+1:

include/ensmallen_bits/nelder_mead/nelder_mead.hpp
```diff
@@ -0,0 +1,234 @@
...
+  NelderMeadType(const size_t maxIterations = 100000,
+                 const double tolerance = 1e-15,
+                 const SimplexerType& simplexer = SimplexerType());
+  /**
+   * Construct the NelderMead optimizer with the given parameters. The defaults
+   * here are not necessarily good for the given problem, so it is suggested
+   * that the values used be tailored to the task at hand.  The maximum number
+   * of iterations refers to the maximum number of points that are processed
+   * (i.e., one iteration equals one point; one iteration does not equal one
+   * pass over the dataset).
```
rcurtin, line 98: Does this comment apply here to an arbitrary function?

include/ensmallen_bits/nelder_mead/nelder_mead_impl.hpp
```diff
@@ -0,0 +1,241 @@
...
+      // Compute expansion.
+      const MatType xexp = (1.0 + alpha * beta) * M - alpha * beta *
+          simplex.col(order(dim));
+      const ElemType fexp = function.Evaluate(xexp);
+      Callback::Evaluate(*this, function, xexp, fexp, callbacks...);
+
+      // Shift values by one to the right and swap the last with the first.
+      const size_t fLowest = order(dim);
+      for (size_t d = dim; d > 0; --d)
+        order(d) = order(d - 1);
```
rcurtin, line 134: Here too you could use `subvec()` to copy all at once, or, if you use a max-heap this step may not be needed.  (I believe `std::priority_queue` is a max-heap.)

include/ensmallen_bits/nelder_mead/nelder_mead_impl.hpp
```diff
@@ -0,0 +1,241 @@
...
+        }
+        else
+        {
+          shrink = true;
+        }
+      }
+      else
+      {
+        // Inside contraction.
+        const MatType xic = (1.0 - gamma) * M + gamma *
```
rcurtin, line 179: Should this have `alpha` multiplied too?  I did the algebra in my head and I think it should, but I didn't double check and I have a headache right now... :smile:

include/ensmallen_bits/nelder_mead/nelder_mead_impl.hpp
```diff
@@ -0,0 +1,241 @@
...
+        fx(order(d)) = function.Evaluate(simplex.col(order(d)));
+
+        Callback::Evaluate(*this, function, simplex.col(order(d)),
+            fx(order(d)), callbacks...);
+      }
+
+      // Sorting can be slow, however there is overwhelming evidence that shrink
+      // transformations almost never happen in practice. See "Efficient
+      // Implementation of the Nelder–Mead Search Algorithm" by S. Singer and S.
+      // Singer.
```
rcurtin, line 212: I thought for sure this had to be a typo, but no, there are actually *two* S. Singers, and they collaborated on this. :smile:

include/ensmallen_bits/nelder_mead/nelder_mead_impl.hpp
```diff
@@ -0,0 +1,241 @@
...
+        }
+      }
+    }
+
+    if (shrink)
+    {
+      for (size_t d = 1; d < dim; ++d)
+      {
+        simplex.col(order(d)) = simplex.col(order(0)) + delta *
+            (simplex.col(order(d)) - simplex.col(order(0)));
```
rcurtin, line 202: It could be possible to express this as two operations (first on columns with index less than `order(0)`, and second on columns with index greater than `order(0)`), using `repmat()`, but it's up to you if you want to rewrite it like that.  I believe it would be more efficient, but I don't know how much of a bottleneck it would be.  The `Evaluate()` would still need to be called over all new points in a loop though.

include/ensmallen_bits/nelder_mead/nelder_mead_impl.hpp
```diff
@@ -0,0 +1,241 @@
...
+            fx(order(d)), callbacks...);
+      }
+
+      // Sorting can be slow, however there is overwhelming evidence that shrink
+      // transformations almost never happen in practice. See "Efficient
+      // Implementation of the Nelder–Mead Search Algorithm" by S. Singer and S.
+      // Singer.
+      order = arma::sort_index(fx);
+      fx = fx.cols(order);
+      simplex = simplex.cols(order);
```
rcurtin, line 215: Is it necessary to reorder `simplex` at the end of the loop?  (And `fx` too?)  Inside the main loop of the optimization, those two members are always accessed with `order()`, so I think it is fine to leave them in the order they're in.

rcurtin (commented):
Awesome, so glad to see more arbitrary optimizers implemented!  Most of my comments are minor, but I think I found a bug and perhaps a couple efficiency improvements. :+1:

PR #232, opened 2020-10-20
Yogi - Adaptive Methods for Nonconvex Optimization 

zoq:
Implementation of Yogi - "Adaptive Methods for Nonconvex Optimization" M. Zaheer et al.

### DIFF
--- a/doc/optimizers.md
+++ b/doc/optimizers.md
@@ -2886,3 +2886,56 @@ optimizer.Optimize(f, coordinates);
  * [SGD](#standard-sgd)
  * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
  * [Differentiable separable functions](#differentiable-separable-functions)
+
+## Yogi
+
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+Yogi is an optimization algorithm based on Adam with more fine grained
+effective learning rate control, and has similar theoretical guarantees on
+convergence as Adam.
+
+#### Constructors
+
+ * `Yogi()`
+ * `Yogi(`_`stepSize, batchSize`_`)`
+ * `Yogi(`_`stepSize, batchSize, beta1, beta2, eps, maxIterations`_`)`
+ * `Yogi(`_`stepSize, batchSize, beta1, beta2, eps, maxIterations, tolerance, shuffle, resetPolicy, exactObjective`_`)`
+
+#### Attributes
+
+| **type** | **name** | **description** | **default** |
+|----------|----------|-----------------|-------------|
+| `double` | **`stepSize`** | Step size for each iteration. | `0.001` |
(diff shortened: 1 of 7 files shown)


doc/optimizers.md
```diff
@@ -2886,3 +2886,56 @@ optimizer.Optimize(f, coordinates);
...
  * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
  * [Differentiable separable functions](#differentiable-separable-functions)
+
+## Yogi
+
+*An optimizer for [differentiable separable functions](#differentiable-separable-functions).*
+
+Yogi is an optimization algorithm based on Adam with more fine grained
+effective learning rate control, and has similar theoretical guarantees on
+convergence as Adam.
```
rcurtin, line 2896: Maybe it is worth mentioning that Yogi uses additive updates instead of multiplicative updates for the `v` parameter?

doc/optimizers.md
```diff
@@ -2886,3 +2886,56 @@ optimizer.Optimize(f, coordinates);
...
+RosenbrockFunction f;
+arma::mat coordinates = f.GetInitialPoint();
+
+Yogi optimizer(0.001, 32, 0.9, 0.999, 1e-8, 100000, 1e-5, true);
+optimizer.Optimize(f, coordinates);
+```
+
+#### See also:
+
+ * [Quasi-Hyperbolic Momentum and Adam For Deep Learning](https://arxiv.org/pdf/1810.06801.pdf)
```
rcurtin, line 2936: Maybe this reference is not relevant here? I may have missed something.

rcurtin (approved):
Interesting paper, but I was surprised that the first 3 pages focus entirely on convergence issues when changing the batch size---and then Yogi itself does not address that issue, and they seem to use constant batch size in all their experiments. I guess that was so that a reviewer would not be able to claim that Yogi, like Adam, can be shown to not converge in some settings...

Anyway the implementation looks just fine to me. :+1: Sorry the review took so long 😄

include/ensmallen_bits/yogi/yogi_update.hpp
```diff
@@ -0,0 +1,158 @@
...
+                const double stepSize,
+                const GradType& gradient)
+    {
+      // Increment the iteration counter variable.
+      ++parent.iteration;
+
+      m *= parent.beta1;
+      m += (1 - parent.beta1) * gradient;
+      
+      const MatType gSquared = gradient % gradient;
```
conradsnicta, line 124: may want to use `arma::square(gradient)` here instead of `gradient % gradient`

zoq: Totally, thanks for the quick review.

rcurtin:
Everything looks good here to me---do you want to go ahead and merge or was there anything else that needed to be handled first?

zoq:
No this should be fine, let me hit the merge button.

PR #234, opened 2020-10-25
AdaSqrt - Second-order Information in First-order Optimization Methods

zoq:
Implementation of AdaSqrt - "Second-order Information in First-order Optimization Methods", Yuzheng Hu, Licong Lin, Shange Tang.


zoq:

zoq:
Let's fix the static analysis build.

zoq:

zoq:
Alright, it's working again.

doc/optimizers.md
```diff
@@ -168,6 +168,64 @@ optimizer.Optimize(f, coordinates);
 
 </details>
 
+#### See also:
+
+ * [Adaptive Subgradient Methods for Online Learning and Stochastic Optimization](http://www.jmlr.org/papers/volume12/duchi11a/duchi11a.pdf)
+ * [AdaGrad in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent#AdaGrad)
+ * [AdaDelta](#adadelta)
+ * [Differentiable separable functions](#differentiable-separable-functions)
```
rcurtin, line 176: Nice catch, thanks for adding this 'see also' section.

include/ensmallen_bits/ada_sqrt/ada_sqrt_update.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+      iterate -= stepSize * std::sqrt(parent.iteration) * gradient /
+          (squaredGradient + parent.epsilon);
+    }
+
+   private:
+    // Instantiated parent class.
+    AdaSqrtUpdate& parent;
+    // The squared gradient matrix.
+    GradType squaredGradient;
```
rcurtin, line 110: Up to you but it may be clearer to name this `sumSquaredGradients`.  I got a little confused and had to double-check the paper to understand.

tests/CMakeLists.txt
```diff
@@ -59,4 +60,4 @@ add_custom_command(TARGET ensmallen_tests
 
 enable_testing()
 add_test(NAME ensmallen_tests COMMAND ensmallen_tests
-WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
\ No newline at end of file
+WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
```
rcurtin, line 63: 🎉

tests/ada_grad_test.cpp
```diff
@@ -1,8 +1,6 @@
 /**
- * @file ada_grad_test.cpp
- * @author Abhinav Moudgil
+ * @file ada_sqrt_test.cpp
```
rcurtin, line 2: Oops, did you mean to commit these changes to `ada_grad_test.cpp` or just to the new test file?

rcurtin (commented):
This paper was a bit harder to read (I guess I read a preprint).  I'm not sure I followed the idea behind the approach well, but the PR looks like it exactly implements the algorithm of the paper.  I only have a couple minor comments.  Sorry this review took so long to---it took forever to find some time to read the paper... :)

tests/ada_grad_test.cpp
```diff
@@ -33,4 +33,4 @@ TEST_CASE("AdaGradLogisticRegressionTestFMat", "[AdaGradTest]")
 {
   AdaGrad adagrad(0.99, 32, 1e-8, 5000000, 1e-9, true);
   LogisticRegressionFunctionTest<arma::fmat>(adagrad, 0.003, 0.006);
-}
+}
```
rcurtin, line 36: Maybe we should revert this?  I don't see a reason to remove the trailing newline :)

tests/ada_sqrt_test.cpp
```diff
@@ -0,0 +1,109 @@
+/**
+ * @file ada_grad_test.cpp
+ * @author Abhinav Moudgil
+ * @author Marcus Edel
+ * @author Conrad Sanderson
```
rcurtin, line 5:
```suggestion
 * @file ada_sqrt_test.cpp
 * @author Marcus Edel
```

tests/ada_grad_test.cpp
zoq: Maybe I missed something but I think we don't use an extra trailing newline for other test files.

zoq: Nevermind, I see what you mean, reverted with the last commit.

include/ensmallen_bits/ada_sqrt/ada_sqrt_update.hpp
```diff
@@ -0,0 +1,123 @@
...
+     * @param stepSize Step size to be used for the given iteration.
+     * @param gradient The gradient matrix.
+     */
+    void Update(MatType& iterate,
+                const double stepSize,
+                const GradType& gradient)
+    {
+      ++parent.iteration;
+
+      squaredGradient += (gradient % gradient);
```
conradsnicta, line 100:
```suggestion
      squaredGradient += arma::square(gradient);
```

zoq: Not sure why I missed this again, thanks.

include/ensmallen_bits/ada_sqrt/ada_sqrt.hpp
```diff
@@ -0,0 +1,168 @@
...
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * terms of the 3-clause BSD license. You should have received a copy of the
+ * 3-clause BSD license along with ensmallen. If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_ADA_SQRT_ADA_SQRT_HPP
+#define ENSMALLEN_ADA_SQRT_ADA_SQRT_HPP
+
+#include <ensmallen_bits/sgd/sgd.hpp>
+#include "ada_sqrt_update.hpp"
```
conradsnicta, line 17: The style of these two `#inlcude` statements differs, which smells weird.
Note that both gcc and clang (and probably other compilers) interpret `#include <file>` and `#include "file"` slightly differently.  The former (with angle brackets) means that usual system paths (and any paths specified via `-I`) are to be searched first, possibly omitting the current directory.  The latter (with double quotes) means that the current directory (where the present file being processed resides) is to be searched first.  

So having two separate styles for `#include` here looks like a bug.  At best this may blow up with `file not found` during compilation.  At worst this may lead to subtle errors if a user happens to have two separate installations of ensmallen on their system (say an old one in `/usr/include`).

include/ensmallen_bits/ada_sqrt/ada_sqrt_impl.hpp
```diff
@@ -0,0 +1,41 @@
...
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_ADA_SQRT_ADA_SQRT_IMPL_HPP
+#define ENSMALLEN_ADA_SQRT_ADA_SQRT_IMPL_HPP
+
+// In case it hasn't been included yet.
+#include "ada_sqrt.hpp"
```
conradsnicta, line 16: Is this `#include` necessary?

`ada_sqrt_impl.hpp` gets included from `ada_sqrt.hpp`, which is in turn included from `ensmallen.hpp`.

If I recall correctly, at the outset we decided not to support usage of only specific optimisers, as it's too messy (maintenance nightmare) and potentially buggy.  Only usage via `#include <ensmallen.hpp>` is supported.

conradsnicta (changes requested):

include/ensmallen_bits/ada_sqrt/ada_sqrt.hpp
zoq: Good point, I don't mind to change it, especially since I mainly used it for convenience, that said we used the different styles in some other places as well, so I'll open another PR to unify the style across the code base.

include/ensmallen_bits/ada_sqrt/ada_sqrt_impl.hpp
zoq: Not at all, it's an artifact from the mlpack code base, I'll remove it and also adapt existing optimizer in another PR.

conradsnicta (approved):

include/ensmallen_bits/ada_sqrt/ada_sqrt_impl.hpp
rcurtin: Yeah, I originally started doing that years ago to be nice to users who strangely included an `_impl.hpp` only. That makes more sense in the context of mlpack where you could choose multiple headers to include, but here in ensmallen it may not be needed, since users should only include the main ensmallen header.

PR #236, opened 2020-10-27
Make sure there's a space in release script output

rcurtin:
I mean, this is not a very substantive pull request.  But, it would display wrong before... :)


zoq (approved):
Nice catch.

PR #248, opened 2021-01-31
add short example program; expand README with example installation

conradsnicta:
This is to partially address the concerns raised by JMLR reviewer 1.
Specifically, the short example program has been added to satisfy the following comment:

> The documentation of ensmallen is detailed but I missed some simple examples in the README file. There are examples in the documentation, but I believe having a few compileable-then-runnable examples in the repository would be a great improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,20 @@ If CMake is not already available on your system, it can be obtained from https:
 If you are using an older system such as RHEL 7 or CentOS 7,
 an updated version of CMake is also available via the EPEL repository via the `cmake3` package.
 
+Example installation:
+
+```
+mkdir build
+cd build
+cmake ..
+sudo make install
+```
+
+
+### Example Usage
+
+See `example.cpp` for example usage of the L-BFGS optimizer.
+
 
 ### License
 
@@ -36,14 +50,15 @@ Unless stated otherwise, the source code for **ensmallen** is licensed under the
 "LICENSE.txt" file.  You may also obtain a copy of the License at
 http://opensource.org/licenses/BSD-3-Clause .
 
+
 ### Citation
 
 Please cite the following paper if you use ensmallen in your research and/or
 software. Citations are useful for the continued development and maintenance of
 the library.
 
-* S. Bhardwaj, R. Curtin, M. Edel, Y. Mentekidis, C. Sanderson.
-  [ensmallen: a flexible C++ library for efficient function optimization](http://www.ensmallen.org/files/ensmallen_2018.pdf).
+* S. Bhardwaj, R. Curtin, M. Edel, Y. Mentekidis, C. Sanderson.  
+  [ensmallen: a flexible C++ library for efficient function optimization](http://www.ensmallen.org/files/ensmallen_2018.pdf).  
   Workshop on Systems for ML and Open Source Software at NIPS 2018.
 
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -28,6 +28,20 @@ If CMake is not already available on your system, it can be obtained from https:
...
+mkdir build
+cd build
+cmake ..
+sudo make install
+```
+
+
+### Example Usage
+
+See `example.cpp` for example usage of the L-BFGS optimizer.
```
coatless, line 43: Maybe add a link to the example and briefly describe the setting?

```suggestion
See [`example.cpp`](example.cpp) for example usage of the L-BFGS optimizer in a linear regression setting.
```

example.cpp
```diff
@@ -0,0 +1,68 @@
...
+  for (size_t i = 0; i < n_points; ++i)
+  {
+    double a = arma::randu();
+    X(1, i) += a;
+    y(i)    += a;
+  }
+  
+  LinearRegressionFunction lrf(X, y);
+  
+  ens::L_BFGS opt;
```
zoq, line 57:
```suggestion
  // create an Limited-memory BFGS optimizer object with default parameters
  ens::L_BFGS opt;
```

Just to provide some additional information.

example.cpp
```diff
@@ -0,0 +1,68 @@
...
+  opt.MaxIterations() = 10;
+  
+  arma::vec theta(n_dims, arma::fill::randu);   // initial point (uniform random)
```
zoq, line 60:
```suggestion
  // initial point (uniform random)
  arma::vec theta(n_dims, arma::fill::randu);
```

Just to be consistent with the comment style.

README.md
conradsnicta: thanks :)

example.cpp
conradsnicta: good idea

zoq (approved):
Looks good to me.

example.cpp
```diff
@@ -0,0 +1,70 @@
...
+  for (size_t i = 0; i < n_points; ++i)
+  {
+    double a = arma::randu();
+    X(1, i) += a;
+    y(i)    += a;
+  }
+  
+  LinearRegressionFunction lrf(X, y);
+  
+  // create an Limited-memory BFGS optimizer object with default parameters
```
rcurtin, line 57:
```suggestion
  // create a Limited-memory BFGS optimizer object with default parameters
```

example.cpp
```diff
@@ -0,0 +1,70 @@
...
+  
+  // initial point (uniform random)
+  arma::vec theta(n_dims, arma::fill::randu);
+  
+  opt.Optimize(lrf, theta);
+  
+  // theta now contains the optimized parameters
+  theta.print("theta:");
+  
+  return 0;
```
rcurtin, line 69:
```suggestion
```
Depending on your level of pedantry: technically this isn't necessary in C++.  Up to you whether you want to remove it---it doesn't make a difference to me.

rcurtin (approved):
Awesome, thanks @conradsnicta for adding this!  I think it might be useful to also add a link in the documentation, so that when a user goes to http://ensmallen.org/docs.html they can find this example too.  I'm not sure what the best way to do that is, but honestly just a link to the Github file would be sufficient, I think.  Once we merge this, I'll open a quick PR adding that link.

rcurtin:
I added a link to the file on ensmallen.org. :+1:

PR #252, opened 2021-02-13
Fix Report callback when no steps were taken.

### LINKED ISSUE #246: Report callback fails when optimizer has no iterations
#### Issue description

I'm using ensmallen via rcppEnsmallen. I've noticed this issue while using the L-BFGS solver. The issue is that the optimization process crashes my R session when the optimization quits without taking any steps while using the built-in report callback (likely due to bad initial parameter estimates). The optimizer does not crash the R session when the report callback is not used. The crash occurs while printing the report, which leads me to believe that there is something within the callback causing the issue. My guess is that the callback assumes that the number of iterations is [greater than 0](https://github.com/mlpack/ensmallen/blob/cdfae670da608ef8498dee1e649f79f1d3279ef9/include/ensmallen_bits/callbacks/report.hpp#L112-L130), which would not be the case when the optimizer fails to take a step.

#### Your environment

 * version of ensmallen: 2.14.2
 * operating system: RedHat 7.7
 * compiler: g++
 * version of Armadillo: 10.1.0

#### Steps to reproduce

Unfortunately, I don't have a reproducible example. My environment is also a bit locked, so I can't readily test any fixes.

#### Expected behavior

Report to print with full details.
```
Optimization Report
--------------------------------------------------------------------------------

Initial coordinates: 
-0.4538 -0.3340  1.1227  ... -0.2663

Final coordinates: 
-0.2655 -0.3284  1.1241  ... -0.1279
iter          loss          loss change   |gradient|    total time    
0             11135977605.1090.000         25055627432.5180.006         

--------------------------------------------------------------------------------

Version:
ensmallen:                    2.14.2 (No Direction Home)
armadillo:                    10.1.0 (Orchid Ambush)

Function:
Coordinates rows:             100
Coordinates columns:          1

Loss:
Initial                       11135977605.109
Final                         11135977605.109
Change                        0.000

Optimizer:
Iterations:                   1
Coordinates max. norm:        25055627432.518
Evaluate calls:               2
Gradient calls:               2
Time (in seconds):            0.006
```

#### Actual behavior

Example bad report output with R session crash:
```
Optimization Report
--------------------------------------------------------------------------------

Initial coordinates: 
 0.0000  0.0000  0.0000  ... 0.0000

Final coordinates: 
 0.2617  0.0238  0.0064  ... -0.0704
iter          loss          loss change   |gradient|    total time    

--------------------------------------------------------------------------------

Version:
ensmallen:               
```

rcurtin:
This fixes #246.

Basically, the `Report` callback was failing (segfaulting) when an optimizer took no steps at all.  These changes try to handle that a bit more gracefully, and print a warning for the user, since they may need to investigate why the optimizer took no steps.

### DIFF
--- a/include/ensmallen_bits/callbacks/report.hpp
+++ b/include/ensmallen_bits/callbacks/report.hpp
@@ -178,7 +178,7 @@ class Report
     if (tookStep)
       output << objectives.size() << std::endl;
     else
-      output << "0 (No steps taken!  Did optimization fail?)" << std::endl;
+      output << "0 (No steps taken! Did the optimization fail?)" << std::endl;
 
     if (epochCalls > 0)
     {
--- a/include/ensmallen_bits/callbacks/report.hpp
+++ b/include/ensmallen_bits/callbacks/report.hpp
@@ -149,6 +149,15 @@ class Report
     output << coordinates.n_cols << std::endl;
     output << std::endl;
 
+    // If we did not take any steps, at least fill what the initial objective
+    // was.
+    const bool tookStep = (objectives.size() > 0);
+    if (objectives.size() == 0)
+    {
+      objectives.push_back(function.Evaluate(coordinates));
+      timings.push_back(optimizationTimer.toc());
+    }
+
     output << "Loss:" << std::endl;
     PrettyPrintElement("Initial", 30);
     output << objectives[0] << std::endl;
@@ -166,7 +175,10 @@ class Report
       output << optimizerStream.str();
 
     PrettyPrintElement("Iterations:", 30);
-    output << objectives.size() << std::endl;
+    if (tookStep)
+      output << objectives.size() << std::endl;
+    else
+      output << "0 (No steps taken!  Did optimization fail?)" << std::endl;
 
     if (epochCalls > 0)
     {
@@ -183,7 +195,7 @@ class Report
       output << stepsizes.back() << std::endl;
     }
 
-    if (hasGradient)
+    if (hasGradient && gradientsNorm.size() > 0)
     {
       PrettyPrintElement("Coordinates max. norm:", 30);
       output << *std::max_element(std::begin(gradientsNorm),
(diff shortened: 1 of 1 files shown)


include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -166,7 +175,10 @@ class Report
       output << optimizerStream.str();
 
     PrettyPrintElement("Iterations:", 30);
-    output << objectives.size() << std::endl;
+    if (tookStep)
+      output << objectives.size() << std::endl;
+    else
+      output << "0 (No steps taken!  Did optimization fail?)" << std::endl;
```
coatless, line 181:
```suggestion
      output << "0 (No steps taken! Did the optimization fail?)" << std::endl;
```

include/ensmallen_bits/callbacks/report.hpp
```diff
@@ -149,6 +149,15 @@ class Report
     output << coordinates.n_cols << std::endl;
     output << std::endl;
 
+    // If we did not take any steps, at least fill what the initial objective
+    // was.
+    const bool tookStep = (objectives.size() > 0);
+    if (objectives.size() == 0)
+    {
+      objectives.push_back(function.Evaluate(coordinates));
```
zoq, line 157: This will fail since not every function implements `Evaluate(coordinates)`, I think in most cases an optimizer will run Evaluate before the actual optimization starts, so one solution might be to check if `evaluateCalls > 0` and then return `objective` which is set during a call of `Evaluate`. In case it's not set we could return `NaN`.

include/ensmallen_bits/callbacks/report.hpp
rcurtin: Thanks!  I agree this is clearer.

include/ensmallen_bits/callbacks/report.hpp
rcurtin: Ohh, good point.  I think that's a better idea.  Updated in 50439fb; let me know what you think. :+1:

zoq (approved):
No further comments from my side, looks good.

PR #255, opened 2021-02-21
Fix compilation issue when ENS_USE_OPENMP is specified.

rcurtin:
I noticed this issue while testing. It seems to be a bug introduced by #249.  I guess the system I was testing on somehow did not set `ENS_USE_OPENMP`.


zoq (approved):
Looks good to me.

birm (approved):

PR #258, opened 2021-02-26
Use normal distribution for CNE initial population generation

### LINKED ISSUE #247: Initial population generation in CNE
CNE seems to generate the initial population by adding uniform noise to the initial guess, instead of Gaussian noise as stated in the comment. 

https://github.com/mlpack/ensmallen/blob/master/include/ensmallen_bits/cne/cne_impl.hpp#L91

```
// Generate the population based on a Gaussian distribution around the given
// starting point.
std::vector<BaseMatType> population;
for (size_t i = 0 ; i < populationSize; ++i)
{
  population.push_back(arma::randu<BaseMatType>(iterate.n_rows, iterate.n_cols) + iterate);
}
```

Possibly, the intended loop body is something like:

```
population.push_back(mutationSize * arma::randn<BaseMatType>(iterate.n_rows, iterate.n_cols) + iterate);
```

rcurtin:
This fixes #247.  I chose against multiplying the normal distribution by `mutationSize` because the documentation only says that it uses a `normal distribution` (e.g. variance 1).  I'm totally open to other strategies here; I don't have any particular opinion on what is best.

I also undid an accidental change from another PR, and commented out the random seed for each test run.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,9 @@
  * Fix test compilation issue when `ENS_USE_OPENMP` is set
    ([#255](https://github.com/mlpack/ensmallen/pull/255)).
 
+ * Fix CNE initial population generation to use normal distribution
+   ([#258](https://github.com/mlpack/ensmallen/pull/258)).
+
 ### ensmallen 2.16.0: "Severely Dented Can Of Polyurethane"
 ###### 2021-02-11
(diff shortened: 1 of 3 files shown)


zoq (approved):
Great!

birm (approved):

PR #262, opened 2021-03-03
Nits on NSGA2

jonpsy:
Hi @say4n , I've been reading your code on NSGA2 and I really like it :).  I've made some small corrections. 

- minor grammar fix
- replace std::endl with "\n"

PR #263, opened 2021-03-04
Improvise NSGA2

jonpsy:
Continuation of #262 .

The goal of this PR is to increase the efficiency of this code, enhance its expressibility and reducing its code size as much as possible.

Summary of my goals:

* Minimize usage of raw ```for``` loops
* Employing std & arma algorithms whenever possible
* Potential error fixes.
* Introduce OMP parallelism [https://ieeexplore.ieee.org/document/6968779] whether on this class or a separate class.

I have a few questions before I do that, let me know your thoughts on that.

Providing logs for the convenience of the reader. Stay tuned for more progress updates!

**LOGS**

**Main Loop**
- [x] Initial population should be within (lowerBound, upperBound) so that they exist inside the solution space [link](https://github.com/mlpack/ensmallen/pull/263#discussion_r588002920)
- [x] Incorporate CheckArbitraryFunctionTypeAPI<> for NSGA2.
- [x]  Add RequireDenseFloatingType for NSGA2

**Fast Non Dominated Sorting**
- [x] Discard the empty vector appended to the front. [link](https://github.com/mlpack/ensmallen/pull/263#discussion_r587526998)
- [x]  ```fronts.size() > 0``` ==> ```!fronts.empty()```

**Crowding Distance Assignment**
- [x] Pull the ```crowdingDistance``` 0 initialisation out of the function 
- [x] Remove ```fronts.size() > 0 ``` check, since empty front isn't passed at all. Refer the above section.
- [x]  ```crowdingDistance``` is filled using objective values not by taking average of ```crowdingDistance```
- [x]   ~The calculation seems to be very roundabout, need to confirm with @say4n. If you're reading this, can you check this [patch](https://drive.google.com/file/d/1KjrbkkKL6eviF7dUEjsCNr8BCW-6c4rg/view?usp=sharing)~ As discussed, going with the patch.

**Testing**
- [x] Check for different types of matrices, for ex: ```fmat```


include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -113,6 +114,7 @@ typename MatType::elem_type NSGA2::Optimize(
 
   // Generate the population based on a uniform distribution around the given
   // starting point.
+  //TODO: Shouldn't we check if they're inside the upperBound && lowerBound?
   for (size_t i = 0; i < populationSize; i++)
   {
     population.push_back(arma::randu<MatType>(iterate.n_rows,
```
jonpsy, line 120: Shouldn't we check if they're inside the upperBound && lowerBound? Am I missing something? 

An example https://github.com/baopng/NSGA-II/blob/29a6ec33f87b32a7fb2596091e1a51c897106e7b/nsga2/problem.py#L20

jonpsy, line 120: Shouldn't we check if they're inside the upperBound && lowerBound? Am I missing something? 

An example https://github.com/baopng/NSGA-II/blob/29a6ec33f87b32a7fb2596091e1a51c897106e7b/nsga2/problem.py#L20

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -405,6 +408,7 @@ inline void NSGA2::CrowdingDistanceAssignment(const std::vector<size_t>& front,
 
     for (size_t m = 0; m < numObjectives; m++)
     {
+        //TODO: This should be out of the loop?
       crowdingDistance[front[0]] = std::numeric_limits<double>::max();
       crowdingDistance[front[fSize - 1]] = std::numeric_limits<double>::max();
```
jonpsy, line 413: Setting the same value to the crowdingDistance value of a front multiple times.

jonpsy:
Requesting reviews @rcurtin  @zoq  @ShikharJ since you've mentored the original PR. If I've gone ahead of myself with these changes, please do tell me :).

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -365,6 +365,8 @@ inline void NSGA2::FastNonDominatedSort(
     i++;
     fronts.push_back(nextFront);
   }
+  //Remove the empty final set
+  fronts.pop_back();
```
jonpsy, line 369: Caught a big fish here. Allow me to explain myself,

- The while loop breaks when the newFront is empty. Note that the newFront is pushed **regardless** of any of the conditions.
- At the end of our while loop, the nextFront is not updated at all! That means its an empty vector which gets pushed into our ```front```
-  This causes the while loop to terminate

Result?

You got an additional empty vector in your ```front``` set. Which causes smelly code.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -396,9 +398,7 @@ inline bool NSGA2::Dominates(
 inline void NSGA2::CrowdingDistanceAssignment(const std::vector<size_t>& front,
                                               std::vector<double>& crowdingDistance)
 {
-  if (front.size() > 0)
-  {
-    for (size_t elem: front)
```
jonpsy, line 401: Continuing on the "big fish", this causes this redundant if clause because of that empty vector.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -396,9 +398,7 @@ inline bool NSGA2::Dominates(
...
+    for (size_t elem: front) //TODO: When looping over all the fronts, all will be initialized to zero right?
       crowdingDistance[elem] = 0;
```
jonpsy, line 402: Since we're looping over all the fronts, can we not simply initialize ```crowdDistance``` ahead of calling this function?

include/ensmallen_bits/nsga2/nsga2_impl.hpp
jonpsy: https://github.com/haris989/NSGA-II/blob/7680fd70c5202d0673aa26b9fb50a8e9c387623b/NSGA%20II.py#L71

include/ensmallen_bits/nsga2/nsga2_impl.hpp
jonpsy: https://github.com/haris989/NSGA-II/blob/7680fd70c5202d0673aa26b9fb50a8e9c387623b/NSGA%20II.py#L76, notice there are no checks

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -92,16 +92,17 @@ typename MatType::elem_type NSGA2::Optimize(
   numVariables = iterate.n_rows;
 
   // Cache calculated objectives.
-  std::vector<arma::Col<ElemType> > calculatedObjectives;
+  std::vector<arma::Col<ElemType> > calculatedObjectives; //TODO: Size is fixed, perhaps use matrix?
```
zoq, line 95: The size of the objectives depends on the size of the population, which can increase in case of a crossover, so I don't think you can get away with a const matrix.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
zoq: I have to check the paper, not sure if the constrained was only for the mutation/crossover step or for the initial population as well. Also might be good to reference the paper instead of another python implementation.

zoq:
> Summary of my goals:
> 
> * Minimize usage of raw `for` loops
> * Employing std & arma algorithms whenever possible
> * Potential error fixes.
> * Introduce OMP parallelism [https://ieeexplore.ieee.org/document/6968779] whether on this class or a separate class.

Sounds good to me, if we can show that an OpenMP solution is faster let's integrate it into the current method, but maybe armadillo build against openblas is already faster, haven't read the paper yet.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
jonpsy:
```c++
	  
	  /*if limits are specified it generates the value in 
	    range of minimum and maximum value of the variable*/
	  else
	    {
	      pop_ptr->ind[i].xreal[j] = d*(lim_r[j][1] - lim_r[j][0])+lim_r[j][0];
	    }
```

Source: IITK Kanpur Genetics Laboratory, Original Implementation in C

include/ensmallen_bits/nsga2/nsga2_impl.hpp
jonpsy: The sorting part inside the CrowdingDistance method not implemented.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -139,9 +142,9 @@ typename MatType::elem_type NSGA2::Optimize(
     BinaryTournamentSelection(population, lowerBound, upperBound);
 
     // Evaluate the objectives for the new population.
-    calculatedObjectives.resize(population.size());
-    for (size_t i = 0; i < population.size(); i++)
-      calculatedObjectives[i] = arma::Col<ElemType>(numObjectives, arma::fill::zeros);
+    calculatedObjectives.resize(population.size()); //TODO: Costly? Doing because population may change??
+    std::fill(calculatedObjectives.begin(), calculatedObjectives.end(),
```
conradsnicta, line 146: I'd be wary of replacing simple loops with algorithms from the std namespace. I don't think there is any performance advantage -- in the end the code will do exactly the same thing.  At the same time, this change makes things harder to read/understand, which adds to the technical debt (ie. increases the maintenance burden).

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -150,11 +153,12 @@ typename MatType::elem_type NSGA2::Optimize(
 
     // Perform crowding distance assignment.
     crowdingDistance.resize(population.size());
-
-    for (size_t fNum = 0; fNum < fronts.size(); fNum++)
+    std::fill(crowdingDistance.begin(), crowdingDistance.end(), 0.f);
```
conradsnicta, line 156: Again, a simple and effective 'for' loop was replaced with something more elaborate.  There is no need for this.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
jonpsy: I had a feeling I was overdoing it. Thanks for the review.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
jonpsy: Are you talking about the std::fill or std::for_each?

jonpsy: Because I moved crowdingDistance zero intialisation out from the function. So instead of setting 0 to that vector per front, I did it at once. 

If however you're talking about std::for_each, I'll revert that change.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
zoq: Nice catch!

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -117,6 +118,16 @@ typename MatType::elem_type NSGA2::Optimize(
...
+    
+    // Ensure population is within variable space
+    for(size_t geneIdx = 0; geneIdx < numVariables; geneIdx++)
+    {
+      if(population[i][geneIdx] < lowerBound[geneIdx])
+        population[i][geneIdx] = lowerBound[geneIdx];
+      
+      if(population[i][geneIdx] > upperBound[geneIdx])
+        population[i][geneIdx] = upperBound[geneIdx];
+    }
```
zoq, line 130:
```suggestion
    // Ensure population is within variable space.
    for (size_t geneIdx = 0; geneIdx < numVariables; geneIdx++)
    {
      if (population[i][geneIdx] < lowerBound[geneIdx])
        population[i][geneIdx] = lowerBound[geneIdx];
      
      if (population[i][geneIdx] > upperBound[geneIdx])
        population[i][geneIdx] = upperBound[geneIdx];
    }
```

We tend to be picky about the style.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -176,6 +187,7 @@ typename MatType::elem_type NSGA2::Optimize(
     );
 
     // Yield a new population P_{t+1} of size populationSize.
+    // Discards unfit population from the R_{t} to yield P_{t+1}
```
zoq, line 190:
```suggestion
    // Discards unfit population from the R_{t} to yield P_{t+1}.
```

Pedantic style fix.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -365,6 +377,8 @@ inline void NSGA2::FastNonDominatedSort(
     i++;
     fronts.push_back(nextFront);
   }
+  //Remove the empty final set
```
zoq, line 380:
```suggestion
  // Remove the empty final set.
```

Another minor style fix, not sure I mentioned this at some point, but ensmallen follows https://github.com/mlpack/mlpack/wiki/DesignGuidelines.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -150,7 +161,7 @@ typename MatType::elem_type NSGA2::Optimize(
 
     // Perform crowding distance assignment.
     crowdingDistance.resize(population.size());
-
+    std::fill(crowdingDistance.begin(), crowdingDistance.end(), 0.f);
```
zoq, line 164: I guess before the reset to zero, we got lucky, because in the `CrowdingDistanceAssignment` method we use `front[i + 1]` which should be undefined:

```
crowdingDistance[front[i]] += (crowdingDistance[front[i - 1]] -
            crowdingDistance[front[i + 1]])
```

zoq: Ahh, nevermind we set it to zero before as well.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -365,6 +377,8 @@ inline void NSGA2::FastNonDominatedSort(
     i++;
     fronts.push_back(nextFront);
   }
+  //Remove the empty final set
+  fronts.pop_back();
```
zoq, line 381: Looks like this causes a bad alloc.

jonpsy: The bad alloc is because I used ```!fronts.empty()``` instead of ```!fronts[i].empty()```. Lemme fix that ;)

include/ensmallen_bits/nsga2/nsga2_impl.hpp
jonpsy: @zoq  Regarding this, I think this implementation is wrong. Here's an excerpt from the paper

crowdingdistance

Notice the equation should've been 
```c++
crowdingDistance[front[i]] += (fValue[i+1] - fValue[i-1])/(maxFval - minFval)
```

and not the average of crowdingDistances

jonpsy: I've implemented it in this  [patch](https://drive.google.com/file/d/1KjrbkkKL6eviF7dUEjsCNr8BCW-6c4rg/view)

zoq: Good, can you make the patch part of the PR?

jonpsy: I've no problem with that. Thing is, @say4n  has implemented assignment first and sorting later which I think is wrong but I guess he must have some reason. I was waiting for his review, very well, I'll add this patch on my PR.

say4n: I think I had unrolled the first iteration of some parts of the optimisation loop to be outside the for loop. This had the benefit of instantiating some items as far as I recall. I don't remember all the details now so do take this with a grain of salt! :)

jonpsy: Thanks for clarifying :). I'd say let's merge my patch for now as it is straightforward & adheres to the research paper directly. We can then try your optimization method and see how it all fits in.

say4n: Yes, for sure! Great work btw!!

jonpsy:
All the tests pass, Wohoo!. All the tasks are done as well. @zoq Do you have any comments?

zoq:
> All the tests pass, Wohoo!. All the tasks are done as well. @zoq Do you have any comments?

I'll do another pass over the PR, in the meantime can you update the `HISTORY.md` so it passes the Travis CI check.

include/ensmallen_bits/nsga2/nsga2.hpp
```diff
@@ -180,7 +180,7 @@ class NSGA2 {
    * @tparam MatType Type of matrix to optimize.
    * @param population The elite population.
    * @param objectives The set of objectives.
-   * @param calculatedObjectives Vector to store calculated objectives.
+   * @param populationFval Vector to store calculated objectives.
```
zoq, line 183: I'm not sure I see the reason to change the name of the parameter here, but I don't mind to change it, but let's choose a more descriptive name here, I mean `population` is clear, but `Fval` is not really a parameter name that describes what the parameter is about. Let me know what you think.

include/ensmallen_bits/nsga2/nsga2.hpp
```diff
@@ -250,13 +250,13 @@ class NSGA2 {
    * @param fronts The population is sorted into these Pareto fronts. The first
    *     front is the best, the second worse and so on.
    * @param ranks The assigned ranks, used for crowding distance based sorting.
-   * @param calculatedObjectives The previously calculated objectives.
+   * @param populationFval The previously calculated objectives.
```
zoq, line 253: See comment above.

include/ensmallen_bits/nsga2/nsga2.hpp
```diff
@@ -266,46 +266,28 @@ class NSGA2 {
    * objective for which it is strictly better than the other candidate.
    *
    * @tparam MatType Type of matrix to optimize.
-   * @param calculatedObjectives The previously calculated objectives.
+   * @param populationFval The previously calculated objectives.
```
zoq, line 269: See comment above.

include/ensmallen_bits/nsga2/nsga2.hpp
```diff
@@ -266,46 +266,28 @@ class NSGA2 {
...
-   * @return true if the first candidate is preferred, otherwise, false.
    */
-  bool CrowdingOperator(size_t idxP,
-                        size_t idxQ,
-                        const std::vector<size_t>& ranks,
-                        const std::vector<double>& crowdingDistance);
+  template <typename MatType>
+  void CrowdingDistanceAssignment(
+	  const std::vector<size_t>& front,
+	  std::vector<arma::Col<typename MatType::elem_type>>& populationFval,
```
zoq, line 289: See comment above.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -139,43 +150,26 @@ typename MatType::elem_type NSGA2::Optimize(
...
-
-                  if (arma::approx_equal(population[i], candidateQ, "absdiff", epsilon))
-                    idxQ = i;
-                }
-
-                return CrowdingOperator(idxP, idxQ, ranks, crowdingDistance);
-              }
-    );
+		  CrowdingDistanceAssignment<MatType>(
+			  fronts[fNum], populationFval, crowdingDistance);
```
zoq, line 168: Looks like the indentation is wrong here, it should be two spaces, and the next line should tabbed twice (4 spaces).

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -365,72 +359,76 @@ inline void NSGA2::FastNonDominatedSort(
...
-
-    for (size_t m = 0; m < numObjectives; m++)
+  for (size_t m = 0; m < numObjectives; m++)
+  {
+    std::vector<double> fValues{};
+    std::transform( // Cache the Fvalues of individuals for current objective.
+            front.begin(), front.end(), std::back_inserter(fValues),
+            [&](const size_t& individual) {
+                return populationFval[individual](m);
+            });
```
zoq, line 409: Should be tabbed twice (4 spaces).

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -365,72 +359,76 @@ inline void NSGA2::FastNonDominatedSort(
...
+  std::vector<size_t> sortedIdx(fSize); // Stores the sorted indices of the fronts.
+  std::iota(sortedIdx.begin(), sortedIdx.end(), 0u);
 
-    size_t fSize = front.size();
-
-    for (size_t m = 0; m < numObjectives; m++)
+  for (size_t m = 0; m < numObjectives; m++)
+  {
+    std::vector<double> fValues{};
+    std::transform( // Cache the Fvalues of individuals for current objective.
```
zoq, line 405: Should we use `fValues` here?

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -365,72 +359,76 @@ inline void NSGA2::FastNonDominatedSort(
...
+            [&](const size_t& individual) {
+                return populationFval[individual](m);
+            });
+
+    // Sort the front indices based on ascending FValue of the current objective.
+    std::sort(sortedIdx.begin(), sortedIdx.end(),
+        [&](const size_t& frontIdxA, const size_t& frontIdxB) {
+                          return (fValues[frontIdxA] <
+                                  fValues[frontIdxB]);
+                      });
```
zoq, line 416: See comments above.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -365,72 +359,76 @@ inline void NSGA2::FastNonDominatedSort(
...
+
+    crowdingDistance[front[sortedIdx[0]]] = std::numeric_limits<double>::infinity();
+    crowdingDistance[front[sortedIdx[fSize - 1]]] =
+            std::numeric_limits<double>::infinity();
```
zoq, line 420: Should be tabbed twice (4 spaces).

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -365,72 +359,76 @@ inline void NSGA2::FastNonDominatedSort(
...
-      for (size_t i = 1; i < fSize - 1 ; i++)
-      {
-        crowdingDistance[front[i]] += (crowdingDistance[front[i - 1]] -
-            crowdingDistance[front[i + 1]]) /
-            (std::numeric_limits<double>::max() -
-            std::numeric_limits<double>::min());
-      }
+				crowdingDistance[front[sortedIdx[i]]] +=
+				    (fValues[sortedIdx[i + 1]] - fValues[sortedIdx[i - 1]]) /
+				    scale;
```
zoq, line 429: Can you check the indentation here?

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -365,72 +359,76 @@ inline void NSGA2::FastNonDominatedSort(
...
+  size_t fSize = front.size();
+  std::vector<size_t> sortedIdx(fSize); // Stores the sorted indices of the fronts.
+  std::iota(sortedIdx.begin(), sortedIdx.end(), 0u);
 
-    size_t fSize = front.size();
-
-    for (size_t m = 0; m < numObjectives; m++)
+  for (size_t m = 0; m < numObjectives; m++)
+  {
+    std::vector<double> fValues{};
```
zoq, line 404: Looks like it would be possible to use an `arma::vec` here, unless I missed something and doing so would make sorting more complex?

jonpsy: Nice catch! :)

include/ensmallen_bits/nsga2/nsga2.hpp
jonpsy: My reasoning is ```calculatedObjectives``` is too vague and lengthy on top of it. 

_"calculatedObjectives of what?"_  A user might ask.

Perhaps ```Fval``` is bad naming, how about ```populationFitness```?

include/ensmallen_bits/nsga2/nsga2_impl.hpp
jonpsy: Indeed, we need to store the fValues for the current objective, or are you talking about the naming?

include/ensmallen_bits/nsga2/nsga2_impl.hpp
jonpsy: So, if we're doing an assignment in a new line we should tab twice, right?

jonpsy:
Perhaps a good idea to add this too?

```c++
  traits::CheckArbitraryFunctionTypeAPI<
      FunctionType, BaseMatType>();
  RequireDenseFloatingPointType<BaseMatType>();
```

include/ensmallen_bits/nsga2/nsga2_impl.hpp
zoq: I was talking about the name so the comment matches the parameter name.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
zoq: Yes, everything that doesn't fit in the 80 character limit has to be tabbed twice.

jonpsy:
AppVeyor is failing, but the tests work, can you throw some light on this please?

zoq:
Strange, I have to debug this on a local machine, but looks like the windows build never passed on this PR.

jonpsy:
On second thought, I think its better to keep ```calculatedObjectives``` because fitness doesn't always equate to calculated objective value. For ex: SPEA-II has a clearly distinguished fitness and objective value.

jonpsy:
Let's have #275 merged, so that all the test pass. Had to force-push because of SchafferN1 fix (another force-push coming for because of fonseca =P). Cheers!

jonpsy:
The double and float comparison is causing some really interesting test failures.

zoq (commented):
Looks like you changed the appveyor build settings from debug to release, does that mean ensmallen doesn't build in debug mode?

jonpsy:
No, I was just messing around(for appveyor RDP to Windows) . It's really weird that it fails on SchafferN1Double but passes on SchafferN1DoubleVector!? 

Similarly, passes for SchafferN1Float but fails on SchafferN1FloatVector.

jonpsy:
The ```numGeneration``` value is suspiciously high in our tests (5000), but this confirms it. The paper (and any MOEAD algorithm for that matter) uses ~250 ```numGenerations```. What happened was, this high number of generation lead to a "overfit", our pareto points totally lost diversity.

Infact, there were many coinciding points which is a nightmare for any MOO algorithm. I fixed the default arguments in the test for ```mutationRate``` , ```crossoverProb``` and put the ```numGenerations``` as 250 (see figure below). All the tests pass with flying colors on VS2019.

image

jonpsy:
Test seems to pass when I run them manually on Appveyor but fails when run through CTest (MSVC-2015,17). Tests pass in both CTest and manually on MSVC19 and Linux builds.

testpassed

zoq:
> Test seems to pass when I run them manually on Appveyor but fails when run through CTest (MSVC-2015,17). Tests pass in both CTest and manually on MSVC19 and Linux builds. @zoq, is there something different in CTest that I might be missing?
> 
> CroppedNSGA2Testpass2015

Nothing, no, this could just be a bad seed, did you run the tests locally like 1000x with a different seed to check if the test pass each time? Note you have to uncomment:

https://github.com/mlpack/ensmallen/blob/9f90ec92aa64fa03b83e1f5206af198d73f4359c/tests/main.cpp#L25-L27

zoq:
Hm, not sure I can see the fix in the last commits, was this just a strange windows fluke?

jonpsy:
I changed the tolerance from 1e-3 to 0.1, the problem was there were was one outlier in the Pareto front ```-0.025...``` which failed the test every time. (Bound is [0, 2])

Reg it being a fluke, well, it must be one really lucky fluke to be not caught in [3000](https://github.com/mlpack/ensmallen/pull/263/commits/f3f699a50879d56156082c98a7a2c48f9a4fd3ed) (this change passed the test) attempts :). I tried this stress testing on my local MSVC17,19 as well. 

Notice that the ```numGeneration``` parameter is now 300 because NSGA2 performance caps around this, no need of 4000.

Just to be 100% sure, I will share the pastebin for stress testing inside the Appveyor.

jonpsy:
> Hm, not sure I can see the fix in the last commits, was this just a strange windows fluke?

Your question is well-founded, it looks like Appveyor cmd does not repeat for 3000 times, even after putting ```---repeat-until-fail``` command. So I went ahead and run it manually for like ~50 times. Here's the pastebin (https://pastebin.com/k641DjC7). Let me know if you require further proof.

zoq:
> > Hm, not sure I can see the fix in the last commits, was this just a strange windows fluke?
> 
> Your question is well-founded, it looks like Appveyor cmd does not repeat for 3000 times, even after putting `---repeat-until-fail` command. So I went ahead and run it manually for like ~50 times for MSVC15. Here's the pastebin (https://pastebin.com/k641DjC7). For 100 times: https://pastebin.com/t4yMGdec.
> 
> Let me know if you require further proof.

Thanks, I just was just wondering what it was, thanks for tracking this down.

zoq:
I restarted the windows build. because the build failed, not sure it's because of the RDP session.

jonpsy:
The build fails because of RDP. Its waiting for us to access it, and terminates when we dont. @zoq , do we agree the algorithm is working? If so, I'll revert RDP changes and the build would pass.

zoq:
> The build fails because of RDP. Its waiting for us to access it, and terminates when we dont. @zoq , do we agree the algorithm is working? If so, I'll revert RDP changes and the build would pass.

Ahh, right I have seen you removed it from one section, but looks like not from the other one. This looks good, so let's remove it.

jonpsy:
Hi @zoq , is this GTM?

zoq (commented):
This looks really good to me, just one more really minor style issue.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -146,20 +160,21 @@ typename MatType::elem_type NSGA2::Optimize(
...
-      CrowdingDistanceAssignment(fronts[fNum], crowdingDistance);
+      CrowdingDistanceAssignment<BaseMatType>(
+          fronts[fNum], calculatedObjectives, crowdingDistance);
     }
 
     // Sort based on crowding distance.
     std::sort(population.begin(), population.end(),
-              [this, ranks, crowdingDistance, population](MatType candidateP,
-                                                          MatType candidateQ)
+              [this, ranks, crowdingDistance, population](BaseMatType candidateP,
```
zoq, line 176: Can you use the same indentation here you used for the other `sort` and `transform`.

jonpsy:
Once #283 is merged, I will incorporate that feature along with the rebase changes + static analysis fix changes.

zoq:
> Once #283 is merged, I will incorporate that feature along with the rebase changes + static analysis fix changes.

Nice, just merged #283.

jonpsy:
All right, waiting for the green tick :) .

jonpsy:
All green! 🥳

zoq (approved):
Thanks for another great contribution :+1:

zoq:
Thanks for digging into the optimization method, really nice improvements.

PR #265, opened 2021-03-06
Remove AdamSchafferFunctionN2Test test case

zoq:
Let's see if removing the approximation improves test stability.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -12,6 +12,9 @@
  * Fix compilation warnings
    ([#259](https://github.com/mlpack/ensmallen/pull/259)).
 
+ * Remove `AdamSchafferFunctionN2Test` test from Adam test suite to prevent
+   spurious issue on some aarch64 ([#265](https://github.com/mlpack/ensmallen/pull/259)).
+
 ### ensmallen 2.16.0: "Severely Dented Can Of Polyurethane"
 ###### 2021-02-11
  * Expand README with example installation and add simple example program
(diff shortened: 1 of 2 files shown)


rcurtin:
Any thoughts on what we should do here?  I'm still unable to reproduce the problem, so to me it's not clear which way we should go to try and solve it.  (Maybe make the starting point a little closer to the optimum to help convergence?)

rcurtin:
I thought about it a bit more.  Honestly, I would be fine simply removing the `AdamSchafferFunctionN2Test`; Adam is already tested very many different ways, so I have a feeling that the arm64 error (and other architectures) does not actually have to do with any incorrect computation---if it did, I would expect lots of other failing Adam tests too.  Let me know what you think.

zoq:
> I thought about it a bit more. Honestly, I would be fine simply removing the `AdamSchafferFunctionN2Test`; Adam is already tested very many different ways, so I have a feeling that the arm64 error (and other architectures) does not actually have to do with any incorrect computation---if it did, I would expect lots of other failing Adam tests too. Let me know what you think.

Same, thought here, I spend hours to reproduce the issue, and I don't think there is an issue with the implementation either. So I would also just remove or comment out the test for now.

rcurtin:
Yeah---not a very satisfying end to the bug hunt, but I think in this case it's the best we can do. :+1:  Do you want to make that change in this PR?  Or should I open another one for it?

zoq:
I can integrate it here.

zoq:
Ohh that is interesting, I guess we already knew that we would remove the test at some point.

https://github.com/mlpack/ensmallen/blob/39aa37d65533544b7c988d96614d3694413e69b0/tests/adam_test.cpp#L326-L330

rcurtin (approved):
Well, I guess the comment has come true---we did remove the test in the end. :)

rcurtin:
Ok, now let's have a release and hopefully that will fix our original issue. :)

PR #267, opened 2021-03-07
CNE distribution test

zoq:
With https://github.com/mlpack/ensmallen/pull/258/files we switched the distribution but with that change it looks like the CNE test fails on across all Windows builds, this is basically here to see if that was just a fluke or if that has something to do with the change.


rcurtin:
Interesting, I am surprised this seems to have caused an issue.  Do you think you can try running the `CNESchafferFunctionN4Test` on AppVeyor printing per-iteration output, both when `randu` and `randn` are used?

Also, I see a minor bug in the test implementation:

```c++
TEST_CASE("CNESchafferFunctionN4Test", "[CNETest]")
{
  SchafferFunctionN4 f;
  CNE optimizer(500, 1600, 0.3, 0.3, 0.3, -1);

  // We allow a few trials.
  for (size_t trial = 0; trial < 5; ++trial)
  {
    arma::mat coordinates = arma::mat("0.5; 2");
    optimizer.Optimize(f, coordinates);

    if (trial != 4)
    {
      if (coordinates(0) != Approx(0).margin(0.1))
        continue;
      if (abs(coordinates(1)) != Approx(1.25313).margin(0.1))
        continue;
    }

    REQUIRE(coordinates(0) == Approx(0).margin(0.1));
    REQUIRE(abs(coordinates(1)) == Approx(1.25313).margin(0.1));
  }
}
```

There should be a `break` before the end of the loop, so that if the test passes, we don't run another trial.

zoq:
Let's squash and merge this one, no need to have to debug commits in the history.

rcurtin (approved):
Yeah, I guess I could understand how the lack of `break` could cause the issue here.  If the AppVeyor seed (and others) just happened to fail on the last iteration, then this would effectively be a 1-trial test (since the first 4 results would be ignored).

Want to add a HISTORY message?  Then I think it's good for merge. :+1:

shrit (approved):
Good idea, I am not sure if this can be applied to the failing aarch64 issue in mlpack.

Issue #266, opened 2021-03-07
Embedded assembly code not working for Apple M1


joemsong:
Could the embedded assembly code be replaced for Apple M1 processor?

Thank you.

--------
% brew install --build-from-source ensmallen
==> Downloading https://github.com/mlpack/ensmallen/archive/2.16.1.tar.gz
Already downloaded: /Users/joesong/Library/Caches/Homebrew/downloads/7fe6be0479b5520276f672eddcec999809758c01481a2cd7f6d0441b415fc045--ensmallen-2.16.1.tar.gz
==> cmake ..
==> make install
Last 15 lines from /Users/joesong/Library/Logs/Homebrew/ensmallen/02.make:
    #define CATCH_BREAK_INTO_DEBUGGER() if( Catch::isDebuggerActive() ) { CATCH_TRAP(); }
                                                                          ^
/tmp/ensmallen-20210306-15863-bv9x46/ensmallen-2.16.1/tests/catch.hpp:5269:34: note: expanded from macro 'CATCH_TRAP'
    #define CATCH_TRAP() __asm__("int $3\n" : : ) /* NOLINT */
                                 ^
<inline asm>:1:6: note: instantiated into assembly here
        int $3
            ^
[ 40%] Building CXX object tests/CMakeFiles/ensmallen_tests.dir/iqn_test.cpp.o
cd /tmp/ensmallen-20210306-15863-bv9x46/ensmallen-2.16.1/build/tests && /opt/homebrew/Library/Homebrew/shims/mac/super/clang++  -I/tmp/ensmallen-20210306-15863-bv9x46/ensmallen-2.16.1/include -DNDEBUG -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -Wall -Wpedantic -Wunused-parameter -std=gnu++11 -o CMakeFiles/ensmallen_tests.dir/iqn_test.cpp.o -c /tmp/ensmallen-20210306-15863-bv9x46/ensmallen-2.16.1/tests/iqn_test.cpp
2 errors generated.
make[2]: *** [tests/CMakeFiles/ensmallen_tests.dir/main.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [tests/CMakeFiles/ensmallen_tests.dir/all] Error 2
make: *** [all] Error 2

zoq:
[...]

PR #268, opened 2021-03-07
Catch2 update 2.13.4

zoq:
Catch2 update 2.13.4 which will fix https://github.com/mlpack/ensmallen/issues/266.


rcurtin (approved):
I'm not sure why the memory tests are failing (looks like they ran fine); the static code analysis issues can be ignored; and the Windows build issues appear to be #267 (so this PR wouldn't solve them anyway).  So, looks good otherwise! :smile:

zoq:
> I'm not sure why the memory tests are failing (looks like they ran fine); the static code analysis issues can be ignored; and the Windows build issues appear to be #267 (so this PR wouldn't solve them anyway). So, looks good otherwise!

The memory job run into the build timeout which is set to 360 minutes, usually we just run the memory check on changed tests, but since an update of the catch2 header will trigger all tests to be run we run into the timeout. But agreed, save to ignore in this case.

rcurtin:
Ahh, ok, thanks---I must have just missed the 'job terminated' line in the log. :+1:

PR #270, opened 2021-03-10
Typo in LBFGS() method of AugLagrangian

matteosecli:

PR #271, opened 2021-03-14
Swapped default params in AugLagrangian docs

matteosecli:


zoq (commented):
Whoops, nice catch! Can you add the PR to the `HISTORY.md` file, there is already an entry `Fix typo in documentation (#270).` so if you can add the PR number to that entry that would be great.

matteosecli:
> Whoops, nice catch! Can you add the PR to the `HISTORY.md` file, there is already an entry `Fix typo in documentation (#270).` so if you can add the PR number to that entry that would be great.

Sure, done! 🙂

zoq (approved):
Thanks no more comments from my side.

zoq:
Thanks again for the contribution.

PR #273, opened 2021-03-16
Adding ZDT Test Suite

jonpsy:
Pizza!

**Declared files**
- [x] ```zdt1_function.hpp```
- [x] ```zdt2_function.hpp```
- [x] ```zdt3_function.hpp```
- [x] ```zdt4_function.hpp```
- [ ] ~zdt5_function.hpp~ [skipping for now]
- [x] ```zdt6_function.hpp```

 **TODO**

- [x] Write docs for each declared files.
- [x] Write the interface for each declared files.
- [x] Test working against [pagmo](https://esa.github.io/pagmo2/docs/cpp/problems/zdt.html)
- [x] Test numerical stability and corner cases & add to test case.
- [x] Add in nsga2_test.cpp
- [x] Add Pareto Front for each problem suite.

Pareto Fronts plotting: https://colab.research.google.com/drive/1Hgmvj4ZQVdT6OgopLPabldeyaOm0DInI?usp=sharing


jonpsy:
All this in just one day! Phew!

jonpsy:
Few updates:
a) I've tested all the ZDT(1-6 excluding 5) against pagmo, results match. I can create a dummy repository if one wants to verify this.
b) Interestingly, NSGA-II (current implementation) fails on this, but #263 passes, mostly owing to the population being out-of-bounds.
c) I think this is good to merge, critique's welcome.

include/ensmallen_bits/problems/zdt3_function.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * Evaluate the objectives with the given coordinate.
+     *
+     * @param coords The function coordinates.
+     * @return arma::Col<typename MatType::elem_type>
+     */
+    arma::Col<typename MatType::elem_type> Evaluate(const MatType& coords)
+    {
+      double pi = arma::datum::pi;
```
zoq, line 83: I would probably use it directly, since it's only used once and it's already an alias.

zoq (commented):
Looks pretty good to me, I made some comments about the style, I'll check the function implementation next, the optimizer is already able to solve the problems, so that is a good sign.

jonpsy:
> Looks pretty good to me, I made some comments about the style, I'll check the function implementation next, the optimizer is already able to solve the problems, so that is a good sign.

Just to make your job easier, I've created a repository comparing pagmo results to mine. https://github.com/jonpsy/MOO-TestSuite

zoq:
> > Looks pretty good to me, I made some comments about the style, I'll check the function implementation next, the optimizer is already able to solve the problems, so that is a good sign.
> 
> Just to make your job easier, I've created a [repository](https://github.com/jonpsy/MOO-TestSuite) comparing pagmo results to mine. Ofcourse, these are trivial cases, if you have any edge cases in mind do let me know so I can test and compare them. Cheers!

Usually I take a look at the function definition (paper, wikipedia) and compare it with the implementation, since there is a chance that an existing implementation does something different.

jonpsy:
I see. Here's the link to the research paper: https://www.mitpressjournals.org/doi/abs/10.1162/106365600568202

jonpsy:
Now that NSGA-II fix is close to merge, we can start working on this. ZDT5 is also remaining, but since it only works on  bitstring type I would need to use ```std::enable_if < is_bitset<....>>```, something along these lines.

jonpsy:
Keep open

jonpsy:
Updates on this: pymoo has implemented the Pareto front for each of these so I have included the Ipynb notebook in this pull request. I was thinking we could test if the Pareto Front from the Pymoo implementation can also be used to test our implementation, see if the outputs match? Your thoughts @zoq ?

zoq:
> Updates on this: pymoo has implemented the Pareto front for each of these so I have included the Ipynb notebook in this pull request. I was thinking we could test if the Pareto Front from the Pymoo implementation is the same as ours, see if the outputs match? Your thoughts @zoq ?

Isn't that what GetParetoFront returns?

jonpsy:
Good question, ```GetParetoFront``` (from the optimization algorithm) returns the solution set that we analyzed after doing everything. Whereas the Pareto Front here is the "Ideal/reference" Pareto Front, it's kinda like the "expected value". 

More importantly, you've brought up a valid concern, the name of these two shouldn't be same to avoid confusion. I guess I'll name this to "GetTrueParetoFront/GetIdealParetoFront/GetReferenceFront", whats your pick :) ?

zoq:
`GetReferenceFront` sounds like a good name to me, but if you prefer another name, please feel free to go with that.

zoq:
> Updates on this: pymoo has implemented the Pareto front for each of these so I have included the Ipynb notebook in this pull request. I was thinking we could test if the Pareto Front from the Pymoo implementation is the same as ours, see if the outputs match? Your thoughts @zoq ?

What I don't like about the idea is that we don't necessary test if the implementation is correct but if our implementation matches with Pymoo, on the other side, it's a good start, if they don't match we might want to check if Pymoo introduces an issue, just to make sure we don't make the same mistake.

jonpsy:
I believe we've clarified this in our video session. There's no concrete formula for generating the pareto front, so we're using our own mathematical equation which sufficiently generates the front. In that regard, pymoo does a good job in it and is why we're using it as reference.

I'll update the ipynb notebook and paste the theoretical pareto front besides the pymoo generated Pareto Front.

zoq:
> I believe we've clarified this in our video session. There's no concrete formula for generating the pareto front, so we're using our own mathematical equation which sufficiently represents the front. In that regard, pymoo does a good job in it and is why we're using it as reference.
> 
> I'll update the ipynb notebook and paste the theoretical pareto front besides the pymoo generated Pareto Front.

Agreed, as mentioned in the meeting as long as we are clear in the comments it's good.

zoq:
Looking at the Travis output, this is an easy issue to fix.

jonpsy:
Indeed. Currently, two things bother me:

a) How do we evaluate the performance on ZDT? 

In this regard, the paper has defined the g objective and its optimal value on any given coordinates. We can use ```coords``` and evaluate it with ```g``` such that it is approximately equal to ```g*```. 

For instance, look at this [code](https://pastebin.com/JECtRwhz), this test passed when asserted 3000 times with random seed. We will have to create an API to expose the ```g``` function instead of manually writing, for eg.

```c++
 ...
 opt.Optimize(objectives, coords);
 double gValue = ZDT_ONE.GetGValue();
 double expectedGValue = ZDT_ONE.GetOptimalGValue();

 REQUIRE(gValue == Approx(expectedGValue).margin(0.99));
```

b) ZDT5

This has been bothering me for a while now, the paper, pymoo and pagmo all have very different interpretations on the bitstring thingy and its confusing me a lot. Still working on it.

jonpsy:
Very well, I'll comment that the g objective function is taken from the ZDT1 implementation and to refer the docs for optimal g value.  Reg. the CI-timeout, if you recall in NSGA-II improvisation we were able to bring down the numGenerations from 1500 =>300 to achieve similar results, which should've loaned us a huge amount of CI-time. However, I agree with few examples should be enough, so I'll only test against ZDT_ONE as per the code I pasted before, do we agree?

zoq:
> Very well, I'll comment that the g objective function is taken from the ZDT1 implementation and to refer the docs for optimal g value. Reg. the CI-timeout, if you recall in NSGA-II improvisation we were able to bring down the numGenerations from 1500 =>300 to achieve similar results, which should've loaned us a huge amount of CI-time. However, I agree with few examples should be enough, so I'll only test against ZDT_ONE as per the code I pasted before, do we agree?

Sounds like a good plan to me.

jonpsy:
Great, only ZDT5 is left. It's a lot confusing  if you could weigh in that'd be awesome.

zoq:
> Great, only ZDT5 is left. It's a lot confusing if you could weigh in that'd be awesome.

I looked at the pymoo implementation (https://github.com/msu-coinlab/pymoo/blob/master/pymoo/problems/multi/zdt.py#L99) which looks like a reasonable approach to me, also I got the impression that a lot of frameworks just don't implement the function. Anyway, I would just go with the pymoo interpretation, let me know if you like me to look into something specific.

jonpsy:
I didn't quite understand what they're doing. To begin with, ```x_1``` should be a bitstring of length 30 and x2 .... x11 should be a bitstring of length 5. I'm not sure how they're ensuring that they're a bitstring or are they converting it in some way? Also the paper was clear that there are 11 dimensions, but this implementation somehow calculates ```numDimensions``` in a very obscure way.

To make things worse, here's pagmos [implementation](https://pastebin.com/7nVE5nuw), especially two parts

a)
```c++
    // Convert the input vector into rounded values (integers)
    vector_double x;
    std::transform(x_double.begin(), x_double.end(), std::back_inserter(x),
                   [](double item) { return std::round(item); });
```

So basically, if coordinates are ```0.03``` it becomes ```0```. But how could they guarantee coordinates would be within 0-1?

b)
  ```c++
   // Counts how many 1s are there in the first (30 dim)
    u[0] = static_cast<vector_double::size_type>(std::count(x.begin(), x.begin() + 30, 1.));
```
Not getting what the "first 30" dim is supposed to mean.


And here's the excerpt from the original paper.

ZDT-T5

The paper is guaranteeing that the num dimension is 11 but neither of the two implementation seem to follow it. So weird.

**Comment by zoq** [comment]
Thanks for the extra insights, will dig into the two implementations and the paper, and will let you know what I think.

**Comment by jonpsy** [comment]
As discussed, let's skip ZDT5 as of now.

**Comment by zoq** [comment]
> As discussed, let's skip ZDT5 as of now. Let me know if you find any other changes peculiar  @zoq

Yes, let us fix the build before we merge this :+1:

**Comment by jonpsy** [comment]
Allright, this should work smooth like a butter.

**Comment by jonpsy** [comment]
I think this is ready.

File: tests/nsga2_test.cpp [hunk 2]
```diff
@@ -412,4 +412,35 @@ TEST_CASE("NSGA2FonsecaFlemingTestVectorFloatBounds", "[NSGA2Test]")
   }
 
   REQUIRE(allInRange);
+}
+
+TEST_CASE("NSGA2ZDTONETest", "[NSGA2Test]")
```
zoq, line 417: Can you add a simple description for the test on top?

zoq: Also, do you mind to add a comment, that we also tested the other ZDT functions, but due to runtime constraints left didn't include them in the test suite?

jonpsy: Done, do you mind confirming ?

zoq (approved):
Looks great to me, no more comments from my side.

tests/nsga2_test.cpp
zoq: Thanks, looks perfect.

zoq:
Thanks again, great work :+1:

PR #274, opened 2021-03-18
SchafferN1 Initial Point fix

jonpsy:
Deceptively innocent but causes compile-time error when ```arma::fmat``` is used to collect initial point. Can we have a quick merge on this, please? @zoq

### DIFF
--- a/include/ensmallen_bits/problems/schaffer_function_n1.hpp
+++ b/include/ensmallen_bits/problems/schaffer_function_n1.hpp
@@ -63,7 +63,10 @@ class SchafferFunctionN1
   //! Get the starting point.
   MatType GetInitialPoint()
   {
-    return arma::vec(numVariables, 1, arma::fill::zeros);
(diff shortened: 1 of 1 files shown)


rcurtin (approved):
Ah, nice catch. :+1:

zoq (commented):
Can you update the `HISTORY.md` file as well?

jonpsy:
Sure, for the next step, should I add float tests for NSGA-II?

rcurtin:
At least personally I'd be willing to ignore the history check for this one, since this is basically test code only.  But @jonpsy if you wanted to add float tests for NSGA-II in this PR, we should add a history entry for that. :+1:

jonpsy:
I'm planning to do that in separate PR, @rcurtin  I agree, this is a minor change, can we ignore HISTORY.md pls? :)

zoq (approved):
Fine with me.

PR #275, opened 2021-03-19
InitPoint

jonpsy:
Just checked the init point return state for all functions inside ```problems/``` module. Found another culprit.


zoq (approved):
Thanks, no comments from my side.

rcurtin (approved):
Awesome, thanks for the fix. :+1:

Issue #272, opened 2021-03-16
Bounds of multiobjective problems should be handled automatically. 


jonpsy:
Speaking in the context of multiobjective algorithms, when defining an objective function test case, it's important to also define the variable space in which it applies.  That means, its important to define the upper and lowerbound **inside the problem**. 

The bounds are an integral part of the problem itself, and users shouldn't be bothered on the bounds to be set in the first place.

Few examples:

**a) Sphere function**
![sphere](https://user-images.githubusercontent.com/45007169/111261352-5d89a080-8648-11eb-925d-677167c06498.png)

Bounds of variable space : [0, 10]

**Violations:**
  
- **PSO Test**: Nearly all of the PSO tests(which use Spherical function), don't utilize the lower and upper bounds of the function. It simply is an unconstrained problem.

**b) Schaffer function**
![schaffern1](https://user-images.githubusercontent.com/45007169/111261917-3bdce900-8649-11eb-813b-649a7651dbab.png)

**Violations**
- **NSGA2 Test**: NSGA2Test sets the bounds customly to ```[-1000, 1000]``` , which again brings me back to the main point, users **shouldn't be bothered to fill the bounds of a problem**.

jonpsy:
An added advantage to bounds being inside the problem is the optimizer doesn't require ```lowerBound``` and ```upperBound``` argument anymore. It can simply extract it from the problem itself.

zoq:
> NSGA2 Test: NSGA2Test sets the bounds customly to [-1000, 1000] , which again brings me back to the main point, users shouldn't be bothered to fill the bounds of a problem.

All the problems we have in ensmallen are there for testing, so I'm not sure putting the bounds in the problem definition saves a user from specifying bounds for his problem, the same applies for the optimizer parameters, which have to be tailored down to the specific problem. Maybe I missed the point here?

jonpsy:
Sorry, I wasn't clear. I meant, Schaffer N1 Problem which is used inside NSGA2 Test. In NSGA2Test we **manually** put lowerBound as -1000 and upperBound as 1000 whereas Schaffer N1 Problem's domain is [-10, 10].

zoq:
I see what you mean; there are two main problems I have with putting the bounds into the problem:

1. The current function/problem interface is super simple, in the simplest form, you just have to implement `Evaluate` or `EvaluateConstraints`, putting the bounds in the function adds two more functions to the interface.
[...]

PR #276, opened 2021-03-20
Clarifying comments about test problems

rcurtin:
In the discussion of #272 we realized it would be a good idea to add comments to each of the test problems that implement `GetInitialPoint()`, `GetFinalObjective()`, and `GetFinalPoint()`, in order to indicate that these are not necessary if you just want to optimize a function, and that these are only there just for the convenience of our testing infrastructure.

So, this PR just adds those comments, and moves those functions to the bottom of the class.


zoq (approved):
Looks great to me :+1:

Issue #282, opened 2021-04-03
Framework for Multi-Objective Optimizations - II


jonpsy:
This would be a long-standing discussion, the goal is to discuss the return types, testing and other functionalities of MOEA(Multiobejctive evolutionary algorithm)  and how it all can seamlessly fit with the established single-objective optimizer(SOO) API.

First, let's revise the goals of an MOEA:

-  To approximate the Pareto Front  (solution set) ---- (1)
-   To have a diverse set of solution       ---- (2)

**I. Optimize and iterate**
        
- In ```Optimize()``` we return the final objective for SOO, emulated in MOO by returning ```min arma::accu``` of the objective vector.
- ```iterate``` is the optimal variable in SOO, and the first element of ```bestFront``` in MOO. Ideally, ```iterate``` should have been a vector of points because there are multiple optimal variables .

Clearly, it seems we haven't incorporated the spirit of MOEA to its full extent. A quick fix would be:

   => ```Optimize()``` should return the **variance** of the Pareto Front, this checks off (2) since it gives an idea of "diversity"
   => ```iterate``` should be set to the median of ```bestFront```.

**II.Type traits [DONE]**

We may require some additional type-traits for MOEA. For ex: ```CheckArbitrartyFunctionTypeAPI```, check if a policy function has ```Evaluate()```. We would need something like ```CheckMultiArbitraryFunctionTypeAPI```(or a shorter name),  to check for ```Evaluate()``` for each function policy using variadic template recursion.

**III. Test Metrics**

Building from our discussion from I) , we can get a scalar representation of how good our model is by using Metrics. ```Optimize``` can take a template argument for that, but that breaks the existing API.

Rather, we could use it  in ```test_function_tools.hpp```, such that ```objective``` is replaced by ```performance```, where performance is the retval of metric.

A helper function  like:
```c++ 
template <typename Metric, typename PointType>
 PointType::elem_type EvaluatePerformance(Metric metric, std::vector<PointType> solutionSet)
{  return metric.Evaluate(solutionSet); }
```
which will be used as:  #278  for more
```c++
template <typename FunctionSuiteType,
          typename MatType = arma::mat,
          typename OptimizerType = ens::NSGA2,
          typename Metric = ens::indicator::HyperVolumeIndicator>
void MultiObjectiveFunctionTest(...) 
{
   ElemType performance = EvaluatePerformance<Metric>(f.GetSolutionSet);
[...]

PR #283, opened 2021-04-03
Recursively check for Evaluate() in MOO

jonpsy:
Refer #282 Section II.

Is there a strong reason for having ```template<MatType, FunctionType>``` instead of ```template<FunctionType, MatType>```, the problem is when we do recursive call, I need the FunctionType to be the at the end for it to work. I don't see any other way.

Infact if we go by ```template<MatType, FunctionType>```, I can integrate ```CheckMultiArbitraryFunctionTypeAPI()``` into ```CheckArbitraryFunctionTypeAPI()```, and it will work like normal + MOO checks.

### DIFF
--- a/include/ensmallen_bits/function/static_checks.hpp
+++ b/include/ensmallen_bits/function/static_checks.hpp
@@ -375,7 +375,7 @@ inline void CheckSparseFunctionTypeAPI()
 /**
  * Perform checks for the ArbitraryFunctionType API.
  */
-template<typename FunctionType, typename MatType, std::size_t I = 0U>
+template<typename FunctionType, typename MatType, std::size_t I = 0>
 inline void CheckArbitraryFunctionTypeAPI()
 {
 #ifndef ENS_DISABLE_TYPE_CHECKS
@@ -387,8 +387,8 @@ inline void CheckArbitraryFunctionTypeAPI()
 #endif
 }
 
-template<typename... FunctionAndMatTypes, std::size_t I = 0U>
-typename std::enable_if<I < sizeof...(FunctionAndMatTypes) - 2U, void>::type
+template<typename... FunctionAndMatTypes, std::size_t I = 0>
+typename std::enable_if<I < sizeof...(FunctionAndMatTypes) - 2, void>::type
 CheckArbitraryFunctionTypeAPI()
 {
 #ifndef ENS_DISABLE_TYPE_CHECKS
--- a/include/ensmallen_bits/function/static_checks.hpp
+++ b/include/ensmallen_bits/function/static_checks.hpp
@@ -375,7 +375,7 @@ inline void CheckSparseFunctionTypeAPI()
 /**
  * Perform checks for the ArbitraryFunctionType API.
  */
-template<typename FunctionType, typename MatType, std::size_t I = 0>
+template<typename FunctionType, typename MatType>
 inline void CheckArbitraryFunctionTypeAPI()
 {
 #ifndef ENS_DISABLE_TYPE_CHECKS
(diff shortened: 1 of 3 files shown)


include/ensmallen_bits/function/static_checks.hpp
```diff
@@ -387,6 +387,40 @@ inline void CheckArbitraryFunctionTypeAPI()
...
+* Multiobjective case.
+*/
+template <typename MatType,
+          typename FunctionType,
+          std::size_t I = 1U>
+typename std::enable_if<I == 1U, void>::type
+CheckMultiArbitraryFunctionTypeAPI()
+{
+  CheckArbitraryFunctionTypeAPI<FunctionType, MatType>();
+}
```
jonpsy, line 401: I wouldn't need to do switch the order and then feed to ```CheckArbitraryFunctionTypeAPI``` , if not for this.

jonpsy:
I think template specialization is the only way out of it. Thoughts? @zoq

zoq:
I don't mind the solution, I think it's fine; it would also be fine for me to change the order, it's an internal function so we will not break anything. @rcurtin what do you think?

jonpsy:
Or, we could skip this entire thing and write
```c++

 for(....)
{
    using FunctionType = typename std::tuple_element<
        I, typename std::tuple<ArbitraryFunctionType...>>::type;
    traits::CheckArbitraryFunctionTypeAPI<
        FunctionType, MatType>();
....
}
````
Inside Evaluate objectives

**Comment by jonpsy** [comment]
* Everything happens at compile time
* Uses the exact same interface.
* Automagically handles a higher number of functions.


Perfect!

File: include/ensmallen_bits/function/static_checks.hpp [hunk 2]
```diff
@@ -375,7 +375,7 @@ inline void CheckSparseFunctionTypeAPI()
 /**
  * Perform checks for the ArbitraryFunctionType API.
  */
-template<typename FunctionType, typename MatType>
+template<typename FunctionType, typename MatType, std::size_t I = 0U>
```
zoq, line 378: Do you mind to use `0` instead of `0U` here and  you used that in a couple of other places as well, I know this is a preference, but ideally I like to keep the same style across the entire codebase.

include/ensmallen_bits/function/static_checks.hpp
```diff
@@ -387,6 +387,26 @@ inline void CheckArbitraryFunctionTypeAPI()
...
+{
+#ifndef ENS_DISABLE_TYPE_CHECKS
+  constexpr size_t size = sizeof...(FunctionAndMatTypes);
+  using TupleType = typename std::tuple<FunctionAndMatTypes...>;
+  using FunctionType = typename std::tuple_element<I, TupleType>::type;
+  using MatType = typename std::tuple_element<size - 1, TupleType>::type;
+
+  static_assert(CheckEvaluate<FunctionType, MatType, MatType>::value,
+      "One of the provided FunctionType does not have a correct definition of Evaluate(). "
+      "Please check that the corresponding FunctionType fully satisfies the requirements"
```
zoq, line 402:
```suggestion
      "Please check that the corresponding FunctionType fully satisfies the requirements "
```

Missing space at the end of the line.

zoq:

jonpsy:
Trying this on NSGA2 is giving me some errors, I'll fix it and ping you back.

zoq (commented):
Would be nice if you could add a MMO function to:

https://github.com/mlpack/ensmallen/blob/master/tests/function_test.cpp

zoq:
> Would be nice if you could add a MMO function to:
> 
> https://github.com/mlpack/ensmallen/blob/master/tests/function_test.cpp

Nevermind, it's another test suite.

zoq (approved):
Looks great to me, thanks for putting this together :+1:

zoq:

jonpsy:
@zoq  I suppose I need to rebase ?

zoq:
> @zoq I suppose I need to rebase ?

No, I just like to see if the `mlpack master build test` works as expected.

jonpsy:
mlpack master build fails:

```bash
g.Component(0).Mean()[0] == Approx(0.5).epsilon(0.05)

FAILED:
  REQUIRE( g.Component(0).Mean()[0] == Approx(0.5).epsilon(0.05) )
with expansion:
  0.5299031563 == Approx( 0.5 )
at /home/jenkins/workspace/pull-requests-ensmallen-mlpack-build/mlpack/src/mlpack/tests/gmm_test.cpp:276
```

zoq:
> mlpack master build fails:
> 
> ```shell
> g.Component(0).Mean()[0] == Approx(0.5).epsilon(0.05)
> 
> FAILED:
>   REQUIRE( g.Component(0).Mean()[0] == Approx(0.5).epsilon(0.05) )
> with expansion:
>   0.5299031563 == Approx( 0.5 )
> at /home/jenkins/workspace/pull-requests-ensmallen-mlpack-build/mlpack/src/mlpack/tests/gmm_test.cpp:276
> ```
> 
> Is this expected?

Some mlpack tests are non-deterministic, so they can fail sometimes, in most cases a simple threshold adjustment is all what is needed.

zoq:
Thanks again for the contribution.

PR #284, opened 2021-04-07
Code analyzer adjustments

zoq:
Some simple adjustments to make the code analyzer happy.

### DIFF
--- a/include/ensmallen_bits/fw/proximal/proximal_impl.hpp
+++ b/include/ensmallen_bits/fw/proximal/proximal_impl.hpp
@@ -45,7 +45,7 @@ inline void Proximal::ProjectToL1Ball(MatType& v, double tau)
   MatType simplexSum = arma::cumsum(simplexSol);
 
   double nu = 0;
-  size_t rho = 0;
+  size_t rho = simplexSol.n_rows - 1;
   for (size_t j = 1; j <= simplexSol.n_rows; j++)
   {
     rho = simplexSol.n_rows - j;
--- a/include/ensmallen_bits/pso/update_policies/lbest_update.hpp
+++ b/include/ensmallen_bits/pso/update_policies/lbest_update.hpp
@@ -73,7 +73,8 @@ class LBestUpdate
      *
      * @param parent Instantiated parent class.
      */
-     Policy(const LBestUpdate& /* parent */) { /* Do nothing. */ }
+     Policy(const LBestUpdate& /* parent */) : n(0)
+     { /* Do nothing. */ }
 
      /**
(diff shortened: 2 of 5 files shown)


zoq:

zoq:
I'm not going to fix the NSGA warnings, because https://github.com/mlpack/ensmallen/pull/263 will fix most of the issues. Also once this is merged, I will remove the old static analysis job.

include/ensmallen_bits/fw/proximal/proximal_impl.hpp
```diff
@@ -45,7 +45,7 @@ inline void Proximal::ProjectToL1Ball(MatType& v, double tau)
   MatType simplexSum = arma::cumsum(simplexSol);
 
   double nu = 0;
-  size_t rho = 0;
+  size_t rho = simplexSol.n_rows - 1;
```
rcurtin, line 48: It looks also like the definition of `rho` could be pushed inside the loop.

rcurtin (approved):
Looks like CMake isn't found on polar and that's why the `mlpack master build test` job fails, but the changes all look fine to me. :+1:

include/ensmallen_bits/fw/proximal/proximal_impl.hpp
zoq: In this case we would have to calculate `theta` inside the loop as well, https://github.com/mlpack/ensmallen/blob/1cd90c6158d320f388b1ae6dd6cdfc6d15b632cb/include/ensmallen_bits/fw/proximal/proximal_impl.hpp#L56 maybe I'm missing something? I guess we could also ignore the warning because `rho` will always be `> 0` but I'I figured it might be easier to just initialize it once.

zoq:
> Looks like CMake isn't found on polar and that's why the `mlpack master build test` job fails, but the changes all look fine to me. +1

Yeah, I have the Dockerfile (https://github.com/mlpack/jenkins-conf/blob/master/Dockerfiles/jenkins-amd64-debian.Dockerfile) I just have to use it in the build config. But I figured I should ask first before I do the change.

jonpsy:
I have resolved the warnings for nsga2. Hope to see this merged soon :D !

zoq:

zoq:

zoq:

include/ensmallen_bits/fw/proximal/proximal_impl.hpp
rcurtin: You're right, how did I overlook that? 😄 Sorry about that.

zoq:

zoq:

PR #120, opened 2019-07-26
Framework for Multi-Objective Optimizations

favre49:
This PR will add a framework for multi-objective optimization. This will include the DTLZ problems, the NSGA-III optimizer and some indicators (hypervolume, generational distance, epsilon).

PR #285, opened 2021-04-11
Add MOO Indicators

jonpsy:
Ported ```igd.hpp``` and ```epsilon.hpp```. #120 

LOGS:
- [x] Move the code to utility/indicators/*
- [x] Register at ensmallen.hpp
- [x] Change ```bestFront``` type to ```arma::cube``` in ```nsga2.hpp```, ```nsga2_impl.hpp``` and ```nsga2_test.cpp```.
- [x] NSGA2 test passes.
- [x]  Optimize the code.
- [x]  Write a sample test for both float/double for each indicator and compare with a hand calculated value. 
- [x]  Minor doc fixes.
- [x]  Expose the Pareto frontier (in objective space) through a public interface.

Test IPythonNotebook: https://colab.research.google.com/drive/1pFvoeIdqo8FQgEw5nOx1ANXH8AvX9eR3?usp=sharing


include/ensmallen_bits/indicators/epsilon.hpp
```diff
@@ -0,0 +1,79 @@
...
+  {
+    /**
+     * Find the epsilon value of the front with respect to the given reference
+     * front.
+     *
+     * @param front The given approximation front.
+     * @param referenceFront The given reference front.
+     * @return The epsilon value of the front.
+     */
+    double Indicate(arma::cube& front,
```
jonpsy, line 56: Is Indicate really the word we're going for? I'd suggest EvaluateFront / Evaluate or something like that. @zoq Your thoughts?

zoq: `Evaluate` sounds good, also do you mind to put this into an existing optimizer, this is what we did for policies as well, and I'd like to keep the root folder clean.

include/ensmallen_bits/indicators/igd.hpp
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Find the IGD value of the front with respect to the given reference
+     * front.
+     *
+     * @param front The given approximation front.
+     * @param referenceFront The given reference front.
+     * @return The IGD value of the front.
+     */
+    double Indicate(arma::cube& front,
+      arma::cube& referenceFront)
```
zoq, line 53: Looks like this would fit in one line.

include/ensmallen_bits/indicators/igd.hpp
```diff
@@ -0,0 +1,82 @@
...
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+#ifndef ENSMALLEN_INDICATORS_IGD_HPP
+#define ENSMALLEN_INDICATORS_IGD_HPP
+
+namespace ens {
+
+  /**
```
zoq, line 20: I think we can remove one space here.

include/ensmallen_bits/indicators/epsilon.hpp
```diff
@@ -0,0 +1,79 @@
...
+   *    author   = {E. Zitzler and L. Thiele and M. Laumanns and C. M. Fonseca and
+   *                V. G. da Fonseca},
+   *    title    = {Performance assessment of multiobjective optimizers: an
+   *                analysis and review},
+   *    journal  = {IEEE Transactions on Evolutionary Computation},
+   *    year     = {2003},
+   * }
+   * @endcode
+   */
+  class Epsilon
```
zoq, line 46: I think we can remove one space here.

zoq (commented):
We should add a simple test that e.g. compares the indicator with some manually calculated values, what do you think?

jonpsy:
Sure.Another thing, after reading his code. I think it's an excellent idea to have front has ```arma::cube`` instead of ```std::vector<arma::mat>```. 

Shall I make changes in nsga2 for bestFront in this PR?

include/ensmallen_bits/indicators/epsilon.hpp
jonpsy: I suppose you mean I should put this into optimizers similar to how ```sgd``` has ```decay_policies```? If I'm correct till here, I believe that's the wrong way to go. An indicator is essentially a metric, so I was following the conventions as is provided [here](https://www.mlpack.org/doc/stable/doxygen/lmetric_8hpp_source.html), notice ```lmetric.hpp``` is inside ```metric/```, I'm going for something similar :D.

jonpsy: > `Evaluate` sounds good, also do you mind to put this into an existing optimizer, this is what we did for policies as well, and I'd like to keep the root folder clean.

This is actually genius! So, maybe we could do 
```c++
<MatType, ...ArbitraryFunctionType, IndicatorType = ens::HyperVolume<MatType>> 
typename MatType::elem_type Optimize()
{
/*code*/
return IndicatorType.Evaluate(bestFront);
}
```
This way Optimize would return a single value as per our choice of metric! Perfect!
Then we could have ```indicator_policy/``` for nsga2 folder and others. But I would need to implement the same indicator code for each and every MOO optimizer, so that's the bad part. Am I following what you said correctly?

jonpsy: Oh wait, I would need  reference front .... this won't work :(

jonpsy: Okay, as discussed, I will move this to utility/indicators and make it something similar to lmetric.hpp.

zoq:
> Sure.Another thing, after reading his code. I think it's an excellent idea to have front as `arma::cube` instead of `std::vector<arma::mat>`.
> 
> Shall I make changes in nsga2 for bestFront in this PR?

Yes, good idea, let's make the change part the this PR.

jonpsy:
I have a feeling the implementation or the docs may be wrong after reading [this](https://pymoo.org/misc/performance_indicator.html#nb-igd-plus). All that's left is minor doc fixes, confirming/optimizing code and final test. I might not be able to give this time this month due to my exam, so I'll update the TO-DO for the tasks left in case somebody picks it up before I get back to it. @zoq  Could you review the work done till now, I will pick it up when I'm back, thanks! :)

include/ensmallen_bits/utility/indicators/epsilon.hpp
```diff
@@ -0,0 +1,83 @@
...
+#define ENSMALLEN_INDICATORS_EPSILON_HPP
+
+namespace ens {
+
+/**
+ * The epsilon indicator is a binary quality indicator that was proposed by
+ * Zitzler et. al. in response to the observation that quality indicators could
+ * only give information that one approximation set is not worse than another.
+ * The binary epsilon indicator, however, is capable of detecting if an
+ * approximation is better than another.
```
jonpsy, line 26: This description is too generic as all the indicators proposed in the paper satisfy this property.

include/ensmallen_bits/utility/indicators/igd.hpp
```diff
@@ -0,0 +1,86 @@
...
+      {
+        ElemType min = std::numeric_limits<ElemType>::max();
+        for (size_t j = 0; j < front.n_slices; j++)
+        {
+          ElemType dist = 0;
+          for (size_t k = 0; k < front.slice(j).n_rows; k++)
+          {
+            ElemType z = referenceFront(k, 0, i);
+            ElemType a = front(k, 0, j);
+            dist += std::pow(std::max(z - a, 0.0), 2);
```
jonpsy, line 70: It's worth noting that, for maximization of objectives problem it's ```z-a``` but for minimization, it's ```a-z```. This is the error I was talking about before.

jonpsy:
@zoq  I think this is good to go.

jonpsy:
The failing MSVC15 probably has to do with ```ElemType{0.0}``` I've done, though I have to confirm it with RDP session.

jonpsy:
Interface for Pareto optimal set in variable space : ```ParetoSet```
Interface for Pareto optimal set in objective space: ```ParetoFront``` (in place of Front)

include/ensmallen_bits/nsga2/nsga2.hpp
```diff
@@ -168,9 +169,13 @@ class NSGA2 {
   //! Modify value of upperBound.
   arma::vec& UpperBound() { return upperBound; }
 
-  //! Retrieve the best front (the Pareto frontier).  This returns an empty vector until `Optimize()`
-  //! has been called.
-  const std::vector<arma::mat>& Front() const { return bestFront; }
+  //! Retreive the pareto optimal points in variable space. This returns an empty cube
```
zoq, line 172:
```suggestion
  //! Retrieve the Pareto optimal points in variable space. This returns an empty cube
```

Minor spelling fix and captial Pareto just to be consistent.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -203,22 +204,25 @@ typename MatType::elem_type NSGA2::Optimize(
...
-  iterate = front[0];
+  // Calculate the objectives of the surviving population.
+  EvaluateObjectives(population, objectives, calculatedObjectives);
+  // Set the candidates from the Pareto Front as the output.
+  paretoFront.resize(calculatedObjectives[0].n_rows, calculatedObjectives[0].n_cols,
+      fronts[0].size());
+  // The Pareto Front is stored, can be obtained via ParetoFront() getter.
+  for (size_t solutionIdx = 0; solutionIdx < fronts[0].size(); ++solutionIdx)
+    paretoFront.slice(solutionIdx) =
+        arma::conv_to<arma::mat>::from(calculatedObjectives[fronts[0][solutionIdx]]);
```
zoq, line 222: Can you use `{}` technically it's a single line, but I think with `{}` it's easier to read.

include/ensmallen_bits/utility/indicators/epsilon.hpp
```diff
@@ -0,0 +1,92 @@
...
+      // Convenience typedefs.
+      typedef typename CubeType::elem_type ElemType;
+      ElemType eps = 0;
+      for (size_t i = 0; i < referenceFront.n_slices; i++)
+      {
+        ElemType epsjMin = std::numeric_limits<ElemType>::max();
+        for (size_t j = 0; j < front.n_slices; j++)
+        {
+          arma::Mat<ElemType> frontRatio = front.slice(j) / referenceFront.slice(i);
+          frontRatio.replace(arma::datum::inf, -1.); // Zero division case
```
zoq, line 77:
```suggestion
          frontRatio.replace(arma::datum::inf, -1.); // Handle zero division case.
```

jonpsy:
Thanks for the reviews, I've resolved it, let me know if you have any further queries. @say4n Would you like to weigh in here? Would appreciate your reviews as well, especially on the ```ParetoFront, ParetoSet``` interface part.

zoq (approved):
Looks good to me, no further comments from my side.

say4n:
Hi, @jonpsy it LGTM! Good job!! 👍

zoq:
Thanks for working on this.

jonpsy:
Related to #282.

PR #286, opened 2021-04-11
Lmcma

onikolskyy:
Hey @rcurtin,
here is the start of implementation for lmcma.
(Refer to: https://arxiv.org/abs/1511.00221)
There are still some more TODOs with respect to the original algorithm:

- [ ] Ensure correct templating
- [ ] Add the option to evaluate separable functions, which map to multi-dim spaces
- [ ] Add mirror sampling and possibly change the sampling method from Normal distribution to Radermacher R.V. to make the sampling procedure cheaper
- [ ] Eliminate the unnecessary copying of objects
- [ ] Add tests
- [ ] I have also that stupid `multiple definition of `ens::LMCMA::LMCMA(unsigned long)` while linking, when buildings tests. I don't get the reason, I guess, some support would be good here.


include/ensmallen_bits/lmcma/lmcma.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+
+namespace ens {
+
+
+class LMCMA
+{
+  public:
+
+    LMCMA(std::size_t N_dim);
```
jonpsy, line 16: numDimensions, variable names should be as descriptive as possible.

include/ensmallen_bits/lmcma/lmcma.hpp
```diff
@@ -0,0 +1,83 @@
...
+                                         CallbackTypes&&... callbacks);
+  private:
+
+    template <typename BaseMatType>
+    size_t update(std::size_t t,
+                         const BaseMatType& p,
+                         BaseMatType& P,
+                         BaseMatType& V,
+                         arma::umat& L,
+                         arma::umat& J);
```
jonpsy, line 34: Spacing issue

include/ensmallen_bits/lmcma/lmcma.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+
+      template <typename BaseMatType>
+      void reconstruct(const BaseMatType& P,
+                       const BaseMatType& V,
+                       const arma::umat& J,    /* TODO: why umat? */
+                       const std::size_t n_updates,
+                       BaseMatType& z);
```
jonpsy, line 42: ```BaseMatType``` is only used to convert Row/Col ==> Mat type inside our ```Optimize()``` function. For others, the convention is to just write ```MatType```. See for ex: [cne.hpp](https://github.com/mlpack/ensmallen/blob/1b6c1d93f09adc2313c67827355e6f9c9f958d3c/include/ensmallen_bits/cne/cne.hpp#L152). 

Also, function names should start with capital followed by small case letters.

```suggestion
      template <typename MatType>
      void Reconstruct(const MatType& P,
                       const MatType& V,
                       const arma::umat& J,    /* TODO: why umat? */
                       const std::size_t n_updates,
                       MatType& z);
```

include/ensmallen_bits/lmcma/lmcma.hpp
```diff
@@ -0,0 +1,83 @@
...
+      float c1;
+      float z_bias;
+      float c_sigma;  // conv combination factor for computing stepsize
+      float d_sigma;  // scale combination factor for computing stepsize
+      arma::mat w;        // weights for computation of one mean-step
+      float mu_w;         //
+      std::size_t mu;     // number of best species
+      std::size_t m;      // number of step vectors stored
+
+      std::size_t N;      // number of iterations between step vectors which are saved
```
jonpsy, line 76: See comment related to variable names above.

include/ensmallen_bits/lmcma/lmcma_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+            typename MatType,
+            typename... CallbackTypes>
+    typename MatType::elem_type LMCMA::optimize(SeparableFunctionType& f,
+                                         MatType& z,
+                                         float sigma,         // TODO: remove from here
+                                         std::size_t n_iter,  // TODO: Remove from here
+                                         CallbackTypes&&... callbacks)
+    {
+
+
```
jonpsy, line 49: Unnecesary spaces

include/ensmallen_bits/lmcma/lmcma_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+      // Convenience typedefs.
+      typedef typename MatType::elem_type ElemType;
+      typedef typename MatTypeTraits<MatType>::BaseMatType BaseMatType;
+
+      // Make sure that we have the methods that we need.  Long name...
+      traits::CheckArbitrarySeparableFunctionTypeAPI<
+              SeparableFunctionType, BaseMatType>();
+      RequireDenseFloatingPointType<BaseMatType>();
+
+      BaseMatType& iterate = (BaseMatType&) z;
```
jonpsy, line 59: Can we instead use ```iterateIn``` instead of z? Unless it has a symbolic meaning in the actual paper. See [cne_impl.hpp](https://github.com/mlpack/ensmallen/blob/1b6c1d93f09adc2313c67827355e6f9c9f958d3c/include/ensmallen_bits/cne/cne_impl.hpp#L89)

include/ensmallen_bits/lmcma/lmcma.hpp
jonpsy: Same for other functions.

include/ensmallen_bits/lmcma/lmcma_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+        ranks_mixed[idx[i]] = i;
+      }
+
+      float mean_prev = 0;
+      float mean_cur = 0;
+
+      for(size_t i = 0; i < lambda; i++)
+      {
+        mean_cur+= ranks_mixed[i];
+        mean_prev+= ranks_mixed[lambda + i];
```
jonpsy, line 287:
```suggestion
        mean_cur += ranks_mixed[i];
        mean_prev += ranks_mixed[lambda + i];
```

Nit

jonpsy:
Hey @onikolskyy , overall great work! I've left minor comments relating to styling issues. Hope it helps!

include/ensmallen_bits/lmcma/lmcma_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+          f_eval[k] = f.Evaluate(z);
+          X.col(k) = z;
+        }
+
+        // recombination
+        index_old = arma::uvec(index);
+        index = arma::sort_index(f_eval);
+
+        m_old = BaseMatType(m_new);
+        m_new = BaseMatType(N_dim,1, arma::fill::zeros);                // TODO: possible without reinintialization - would save some time
```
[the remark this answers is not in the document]
rcurtin: You might even be able to turn the entire loop into one expression---I haven't played with it though, so I'm not 100% sure.

include/ensmallen_bits/lmcma/lmcma_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+        sum+= log(i+1);
+
+      for(size_t i = 0; i < mu; i++)
+      {
+        float phi = std::log(mu+1) - std::log(i+1);
+        w[i] = (std::log(mu+1) - std::log(i+1) ) / (mu * log(mu + 1) - sum);
+      }
+
+      // const to recompute covariance steps
+      mu_w = 1 / std::pow(arma::norm(w,2), 2);
```
rcurtin, line 89: Just a style comment that applies throughout... we don't use `_` in variable names, instead preferring camelcase.  So, e.g., this would be `muW`.  See here for more information: https://github.com/mlpack/mlpack/wiki/DesignGuidelines#style-guidelines.  If you don't mind going through there and adjusting your code accordingly, I'd really appreciate it. :)

tests/CMakeLists.txt
```diff
@@ -44,7 +45,7 @@ set(ENSMALLEN_TESTS_SOURCES
     svrg_test.cpp
     swats_test.cpp
     wn_grad_test.cpp
-)
+    )
```
rcurtin, line 48: No need to indent here. :)
```suggestion
)
```

tests/lmcma_test.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include "catch.hpp"
+#include "test_function_tools.hpp"
+
+using namespace ens;
+using namespace ens::test;
+
+TEST_CASE("LMCMAInitTest", "[LMCMATest]")
+{
+  LMCMA c(1);
+}
```
rcurtin, line 15: I imagine you aren't quite done with the test suite yet, but if you could perhaps replicate some of the other tests used for arbitrary optimizers, I think that would be sufficient here. :+1:

rcurtin (commented):
Hey @onikolskyy, thanks for contributing this!  I haven't taken a close pass and haven't checked any of the LM-CMA implementation in detail, but hopefully these high-level comments are helpful. :+1:

jonpsy:
Hey @onikolskyy glad to see you're back. If you're interested, I could read the research paper and we can work on this together. It'll be done in a flash :).

Let me know what you think.

onikolskyy:
hi @jonpsy glad to hear from you.
If you are motivated, than lets get this ready!

jonpsy:
Great, you can start with addressing the requested changes by Ryan and me. It will take me some weeks to dive into the paper to give an insightful remark. If you have a reference python implementation that'll be great.

onikolskyy:
Hi @jonpsy  here's the paper https://arxiv.org/pdf/1511.00221.pdf

I did not find a python package, but there is c-code (there is a reference in the paper)

include/ensmallen_bits/lmcma/lmcma.hpp
onikolskyy: Thanks! fixed

include/ensmallen_bits/lmcma/lmcma.hpp
onikolskyy: Makes sence

include/ensmallen_bits/lmcma/lmcma_impl.hpp
```diff
@@ -0,0 +1,298 @@
...
+            //lambda(100),
+            w( 1,std::floor( (4 + (size_t)std::floor(3*std::log(N_dim))) / 2 ), arma::fill::zeros),
+            mu(std::floor( (4 + (size_t)std::floor(3 * std::log(N_dim))) / 2 ) )
+    {}
+
+
+    template<typename SeparableFunctionType,
+            typename MatType,
+            typename... CallbackTypes>
+    typename MatType::elem_type LMCMA::optimize(SeparableFunctionType& f,
```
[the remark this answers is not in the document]
onikolskyy: Right, I kept as near as possible to the current CMAES implementation, where a `SeparableFunction` is passed. 
But this should be an arbitrary function in both cases, I guess.

zoq (commented):
I liked the paper, happy to provide some more feedback, if you like to move forward with the implementation.

zoq:
Happy to finish the implementation as well.

PR #287, opened 2021-04-24
Switch from Ubuntu Trusty 14.04 to Ubuntu Focal 20.04

zoq:
For some reason travis marks the build as failed (I guess it's because it can't download some packages), let's see if switching to an more up-t-date ubuntu solves the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 os: linux
-dist: trusty
+dist: focal
 language: cpp
 
 env:


.travis.yml
```diff
@@ -1,5 +1,5 @@
 os: linux
-dist: trusty
+dist: xenial
```
birm, line 2: Isn't xenial 16.x?

zoq: You are right, let's see if this fixes the issue, if it does I fix the commit message and the title.

conradsnicta:
@zoq 16.04 is quite ancient at this point.  I suggest moving directly to 20.04, or 18.04 as an interim trade-off.

According the to Ubuntu [release page](https://wiki.ubuntu.com/Releases), 16.04 will be out of support at the end of April 2021 (ie. this month). On top of that, Xenial has a really old version of Armadillo (6.500), while ensmallen requires at least version 8.400.

Anecdotally, I don't think many people use 16.04 anymore, especially on the desktop. The vast majority of installations that I encounter are at 20.04.  So there's not much value testing against 16.04.

zoq:
Makes sense, updated to 20.04.

jonpsy:
I think this PR is a good upgrade. What's the latest here?

rcurtin (approved):
Awesome, looks good to me.  Also, I bumped the timeout on the mlpack master build.  (I have no idea why it took 180 minutes on dealgood to not even build everything, maybe network latency or contention.  Anyway, there's no lack of resources, so I just bumped it to 360 minutes.)

birm (approved):
Awesome, good to keep our tests up to date!

PR #288, opened 2021-05-02
Use Catch2 seed command-line argument to set ensmallen seeds

zoq:
Use Catch2 seed command-line argument to set ensmallen seeds; this enables us to improve the automatic test pipeline to run the tests multiple times with different random seeds.

### DIFF
--- a/tests/main.cpp
+++ b/tests/main.cpp
@@ -17,18 +17,23 @@
 
 int main(int argc, char** argv)
 {
-  /**
-   * Uncomment these three lines if you want to test with different random seeds
-   * each run.  This is good for ensuring that a test's tolerance is sufficient
-   * across many different runs.
-   */
-  //size_t seed = std::time(NULL);
-  //srand((unsigned int) seed);
-  //arma::arma_rng::set_seed(seed);
+  Catch::Session session;
+  const int returnCode = session.applyCommandLine(argc, argv);
+  // Check for a command line error.
+  if (returnCode != 0)
+    return returnCode;
 
   std::cout << "ensmallen version: " << ens::version::as_string() << std::endl;
-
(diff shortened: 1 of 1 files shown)


zoq:

conradsnicta:
good idea

tests/main.cpp
```diff
@@ -17,18 +17,23 @@
...
+  if (returnCode != 0)
+    return returnCode;
 
   std::cout << "ensmallen version: " << ens::version::as_string() << std::endl;
-
   std::cout << "armadillo version: " << arma::arma_version::as_string() << std::endl;
 
-  return Catch::Session().run(argc, argv);
+  // Use Catch2 command-line to set the random seed.
+  // -rng-seed <'time'|number>
```
rcurtin, line 30: It seems like by default, `-rng-seed` is set to `'time'`, but maybe it is a good thing to have a deterministic random seed?  I'm not sure quite what the way would be to force a default that's deterministic.  Here's the relevant bit from `catch.hpp`:

```
        auto const setRngSeed = [&]( std::string const& seed ) {
                if( seed != "time" )
                    return clara::detail::convertInto( seed, config.rngSeed );
                config.rngSeed = static_cast<unsigned int>( std::time(nullptr) );
                return ParserResult::ok( ParseResultType::Matched );
            };
```

zoq: It's the other way around, the default seed should be deterministic, at least it was deterministic on my system. Using `./ensmallen_tests --rng-seed=time` returns a seed for each run, but `./ensmallen_tests` returns the same seed.

rcurtin: Ohh, ok!  Sorry about that.  I misread the code.  Deterministic by default is definitely what we'd hope for. :+1:

rcurtin (approved):
Looks good, thanks for the explanation. :+1:

PR #289, opened 2021-05-28
Make Callback flexible for MultiObjective Optimizers.

jonpsy:
Related to #282  

- [x] Create ```MOOTypedForms``` struct specifically for MOO related type forms.
- [x] Create ```GenerationalStepTaken```, replacement for StepTaken in MOO.
- [x] Register the above using ENS MACRO.
- [x] Write in built callback for Querying Front.
- [x] Adapt the tests for this.

@zoq Let me know if you agree with this methodology. You can check the commit message to see what I've done.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -25,7 +25,7 @@
  * Add Indicators for Multiobjective optimizers
    ([#285](https://github.com/mlpack/ensmallen/pull/285)).
 
- * Make Callback traits adaptable for MOO
+ * Make Callback flexible for MultiObjective Optimizers
    ([#289](https://github.com/mlpack/ensmallen/pull/289)).
 
 ### ensmallen 2.16.1: "Severely Dented Can Of Polyurethane"
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -25,6 +25,9 @@
  * Add Indicators for Multiobjective optimizers
    ([#285](https://github.com/mlpack/ensmallen/pull/285)).
 
+ * Make Callback traits adaptable for MOO
+   ([#289](https://github.com/mlpack/ensmallen/pull/289)).
+
 ### ensmallen 2.16.1: "Severely Dented Can Of Polyurethane"
 ###### 2021-03-02
  * Fix test compilation issue when `ENS_USE_OPENMP` is set
--- a/include/ensmallen_bits/callbacks/callbacks.hpp
+++ b/include/ensmallen_bits/callbacks/callbacks.hpp
@@ -776,17 +776,6 @@ class Callback
       function, coordinates, objectives, frontIndices);
   }
 
- /**
-  * Invoke the StepTaken() callback if it exists.
-  * Specialization for MultiObjective case.
-  *
-  * @param callback The callback to call.
-  * @param optimizer The optimizer used to update the function.
-  * @param function Function to optimize.
-  * @param coordinates Starting point.
-  * @param objectives The set of calculated objectives so far.
-  * @param frontIndices The indices of the members belonging to Pareto Front.
-  */
   template<typename CallbackType,
            typename OptimizerType,
            typename FunctionType,
@@ -810,17 +799,6 @@ class Callback
     return false;
   }
 
- /**
-  * Invoke the StepTaken() callback if it exists.
-  * Specialization for MultiObjective case.
-  *
-  * @param callback The callback to call.
-  * @param optimizer The optimizer used to update the function.
-  * @param function Function to optimize.
-  * @param coordinates Starting point.
-  * @param objectives The set of calculated objectives so far.
-  * @param frontIndices The indices of the members belonging to Pareto Front.
-  */
   template<typename CallbackType,
            typename OptimizerType,
            typename FunctionType,
--- a/include/ensmallen_bits/callbacks/callbacks.hpp
+++ b/include/ensmallen_bits/callbacks/callbacks.hpp
@@ -752,28 +752,28 @@ class Callback
   * @param optimizer The optimizer used to update the function.
   * @param function Function to optimize.
   * @param coordinates Starting point.
-  * @param calculatedObjectives The set of calculated objectives so far.
+  * @param objectives The set of calculated objectives so far.
   * @param frontIndices The indices of the members belonging to Pareto Front.
   */
   template<typename CallbackType,
            typename OptimizerType,
            typename FunctionType,
            typename MatType,
-           typename ObjectivesArrayType,
+           typename ObjectivesVecType,
            typename IndicesType>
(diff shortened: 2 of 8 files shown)


include/ensmallen_bits/callbacks/callbacks.hpp
```diff
@@ -744,6 +744,101 @@ class Callback
...
+           typename IndicesType>
+    static typename std::enable_if<
+    callbacks::traits::HasMOOStepTakenSignature<
+    CallbackType, OptimizerType, FunctionType, MatType, ObjectivesArrayType,
+    IndicesType>::hasBool, bool>::type
+    StepTakenFunction(CallbackType& callback,
+      OptimizerType& optimizer,
+      FunctionType& function,
+      MatType& coordinates,
+      ObjectivesArrayType& calculatedObjectives,
```
zoq, line 772: I would call this `ObjectivesVecType` and `objectives`, everything we pass is calculated in some form so no need to add a prefix for that at the same time this is closer to the existing `StepTakenFunction`.

jonpsy: It is indeed, but i wanted to differentiate between then because it will cause very long codes and might be polluting. 

So, I created a separate MOOTypedForms specifically for MOO based works and then did ```hasBool``` and ```hasVoid``` in them. 

If I were to add this in TypedForms I would have to add

```hasMOObool``` and ```hasMOOVoid```. In my way, we get to not touch our old implementation at all, and add these functionalities.

**Comment by jonpsy** [comment]
@zoq  Similar to ```EndEpoch``` wouldn't it be great to have ```EndGeneration``` as well? Then we can make ```PrintLoss() ``` adapt to that. In fact, if we're smart about this  we can make almost all of our callbacks work for MOO cases.

**Comment by jonpsy** [comment]
@zoq  Similar to ```EndEpoch``` wouldn't it be great to have ```EndGeneration``` as well? Then we can make ```PrintLoss() ``` adapt to that. In fact, if we're smart about this  we can make almost all of our callbacks work for MOO cases.

**Comment by jonpsy** [comment]
The current implementation doesn't throw compile errors, but it doesn't work either. Looks like a template matching issue.

**Comment by jonpsy** [comment]
To test its working, I created this [test code](https://pastebin.com/pWBiTYBM). This is working completely fine given I have defined
StepTaken for MOO Specialization before the usual StepTaken.  That means, in callbacks.hpp

This should be defined first
```c++
template<typename OptimizerType,
         typename FunctionType,
         typename ObjectivesVecType,
         typename IndicesType,
         typename MatType,
         typename ...CallbackTypes>
  static bool StepTaken(OptimizerType& optimizer,
                        FunctionType& functions,
                        MatType& coordinates,
                        ObjectivesVecType& objectives,
                        IndicesType& frontIndices,
                        CallbackTypes&... callbacks)
```
and after that

```c++
  template<typename OptimizerType,
           typename FunctionType,
           typename MatType,
           typename... CallbackTypes>
static bool StepTaken(..)
```

So that the compiler avoids going into the usual StepTaken function. Let me know if this is fine or should I rename the first one as ```MOOStepTaken``` although I think the current implementation is good.

include/ensmallen_bits/callbacks/callbacks.hpp
zoq: Not sure I see the connection, I was referring to the naming of the parameters here.

zoq:
This looks fine to me.

zoq (commented):
Can you update https://github.com/mlpack/ensmallen/blob/master/tests/callbacks_test.cpp as well, the test you posted is a good basis.

jonpsy:
Gotta update the docs as well.

include/ensmallen_bits/callbacks/callbacks.hpp
jonpsy: I was clarifying as to why its similar to ```StepTakenFunction```, I've applied your suggestion on the name change already :+1:. Looks like an instance of miscommunication, haha

jonpsy:
GTM? @zoq

HISTORY.md
```diff
@@ -25,6 +25,9 @@
  * Add Indicators for Multiobjective optimizers
    ([#285](https://github.com/mlpack/ensmallen/pull/285)).
 
+ * Make Callback traits adaptable for MOO
```
zoq, line 28: Let use multi-objective optimization here, just to make sure everyone knows what MOO means.

include/ensmallen_bits/callbacks/query_front.hpp
```diff
@@ -0,0 +1,87 @@
...
+ */
+class QueryFront
+{
+ public:
+  /**
+   * Set up the query front callback class with the specified inputs.
+   *
+   * @param queryRate The frequency at which the Pareto Front is queried.
+   * @param paretoFrontArray A reference to a vector of cube to store the queried fronts.
+   * @param genCounter Internal clock to count generations passed.
```
zoq, line 28: Looks like I can't pass `genCounter`.

include/ensmallen_bits/callbacks/query_front.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+
+ private:
+  //! The rate of query.
+  size_t queryRate;
+  //! A reference to the array of pareto fronts.
+  std::vector<arma::cube>& paretoFrontArray;
+  //! A counter for the current generation.
+  size_t genCounter;
+
```
zoq, line 82:
```suggestion
```

Remove the extra line here.

include/ensmallen_bits/callbacks/query_front.hpp
jonpsy: Yes, this is an internal private counter meant to be in sync with ```generations++``` inside the main loop of optimization.

jonpsy: Think of it similar to ```steps``` in ```early_stop_at_min_loss.hpp```

jonpsy: Ah, as per convention, I should remove the doc from there. Sure!

zoq: Yes, let's remove it from the doc.

jonpsy:
Looks like this is ready

include/ensmallen_bits/callbacks/query_front.hpp
jonpsy: Already done 👍

zoq (approved):
Looks great to me, thanks for doing that.

zoq:
Thanks for another great addition.

PR #291, opened 2021-06-01
Doc fix

### LINKED ISSUE #290: Front() API
-->

#### Problem location

```doc/function_types.md```

#### Description of problem

The example is outdated, uses an older API. I'll do the change, but setting this issue as a reminder so that we don't sleep on it.

jonpsy:
#290


jonpsy:
@zoq  I dont think HISTORY.md is necessary here.

zoq (approved):
Thanks :+1:

zoq:
> @zoq I dont think HISTORY.md is necessary here.

No need, to update the file in this case.

birm (approved):

zoq:
Thanks for the quick fix.

PR #293, opened 2021-06-16
Introduce Policy Methods for MOEA/D-DE

jonpsy:
Continuation of #269 

**Weight decomposition**
- [x] Tchebycheff
- [x] Weighted
- [x]  PBI

**Weigh Init**

- [x] Bayesian Bootstrap

**Testing**
- [x] ZDT

@zoq  @coatless @say4n

PR #295, opened 2021-06-18
Add Das-Dennis weight initialization method

jonpsy:
Continuing #293 

- [x] Add uniform_init.hpp.
- [x] Use ```DefaultMOEAD = MOEAD<Uniform, Tchebycheff>```.
- [x]  Use ```BBSMOEAD = MOEAD<BayesianBootstrap, Tchebycheff>```
- [x] Test and verify its working locally.
- [x] Document uniform_init.hpp.
- [x] Add documentation in optimizers.md.
- [x] HISTORY.md.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -37,6 +37,8 @@
  * Introduce Policy Methods for MOEA/D-DE
    ([#293](https://github.com/mlpack/ensmallen/pull/293)).
 
+ * Add Das-Dennis weight initialization method
+   ([#295](https://github.com/mlpack/ensmallen/pull/295)).
 ### ensmallen 2.16.1: "Severely Dented Can Of Polyurethane"
 ###### 2021-03-02
  * Fix test compilation issue when `ENS_USE_OPENMP` is set
--- a/include/ensmallen_bits/moead/moead.hpp
+++ b/include/ensmallen_bits/moead/moead.hpp
@@ -327,6 +327,7 @@ class MOEAD {
 };
 
 using DefaultMOEAD = MOEAD<Uniform, Tchebycheff>;
+using BBSMOEAD = MOEAD<BayesianBootstrap, Tchebycheff>;
 } // namespace ens
 
 // Include implementation.
--- a/include/ensmallen_bits/moead/moead.hpp
+++ b/include/ensmallen_bits/moead/moead.hpp
@@ -22,6 +22,8 @@
 
 //! Weight initialization policies.
 #include "weight_init_policies/bbs_init.hpp"
+#include "weight_init_policies/uniform_init.hpp"
+
 namespace ens {
 
 /**
@@ -45,7 +47,7 @@ namespace ens {
  *   year={2008},
  * @endcode
  */
-template<typename InitPolicyType = BayesianBootstrap,
+template<typename InitPolicyType = Uniform,
          typename DecompPolicyType = Tchebycheff>
 class MOEAD {
  public:
@@ -72,8 +74,8 @@ class MOEAD {
    * @param upperBound The upper bound on each variable of a member
    *    of the variable space.
    */
-  MOEAD(const size_t populationSize = 150,
-        const size_t maxGenerations = 300,
+  MOEAD(const size_t populationSize = 300,
+        const size_t maxGenerations = 500,
         const double crossoverProb = 1.0,
         const double neighborProb = 0.9,
         const size_t neighborSize = 20,
@@ -111,8 +113,8 @@ class MOEAD {
    * @param upperBound The upper bound on each variable of a member
    *    of the variable space.
    */
-    MOEAD(const size_t populationSize = 150,
-          const size_t maxGenerations = 300,
+    MOEAD(const size_t populationSize = 300,
+          const size_t maxGenerations = 500,
           const double crossoverProb = 1.0,
           const double neighborProb = 0.9,
           const size_t neighborSize = 20,
@@ -324,7 +326,7 @@ class MOEAD {
   DecompPolicyType decompPolicy;
 };
 
-using DefaultMOEAD = MOEAD<BayesianBootstrap, Tchebycheff>;
+using DefaultMOEAD = MOEAD<Uniform, Tchebycheff>;
 } // namespace ens
 
 // Include implementation.
--- a/include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
+++ b/include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
@@ -14,16 +14,43 @@
 
 namespace ens {
 
+/**
+ * The Uniform (Das Dennis) method for initializing weights. This algorithm guarantees
+ * that the distance between adjacent points would be uniform.
+ *
+ * For more information, see the following:
+ * @code
+ * article{zhang2007moea,
+ *   title={MOEA/D: A multiobjective evolutionary algorithm based on decomposition},
+ *   author={Zhang, Qingfu and Li, Hui},
+ *   journal={IEEE Transactions on evolutionary computation},
+ *   pages={712--731},
+ *   year={2007}
+ * @endcode
+ */
 class Uniform
 {
  public:
+  /**
+   * Constructor for Uniform Weight Initializatoin Policy.
+   */
   Uniform()
   {
     /* Nothing to do. */
   }
 
+  /**
+   * Generate the reference direction matrix.
+   *
+   * @tparam MatType The type of the matrix used for constructing weights.
+   * @param numObjectives The dimensionality of objective space.
+   * @param numPoints The number of reference directions requested.
+   * @param epsilon Handle numerical stability after weight initialization.
+   */
   template<typename MatType>
-  MatType Generate(size_t numObjectives, size_t numPoints, double epsilon)
+  MatType Generate(size_t numObjectives,
+                   size_t numPoints,
+                   double epsilon)
   {
     size_t numPartitions  = FindNumParitions(numObjectives, numPoints);
     size_t validNumPoints = FindNumUniformPoints(numObjectives, numPartitions);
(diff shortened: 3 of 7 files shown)


include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
```diff
@@ -0,0 +1,159 @@
...
+    {
+      MatType point{};
+      size_t beta{};
+      std::tie(point, beta) = progressStack.back();
+      progressStack.pop_back();
+
+      if (point.size() + 1 == numObjectives)
+      {
+        point.insert_rows(point.n_rows,
+            RowType(1).fill(delta * static_cast<ElemType>(beta)));
```
jonpsy, line 111: Any cleaner way to insert an element into a vector?

include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * Finds the number of points which can be sampled from the unit 
+   * simplex given the number of partitions.
+   */
+  size_t FindNumUniformPoints(size_t numObjectives, size_t numPartitions)
+  {
+    auto BinomialCoefficient =
+        [](size_t n, size_t k) -> double
+        {
+          return std::tgamma(n + 1) / (std::tgamma(k + 1) * std::tgamma(n - k + 1));
```
jonpsy, line 58: I was wondering why there's an integer overflow happening here; then it hit me. ```std::tgamma(300)``` is shooting off to infinity. This is a very naive approach to finding Binomial Coefficient, I guess I could utilize the fact that the denominator and numerator cancel out a lot of terms. Though it'll certainly be nice if armadillo had a support for this algorithm built-in cc/ @conradsnicta

jonpsy:
The build failure seems unrelated to the PR. @zoq your insights would be helpful, thanks.

jonpsy:
I'm getting errors on ```adagrad_test.cpp``` for the changes I did in moead :). I'm done with cpp.

zoq:
> I'm getting errors on `adagrad_test.cpp` for the changes I did in moead class. I'm done with cpp  .

Huh, do you have a link to the Travis build, looks like it's down for me.

jonpsy:
Fixed it, I was missing a brace. C++ is so fun

include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
```diff
@@ -0,0 +1,206 @@
...
+          // [n * (n - 1) * .... * (n - k + 1)] / [k * (k - 1) * .... * 1].
+          for (size_t i = 0; i < k; ++i)
+          {
+            retval *= (n - i);
+            retval /= (i + 1);
+          }
+
+          return retval;
+        };
+    return BinomialCoefficient(numObjectives + numPartitions - 1, numPartitions);
```
zoq, line 100: Note to myself, check that part in more depth.

jonpsy: Add [this](https://www.geeksforgeeks.org/space-and-time-efficient-binomial-coefficient/)  to your notes.

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -72,8 +74,8 @@ class MOEAD {
    * @param upperBound The upper bound on each variable of a member
    *    of the variable space.
    */
-  MOEAD(const size_t populationSize = 150,
-        const size_t maxGenerations = 300,
+  MOEAD(const size_t populationSize = 300,
```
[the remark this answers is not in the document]
jonpsy: image

This is because the binomial coefficient value (H) should equate to the size of the population. This algorithm is restrictive in that way. You can read Section IV (If I recall correctly) of the cited paper to know more. To ensure this, you can find I've added an ```if``` statement inside ```Generate``` method to catch a mismatch. 

The main point is, there is only a certain number of points that can be generated uniformly given the number of partitions(gaps between points) and the number of objectives.

jonpsy:
Could you re-trigger the CI please? The build was cancelled for some reason. Thanks.

zoq:
> Could you re-trigger the CI please? The build was cancelled for some reason. Thanks.

Done.

jonpsy:
It got cancelled again (~.~) ...

include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
```diff
@@ -0,0 +1,206 @@
...
+    return DasDennis<MatType>(numObjectives, numPoints,
+        numPartitions, epsilon);
+  }
+
+ private:
+  /**
+   * Finds the number of points which can be sampled uniformly from a 
+   * unit simplex given the number of partitions.
+   */
+  size_t FindNumUniformPoints(size_t numObjectives, size_t numPartitions)
```
zoq, line 81:
```suggestion
  size_t FindNumUniformPoints(const size_t numObjectives,
                              const size_t numPartitions)
```

Just to trigger the CI.

jonpsy:
Added ```using BBSMOEAD = MOEAD<BayesianBootstrap, Tchebycheff>``` along with documentation in optimizers.md.

jonpsy:
@coatless  Would you like to have a pass over this? Thanks.

include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
```diff
@@ -0,0 +1,207 @@
...
+                   double epsilon)
+  {
+    size_t numPartitions  = FindNumParitions(numObjectives, numPoints);
+    size_t validNumPoints = FindNumUniformPoints(numObjectives, numPartitions);
+
+    //! The requested number of points is not matching any partition number.
+    if (numPoints != validNumPoints)
+    {
+      size_t nextValidNumPoints = FindNumUniformPoints(numObjectives, numPartitions + 1);
+      std::ostringstream oss;
```
coatless, line 63: Slight hiccup, we actually need to use one of ensmallen's output streams:

Perhaps, we should use: 

```
std::ostream& Info
```

https://github.com/mlpack/ensmallen/blob/0af8004ff989ed4cce3a2bcaf7eeabb1313ac913/include/ensmallen_bits/log.hpp#L73-L77

jonpsy: Thanks

include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
```diff
@@ -0,0 +1,207 @@
...
+      oss << "DasDennis::Generate(): " << "The requested numPoints " << numPoints
+          << " cannot be generated uniformly.\n " << "Either choose numPoints as "
+          << validNumPoints << "(numPartition = " << numPartitions << ") or "
+          << "numPoints as " << nextValidNumPoints << "(numPartition = "
+          << numPartitions + 1 << ").";
```
zoq, line 68:
```suggestion
          << " cannot be generated uniformly.\n " << "Either choose numPoints as "
          << validNumPoints << " (numPartition = " << numPartitions << ") or "
          << "numPoints as " << nextValidNumPoints << " (numPartition = "
          << numPartitions + 1 << ").";
```

Just one last minor style issue here as well.

include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
jonpsy: Okay so, I used  ```oss```  for throwing out error. Which should happen regardless of ```ENS_PRINT_INFO```.  So, once I put message inside ```oss```, I used ```std::logic_error(oss.str());```. Which I guess is the norm in our library.

zoq: Agreed, I don't think this needs special handling, since this is for throwing an exception instead of outputting something, @coatless correct me if I'm wrong.

HISTORY.md
```diff
@@ -37,6 +37,8 @@
  * Introduce Policy Methods for MOEA/D-DE
    ([#293](https://github.com/mlpack/ensmallen/pull/293)).
 
+ * Add Das-Dennis weight initialization method
+   ([#295](https://github.com/mlpack/ensmallen/pull/295)).
```
zoq, line 41:
```suggestion
   ([#295](https://github.com/mlpack/ensmallen/pull/295)).

```

zoq (approved):
Looks great to me.

include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
coatless: @zoq you are correct. We do not have to worry about `oss` under `throw`. As an example, this pattern is used quite frequently elsewhere: 

https://github.com/mlpack/ensmallen/blob/1e70823989a67c89dc4909ff4152bd650b3abb77/include/ensmallen_bits/problems/logistic_regression_function_impl.hpp

coatless: So, this doesn't need to change to using the `Info` portion. Whoops!

include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
coatless: So, thinking about this... The lambda usage is nice, but should we maybe prefer to write this as a standalone function? Similar to the link post? Also, I think the code on the link post is 1-1 with what is written here. Do we need to worry about licensing?

coatless (commented):
See notes.

include/ensmallen_bits/moead/weight_init_policies/uniform_init.hpp
jonpsy: The reason for making it lambda inside the function was to secure the scope, it won't be used elsewhere. Regarding the heavy correlation, GFG (Geeks for geeks) is kinda open-sourced, and is made to help people(like Stack Overflow). Besides, finding Binomial Coefficient in O(n) isn't a ground breaking research that we're using without citing :) .

zoq: According to https://www.geeksforgeeks.org/copyright-information/ this is fine, but let's add a comment which references the source for the code.

jonpsy:
Great, now I have to wait another 2 days.

coatless:
@jonpsy not at all! From the first approval it will automatically give a second after 24 hours. Once the second bot-approval is in, we can merge. :)

coatless (approved):

PR #296, opened 2021-06-18
Add Dirichlet Weight Initialization

jonpsy:
Continuing #293 

- [x] Added dirichilet_init.hpp
- [x] Test and verify locally.
- [x] ```using DirichletMOEAD = MOEAD<Dirichlet, Tchebycheff>```
- [x] Document in dirichlet_init.hpp
- [x] Add documentation in optimizers.md
- [x] History.md

### DIFF
--- a/include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp
+++ b/include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp
@@ -1,0 +1,56 @@
+/**
+ * @file dirichlet_init.hpp
+ * @author Nanubala Gnana Sai
+ *
+ * The Dirichlet method of Weight Initialization.
+ *
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_MOEAD_DIRICHLET_HPP
+#define ENSMALLEN_MOEAD_DIRICHLET_HPP
+
+namespace ens {
+
+/**
+ * The Dirichlet method for initializing weights. Sampling a dirichlet distribution
+ * with parameters set to one returns points lying on unit simplex with uniform
+ * distribution.
+ *
+ */
+class Dirichlet
+{
+ public:
+  /**
+   * Constructor for Dirichlet policy.
+   */
+  Dirichlet()
+  {
+    /* Nothing to do. */
+  }
+
+  /**
+   * Generate the reference direction matrix.
+   *
+   * @tparam MatType The type of the matrix used for constructing weights.
+   * @param numObjectives The dimensionality of objective space.
+   * @param numPoints The number of reference directions requested.
+   * @param epsilon Handle numerical stability after weight initialization.
+   */
+  template<typename MatType>
+  MatType Generate(const size_t numObjectives,
+                   const size_t numPoints,
+                   const double epsilon)
+  {
+    MatType weights = arma::randg<MatType>(numObjectives, numPoints,
+        arma::distr_param(1.0, 1.0)) + epsilon;
+    // Normalize each column.
+    return arma::normalise(weights, 1, 0);
+  }
+};
+
+} // namespace ens
+
+#endif
--- a/include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp
+++ b/include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp
@@ -15,10 +15,9 @@
 namespace ens {
 
 /**
- * The Dirichlet method for initializing weights. Sampling a dirichlet distribution
- * with parameters set to one returns points lying on unit simplex with uniform
- * distribution.
- *
+ * The Dirichlet method for initializing weights. Sampling a 
+ * Dirichlet distribution with parameters set to one returns 
+ * point lying on unit simplex with uniform distribution.
  */
 class Dirichlet
 {
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -37,6 +37,9 @@
  * Introduce Policy Methods for MOEA/D-DE
    ([#293](https://github.com/mlpack/ensmallen/pull/293)).
 
+ * Add Dirichlet Weight Initialization
+   ([#296](https://github.com/mlpack/ensmallen/pull/296)).
+
 ### ensmallen 2.16.1: "Severely Dented Can Of Polyurethane"
 ###### 2021-03-02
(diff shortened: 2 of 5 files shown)


include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp
```diff
@@ -0,0 +1,56 @@
...
+#define ENSMALLEN_MOEAD_DIRICHLET_HPP
+
+namespace ens {
+
+/**
+ * The Dirichlet method for initializing weights. Sampling a dirichlet distribution
+ * with parameters set to one returns points lying on unit simplex with uniform
+ * distribution.
+ *
+ */
```
jonpsy, line 22: @zoq  Not sure what should be the source ?

zoq: I don't think there is one we should reference here.

zoq (commented):
Can you add a test for the policy as well?

jonpsy:
You have any specifics in mind?

zoq:
We can just use the exciting one but change the policy.

jonpsy:
I could test it against a difficult ZDT problem. See how it fares? The paper has tried testing against some ZDT problems. We can see how much the theoretical results match.

zoq:
Right, let's test it one e.g. one ZDT problem.

include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp
```diff
@@ -0,0 +1,56 @@
...
+   */
+  template<typename MatType>
+  MatType Generate(const size_t numObjectives,
+                   const size_t numPoints,
+                   const double epsilon)
+  {
+    MatType weights = arma::randg<MatType>(numObjectives, numPoints,
+        arma::distr_param(1.0, 1.0)) + epsilon;
+    // Normalize each column.
+    return arma::normalise(weights, 1, 0);
```
coatless, line 50: So, thinking about this... We have a gamma distribution given as: 

```
Y_i ~ gamma(alpha_i, 1)
```

From there, we're creating the normalization of:

```
X = sum_{i = 1}^{n} Y_i
V = (Y_1/X, Y_2/X, ... Y_n/X) ~ Dirchlet(alpha_1, ... alpha_n)
```
https://en.wikipedia.org/wiki/Dirichlet_distribution#Gamma_distribution

Looking at: 

http://arma.sourceforge.net/docs.html#normalise

`normalise` is giving is a unit _p_-norm. With `p = 1`, we're getting: 

```
X = sum abs(Y_i)
```

The `abs()` goes against the summation component, but gamma RVs will always be > 0.

So, long story short... We're doing one extra operation, but the end outcome will be the same.

coatless (approved):

include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp
```diff
@@ -0,0 +1,56 @@
...
+   * @param numObjectives The dimensionality of objective space.
+   * @param numPoints The number of reference directions requested.
+   * @param epsilon Handle numerical stability after weight initialization.
+   */
+  template<typename MatType>
+  MatType Generate(const size_t numObjectives,
+                   const size_t numPoints,
+                   const double epsilon)
+  {
+    MatType weights = arma::randg<MatType>(numObjectives, numPoints,
```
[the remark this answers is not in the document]
jonpsy: alpha = 1. See [this](https://cs.stackexchange.com/a/132116).

jonpsy:
Planning to add
```using DirichletMOEAD = MOEAD<Dirichlet, Tchebycheff>```

File: include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp (hunk 2 above)
Comment (jonpsy) [reply include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp:50]: I think ```p=1``` gives something like

```std::sqrt( elem0^2 + elem1^2 + elem2^2 ..... elemn^2)``` and not the sum if I've gotten this correctly. I suppose you're saying, assuming X = sum abs(Y_i); abs part is unnecessary?

**Comment by jonpsy** [comment]
> Right, let's test it one e.g. one ZDT problem.

I tested it against ZDT3 problem. Ran it over 1000 times using random seed using this bash command

```bash
#!/bin/bash
for (( i = 0; i < 1000; i++))
do
 ./ensmallen_tests MOEADDIRICHLETZDT3Test --rng-seed time
done
```

include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp
coatless: Nope, that would be `p=2`/ l2-norm. This is using an l1-norm by column.

`p=1` gives: 

```cpp
std::pow( abs(elem0)^1 + abs(elem1)^1 + abs(elem2)^1 ..... abs(elemn)^1, 1.0);
```
or:

```
(abs(elem0)^1 + abs(elem1^1 + abs(elem2^1  + ..... + abs(elemn^1))^1 
= abs(elem0)^1 + abs(elem1)^1 + abs(elem2)^1  + ..... +  abs(elemn)^1 
= abs(elem0) + abs(elem1) + abs(elem2) + ..... + abs(elemn)
```

 I think by-column is correct instead of by-row due to the weights being structured `numObjectives x numPoints` (norm across objectives under a single point)

include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp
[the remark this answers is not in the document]
coatless: Okay, so this weight policy is being designed with a specific application in mind so that `alpha` doesn't need to be flexible.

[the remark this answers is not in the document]
jonpsy: Yes, since we want uniform distribution. The alphas vectors are set to ones vector.

include/ensmallen_bits/moead/weight_init_policies/dirichlet_init.hpp
jonpsy: Yes, I've done it in a column basis so the normalization of each point is ensured. Just revised L-norms, I was mistaken about the L1 norm formula, thanks for pointing it out.

I think the "abs" part shouldn't really be that of a concern, either way, we know each element is > 0. We have one line of code which do the thing for us, and I really don't think the additional "abs" would severely hinder the performance.

jonpsy:
I suppose it's fair to say this PR is GTM?

zoq (approved):
Looks great to me.

zoq:
Let's see if the build comes back green and merge it afterwards.

zoq:
Thanks :+1:

PR #297, opened 2021-06-22
Update README for tests

rcurtin:
Some of the commentary for the review of the ensmallen JMLR paper suggested adding some instructions in the README for how to build and run the tests.  So, I made some quick changes to the README and here they are.

@jonpsy I also noticed that some of the other HISTORY.md entries for PRs you've worked on were under an already-released version, so I moved them up to the next release. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ cmake ..
 make ensmallen_tests
 ```
 
-Then, you can run the tests with CTest:
+Then, you can run the tests with [CTest](https://cmake.org/cmake/help/latest/manual/ctest.1.html):
 
 ```
 ctest .
--- a/README.md
+++ b/README.md
@@ -40,13 +40,18 @@ sudo make install
 ### Testing
 
 If you would like to run the tests, once you have checked out the repository you
-can follow the steps below:
+can follow the steps below to build the tests:
 
 ```
 mkdir build
 cd build
 cmake ..
 make ensmallen_tests
+```
+
+Then, you can run the tests with CTest:
+
+```
 ctest .
 ```
 
--- a/README.md
+++ b/README.md
@@ -37,6 +37,20 @@ cmake ..
 sudo make install
 ```
 
+### Testing
+
+If you would like to run the tests, once you have checked out the repository you
+can follow the steps below:
+
+```
+mkdir build
+cd build
+cmake ..
+make ensmallen_tests
+ctest .
+```
+
+All of the tests are in the `tests/` subdirectory.
 
 ### Example Usage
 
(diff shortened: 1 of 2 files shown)


jonpsy:
Thoughtful, nice work!

README.md
```diff
@@ -37,6 +37,25 @@ cmake ..
...
+can follow the steps below to build the tests:
+
+```
+mkdir build
+cd build
+cmake ..
+make ensmallen_tests
+```
+
+Then, you can run the tests with CTest:
```
coatless, line 52: Maybe add a link to CTest? 

```suggestion
Then, you can run the tests with [CTest](https://cmake.org/cmake/help/latest/manual/ctest.1.html):
```

coatless (approved):
Looks good!

zoq:
My interpretation is that the instruction in the README to install ensmallen, might create the impression you need `sudo` to install; but you could also just set `CMAKE_INSTALL_PREFIX` to install into a another folder.

```
mkdir build
cd build
cmake ..
sudo make install
```

conradsnicta:
> My interpretation is that the instruction in the README to install ensmallen, might create the impression you need `sudo` to install

I get the same impression.  The JMLR editor's comments are not so much about running the tests, but about installing ensmallen:
"... the authors may want to consider adding instructions for **installing** for testing purposes **without root access**..."

The "testing purposes" part is not really about running the pre-made tests, but about letting people play around with ensmallen on various systems (where they may or may not have root access).

So the problematic part in the current version of README is indeed this bit: ```sudo make install```

Extracts from Armadillo's [README.md](https://gitlab.com/conradsnicta/armadillo-code/-/blob/10.5.x/README.md) that can be adapted for ensmallen:

      Step 4
      (...)
      To install the library and headers in an alternative directory,
      use the additional option CMAKE_INSTALL_PREFIX in this form:
      
      cmake . -DCMAKE_INSTALL_PREFIX:PATH=alternative_directory
      
      (...)

      Step 5
      If you and have access to root/administrator/superuser privileges
      (ie. able to use "sudo") and didn't use the CMAKE_INSTALL_PREFIX option,
      type the following command:
      
      sudo make install
      
      If you don't have root/administrator/superuser privileges,
      make sure that you use the CMAKE_INSTALL_PREFIX option in Step 4,
      and type the following command:
      
      make install
      

Obviously the ```cmake .``` part would need to be changed to ```cmake ..```

rcurtin:
Ha, I guess I must have sampled the words poorly when reading or something... let me try again tomorrow :smile:

conradsnicta:
Created PR with improved installation instructions: https://github.com/mlpack/ensmallen/pull/300

rcurtin:
Let's go with #300 instead.  Next time I'll be a bit more careful when reading the reviewer feedback and try to parse it correctly :see_no_evil:

PR #298, opened 2021-06-22
ZDT4 Evaluate Fix

jonpsy:
Minor fix.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -37,6 +37,9 @@
  * Introduce Policy Methods for MOEA/D-DE
    ([#293](https://github.com/mlpack/ensmallen/pull/293)).
 
+ * ZDT4 Evaluate Fix
+   ([#298](https://github.com/mlpack/ensmallen/pull/298)).
+
 ### ensmallen 2.16.1: "Severely Dented Can Of Polyurethane"
 ###### 2021-03-02
  * Fix test compilation issue when `ENS_USE_OPENMP` is set
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -37,9 +37,6 @@
  * Introduce Policy Methods for MOEA/D-DE
    ([#293](https://github.com/mlpack/ensmallen/pull/293)).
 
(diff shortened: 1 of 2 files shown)


HISTORY.md
```diff
@@ -37,6 +37,9 @@
  * Introduce Policy Methods for MOEA/D-DE
    ([#293](https://github.com/mlpack/ensmallen/pull/293)).
 
+ * ZDT4 Evaluate Fix
+   ([#298](https://github.com/mlpack/ensmallen/pull/298)).
```
zoq, line 41: ZDT wasn't part of a release, so no need to include the PR in the HISTORY.

coatless: @zoq I think he's trying to avoid running into having the CI pick up an error if HISTORY wasn't modified.

jonpsy: Both of you are correct. In either case, this change isn't anything significant, so we can ignore HISTORY.

zoq: Right, happy to ignore the failing build in this case, in favor of a cleaner HISTORY.

zoq (approved):
Looks good to me.

jonpsy:
Let's hold it for a while, I'm getting non-sensical values again. Not sure if algorithm is wrong or test suite or the code I wrote. But, the values are non-sensical for both NSGA-II and MOEAD, so I have a good reason to believe this algo could be wrong.

jonpsy:
Oh! It turned out the bounds in variable space for the problem is strictly [0, 1]. In my testing I used [-1 , 1] that caused ```arma::datum::nan``` values :D.

jonpsy:
Have this merged? @zoq

zoq:
Sure, we can merge this one, was waiting because you wanted to hold it for a while.

jonpsy:
Actually I was holding #299, not this one. This one's ready.

coatless (approved):

zoq:
Thanks :+1:

PR #299, opened 2021-06-22
Repair MOEAD

jonpsy:
Let's keep this open while I fix MOEAD entirely 🔧.

### DIFF
--- a/include/ensmallen_bits/moead/decomposition_policies/pbi_decomposition.hpp
+++ b/include/ensmallen_bits/moead/decomposition_policies/pbi_decomposition.hpp
@@ -63,11 +63,11 @@ class PenaltyBoundaryIntersection
   {
     typedef typename VecType::elem_type ElemType;
     //! A unit vector in the same direction as the provided weight vector.
-    const VecType referenceDirection = weight / arma::norm(weight, 2);
+    const VecType referenceDirection = weight / arma::norm(weight);
     //! Distance of F(x) from the idealPoint along the reference direction.
     const ElemType d1 = arma::dot(candidateFitness - idealPoint, referenceDirection);
     //! The perpendicular distance of F(x) from reference direction.
-    const ElemType d2 = arma::norm(candidateFitness - (idealPoint + d1 * referenceDirection), 2);
+    const ElemType d2 = arma::norm(candidateFitness - (idealPoint + d1 * referenceDirection));
 
     return d1 + static_cast<ElemType>(theta) * d2;
   }
--- a/include/ensmallen_bits/moead/decomposition_policies/pbi_decomposition.hpp
+++ b/include/ensmallen_bits/moead/decomposition_policies/pbi_decomposition.hpp
@@ -63,11 +63,11 @@ class PenaltyBoundaryIntersection
   {
     typedef typename VecType::elem_type ElemType;
     //! A unit vector in the same direction as the provided weight vector.
-    const VecType referenceDirection = weight / arma::norm(weight, 1);
+    const VecType referenceDirection = weight / arma::norm(weight, 2);
     //! Distance of F(x) from the idealPoint along the reference direction.
     const ElemType d1 = arma::dot(candidateFitness - idealPoint, referenceDirection);
     //! The perpendicular distance of F(x) from reference direction.
-    const ElemType d2 = arma::norm(candidateFitness - (idealPoint + d1 * referenceDirection), 1);
+    const ElemType d2 = arma::norm(candidateFitness - (idealPoint + d1 * referenceDirection), 2);
 
     return d1 + static_cast<ElemType>(theta) * d2;
   }
--- a/include/ensmallen_bits/moead/moead_impl.hpp
+++ b/include/ensmallen_bits/moead/moead_impl.hpp
@@ -291,7 +291,7 @@ Optimize(std::tuple<ArbitraryFunctionType...>& objectives,
   } // End of pass over all the generations.
 
   // Set the candidates from the Pareto Set as the output.
-  paretoSet.resize(population[0].n_rows, population[0].n_cols, population.size());
+  paretoSet.set_size(population[0].n_rows, population[0].n_cols, population.size());
 
   // The Pareto Front is stored, can be obtained via ParetoSet() getter.
   for (size_t solutionIdx = 0; solutionIdx < population.size(); ++solutionIdx)
@@ -301,7 +301,7 @@ Optimize(std::tuple<ArbitraryFunctionType...>& objectives,
   }
 
   // Set the candidates from the Pareto Front as the output.
-  paretoFront.resize(populationFitness[0].n_rows, populationFitness[0].n_cols,
+  paretoFront.set_size(populationFitness[0].n_rows, populationFitness[0].n_cols,
       populationFitness.size());
 
   // The Pareto Front is stored, can be obtained via ParetoFront() getter.
--- a/include/ensmallen_bits/moead/moead_impl.hpp
+++ b/include/ensmallen_bits/moead/moead_impl.hpp
@@ -312,6 +312,9 @@ Optimize(std::tuple<ArbitraryFunctionType...>& objectives,
         arma::conv_to<arma::mat>::from(populationFitness[solutionIdx]);
   }
 
+  // Assign iterate to first element of the Pareto Set.
+  iterate = population[0];
+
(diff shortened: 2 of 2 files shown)


jonpsy:
PBI decomposition, arma::norm should be 2 I think.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -312,6 +312,9 @@ Optimize(std::tuple<ArbitraryFunctionType...>& objectives,
         arma::conv_to<arma::mat>::from(populationFitness[solutionIdx]);
   }
 
+  // Assign iterate to first element of the Pareto Set.
+  iterate = population[0];
```
zoq, line 316: Nice catch, quite an important fix.

include/ensmallen_bits/moead/decomposition_policies/pbi_decomposition.hpp
```diff
@@ -63,11 +63,11 @@ class PenaltyBoundaryIntersection
   {
     typedef typename VecType::elem_type ElemType;
     //! A unit vector in the same direction as the provided weight vector.
-    const VecType referenceDirection = weight / arma::norm(weight, 1);
+    const VecType referenceDirection = weight / arma::norm(weight, 2);
     //! Distance of F(x) from the idealPoint along the reference direction.
     const ElemType d1 = arma::dot(candidateFitness - idealPoint, referenceDirection);
     //! The perpendicular distance of F(x) from reference direction.
-    const ElemType d2 = arma::norm(candidateFitness - (idealPoint + d1 * referenceDirection), 1);
+    const ElemType d2 = arma::norm(candidateFitness - (idealPoint + d1 * referenceDirection), 2);
```
jonpsy, line 70: This change is related to @coatless [comment](https://github.com/mlpack/ensmallen/pull/296#discussion_r654672252) on L2 norm. I misunderstood that concept which reflected here.

The paper takes "distance" (i.e. L2 norm) for finding d1 and d2. Here's a reference pymoo implementation

https://github.com/msu-coinlab/pymoo/blob/b91212185ab04d006741372a4a480608b8d32e68/pymoo/decomposition/util.py#L5-L9

zoq:
> It's a little unreasonable for me to ask if you find any more bugs. But from my side, I think this is good to go. I'll still wait till our next meet in Monday to approve these changes.
> 
> Best
> Sai

Currently in a debug session, let you know if I find anything.

include/ensmallen_bits/moead/decomposition_policies/pbi_decomposition.hpp
coatless: @jonpsy should we make a utility function that wraps `norm(x, 2)` called `norm_l2()` to clarify intent and avoid future headaches?

jonpsy: In armadillo? I think that'd be @conradsnicta's call

coatless: No need to go upstream, this could be placed inside `ensmallen`.

By default, [`norm(x)` has `p = 2`](http://arma.sourceforge.net/docs.html#norm) per docs.

jonpsy: I think mathematical abstractions should limit to armadillo. I'll change norm(x, 2) to norm then. That'll be better.

coatless: @jonpsy that wasn't my intent by stating that! In fact, I'd prefer it with the explicit `norm(x, 2)` vs. `norm()`. It's almost always better to be explicit and avoid relying on defaults as they may inadvertently change. 

I'm of the mindset that a general function is great, but creating helper functions to commonly used tasks is ideal.

jonpsy: @coatless I see, I guess I misinterpreted it. I think using ```arma::norm()``` is better. The general notion is that norm refers to  L2 norm unless otherwise stated. It looks like other libraries follow similar culture as well (see ref. implementation).

jonpsy:
@zoq From our last chat about using ```arma::join_rows```, making arma data structures mutable. I was wondering if std::vector is REALLY required? 

I hate the for loop we use in the end to convert each member of bestFront to a part of our exposed paretoFront cube. Feels silly to me.  

Besides the ```population``` variable is simply destroyed instead of being "moved" into bestFront. That bothers me as well.

jonpsy:
To add on this, seeing Rahul's work on NSGA-III, even he didn't rely on std::vectors and used purely armadillo data structures.

coatless:
> To add on this, seeing Rahul's work on NSGA-III, even he didn't rely on std::vectors and used purely armadillo data structures.

I don't think I saw anything with NSGA-III hit any of the repos. Where is the NSGA-III work at?

jonpsy:
@coatless https://github.com/mlpack/ensmallen/pull/120

jonpsy:
https://github.com/mlpack/ensmallen/blob/9d55d24b3c80cbc129398c3cb88bb6bfe612f115/include/ensmallen_bits/moead/moead_impl.hpp#L182

Was wondering if this is really requierd?

jonpsy:
And same with ```EvaluateObjectives``` call at the end of the loop. It's our computational bottleneck, we ought to use it scarcely.

zoq:
We could remove both, in this case we run call `EvaluateObjectives`, `maxGenerations` times instead of `maxGenerations + 2`, but depending on the objective function this saves us only a small fraction.

jonpsy:
Yes, a small fraction to very high as was observed in the portfolio notebook. There might be realistic cases in which Evaluating objectives could be complex, hence my idea.

zoq:
Agreed, I don't think the two calls you mentioned have any real purpose right now, so let's remove them.

jonpsy:
Okay, the first call over them is required because it helps in caching ```populationFitness```. The last-one can be removed though.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -300,7 +300,6 @@ Optimize(std::tuple<ArbitraryFunctionType...>& objectives,
         arma::conv_to<arma::mat>::from(population[solutionIdx]);
   }
 
-  EvaluateObjectives(population, objectives, populationFitness);
   // Set the candidates from the Pareto Front as the output.
   paretoFront.resize(populationFitness[0].n_rows, populationFitness[0].n_cols,
```
conradsnicta, line 304: Using [.set_size()](http://arma.sourceforge.net/docs.html#set_size) would be more efficient than [.resize()](http://arma.sourceforge.net/docs.html#resize). The latter preserves data, which doesn't look necessary here. All the slices of `paretoFront` are overwritten afterwards, so there doesn't seem anything to preserve.

jonpsy: Nice catch!

zoq:
I guess the remaining question here is if we should use `arma::join_rows` over `std::vector.push_back`?

jonpsy:
Yeah, I think we should rely entirely on ```arma``` based codes instead of ```std::vector```.

zoq:
The question is how does that influence the performance, I don't think we will see it with the low dimensional problems we tested so far, but maybe with other problems. If `arma::join` copies the complete matrix I would expect this to be slower.

jonpsy:
Yes, @coatless mentioned the same. I concede we keep it. That and, if we went with my preposition we would have to change NSGA-II as well. That's a lot of last minute changes. 

So, I guess this PR is ready unless you have some final checks?

zoq (approved):
No further comments from my side, thanks for all the patches.

jonpsy:
@coatless  I take it you have no objections?

conradsnicta:
@jonpsy Can you add to HISTORY.md with a short description of this PR and the PR number?

jonpsy:
Sure thing.

zoq:
> @jonpsy Can you add to HISTORY.md a short description of this PR and the PR number?

MOEAD wasn't part of any release yet, and there is already an entry in HISTORY.md for the main implementation, so I don't think  we should add another entry, we could reference #299 for the same entry.

jonpsy:
Gotta agree with @zoq. The chances are more of a cleanup and can be seen as a extended part of original MOEAD PR.

Issue #301, opened 2021-06-23
disable building tests by default


conradsnicta:
With the aim of making ensmallen more approachable (easier/faster to install), I suggest to turn off building the tests by default.

For users that want to try out ensmallen, the tests are not necessary and cause an annoying slowdown in the installation.  The tests are useful to developers of ensmallen, but they are not useful to the users of ensmallen.

A target such as "tests" or "ensmallen_tests" can be used to explicitly enable building of the tests (eg. `make ensmallen_tests`).

Related comments and PR in mlpack:
* https://github.com/mlpack/mlpack/pull/2926
* https://github.com/mlpack/mlpack/issues/2524#issuecomment-825367002

rcurtin:
Yeah, I agree, like you pointed out we did the same for mlpack.  Any other opinions on this one?  I'm happy to put together a quick PR for it. :+1:

zoq:
Sounds like a good idea to me as well.

conradsnicta:
@rcurtin I suggest merging https://github.com/mlpack/ensmallen/pull/300 before the quick PR for this.  The quick PR would need to adjust the wording in README.md, ie. remove mention of tests being built as part of the installation process.

conradsnicta:
issue resolved via https://github.com/mlpack/ensmallen/pull/303

PR #303, opened 2021-06-27
don't build tests by default

conradsnicta:
Disable building the tests by default, as discussed in https://github.com/mlpack/ensmallen/issues/301
Tests can still be optionally built.

Advantages: faster and easier installation by users.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,8 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
+ * Disable building the tests by default for faster installation
+   ([#303](https://github.com/mlpack/ensmallen/pull/303)).
+
  * Improved installation and compilation instructions
    ([#300](https://github.com/mlpack/ensmallen/pull/300)).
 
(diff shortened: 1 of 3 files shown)


conradsnicta:
@rcurtin Might be useful to release ensmallen 2.17 after this PR is merged.

zoq (approved):
Nice, we should release a new version after the PR is merged.

coatless:
@zoq I would say after #299 is merged in.

Otherwise, I agree. Ensmallen is overdue for a new release :)

jonpsy:
> @zoq I would say after #299 is merged in.
> 
> Otherwise, I agree. Ensmallen is overdue for a new release :)

++

jonpsy:
Oh and also, since we broke one public API "BestFront() " from NSGA-II. Should we make a major or a minor release?

rcurtin:
It looks like `Front()` was removed and replaced with `ParetoFront()`?  Why not just re-add `Front()` and mark it deprecated?  I can open a PR for that tomorrow unless you want to do it first.

jonpsy:
I can try depending on how soon my charger gets fixed  :)

birm (approved):
seems like a good idea!

rcurtin:
@jonpsy done in #304---let me know what you think. :)

rcurtin (approved):
Looks great to me!  Let's merge it. :)  Thanks @conradsnicta!

PR #304, opened 2021-06-28
Add Front() back into the NSGA2 class.

rcurtin:
This came out of the discussion for #303---to avoid breaking reverse compatibility, I re-added the `Front()` function to the `NSGA2` class.  Actually it turned out to be slightly more difficult than I thought: to match the API, we have to return a reference, so we actually needed to re-add a member to the `NSGA2` class.  But, that member is only populated if the user calls `Front()`.

Anyway, this can all be removed in the next major version release of ensmallen, but until then, anyone who happens to be using `Front()` from `NSGA2` should not have their code broken. :)

### DIFF
--- a/include/ensmallen_bits/nsga2/nsga2.hpp
+++ b/include/ensmallen_bits/nsga2/nsga2.hpp
@@ -181,7 +181,8 @@ class NSGA2
   /**
    * Retrieve the best front (the Pareto frontier).  This returns an empty
    * vector until `Optimize()` has been called.  Note that this function is
-   * deprecated and will be removed in ensmallen 3.10.0!  Use `ParetoFront()` instead.
+   * deprecated and will be removed in ensmallen 3.x!  Use `ParetoFront()`
+   * instead.
    */
   ens_deprecated const std::vector<arma::mat>& Front()
   {
@@ -378,7 +379,7 @@ class NSGA2
   arma::cube paretoFront;
 
   //! A different representation of the Pareto front, for reverse compatibility
-  //! purposes.  This can be removed when ensmallen 3.10.0 is released!  (Along
+  //! purposes.  This can be removed when ensmallen 3.x is released!  (Along
   //! with `Front()`.)  This is only populated when `Front()` is called.
   std::vector<arma::mat> rcFront;
 };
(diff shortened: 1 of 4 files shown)


include/ensmallen_bits/nsga2/nsga2.hpp
```diff
@@ -177,6 +178,25 @@ class NSGA2 {
   //! `Optimize()` has been called.
   const arma::cube& ParetoFront() const { return paretoFront; }
 
+  /**
+   * Retrieve the best front (the Pareto frontier).  This returns an empty
+   * vector until `Optimize()` has been called.  Note that this function is
+   * deprecated and will be removed in ensmallen 3.10.0!  Use `ParetoFront()` instead.
```
coatless, line 184:
```suggestion
   * deprecated and will be removed in ensmallen 3.0.0!  Use `ParetoFront()` instead.
```

We're still on the 2.x track now...

rcurtin: We've been starting the minor versions with `.10`, a bit like Armadillo, so that versions sort correctly numerically too. :)  Anyway the issue can be avoided by just using `3.x`... :)  I made that change in 9906c22d. :+1:

coatless: @rcurtin missed that. Thanks for the clarification!

jonpsy:
Thanks for taking the time to do this ✌.

zoq (approved):
Looks good to me.

coatless (approved):

conradsnicta:
@rcurtin We should now be good for releasing ensmallen 2.17.0 :)  
Though one more option would be to update the citation details in the readme with the JMLR article

rcurtin:
I think we should wait for the paper to be published by JMLR before updating the citation (but we definitely should update that when that time comes).  @jonpsy had said we should wait on #299 before releasing 2.17.0, which I think is just fine since that is nearly ready to merge. :+1:

jonpsy:
Thanks for the briefing @rcurtin. We're on the finalizing stage of MOEAD Optimizer which should be over within a day or two. On the upcoming meet (Friday, this week) we could go over the entire ensmallen library and finalize changes, discuss release notes etc. On the start of next week, we'll release the new ensmallen, I have a name in mind :).

conradsnicta:
@rcurtin MOEAD optimiser fixes merged: https://github.com/mlpack/ensmallen/pull/299; ensmallen 2.17.0 looks good to go.

PR #314, opened 2021-08-06
Update travis config to build the ensmallen tests

zoq:
I'm currently configuring a Jenkins job to replace the travis build but in the meantime I figured we can still use the existing travis credits.


rcurtin (approved):
I'm surprised we're not already out of travis credits but this change looks good to me. :) I am not sure why the master build is failing but that seems very unrelated...

zoq:
> I'm surprised we're not already out of travis credits but this change looks good to me. :) I am not sure why the master build is failing but that seems very unrelated...

Right, not sure, also like like it takes about 4 hours to run this on dealgood, I guess dealgood was busy building the docker containers. Let's see what happens if we exclude dealgood for now.

PR #315, opened 2021-08-07
GradClipByValue callback

zoq:
Writing a recurrent network test case where the gradient explodes, so implemented gradient clipping.

### DIFF
--- a/doc/callbacks.md
+++ b/doc/callbacks.md
@@ -144,13 +144,43 @@ smorms3.Optimize(lrfTrain, coordinates, cb);
 
 </details>
 
+### GradClipByNorm
+
+Given the gradient, and a maximum norm value, the callback normalizes the
+gradient so that its L2-norm is less than or equal to the given maximum norm
+value.
+
+#### Constructors
+
+ * `GradClipByNorm(`_`maxNorm`_`)`
+
+#### Attributes
+
+| **type** | **name** | **description** | **default** |
+|----------|----------|-----------------|-------------|
+| `double` | **`maxNorm`** | The maximum clipping value. | |
+
+#### Examples:
+
+<details open>
+<summary>Click to collapse/expand example code.
+</summary>
+
+```c++
+AdaDelta optimizer(1.0, 1, 0.99, 1e-8, 1000, 1e-9, true);
+
+RosenbrockFunction f;
+arma::mat coordinates = f.GetInitialPoint();
+optimizer.Optimize(f, coordinates, GradClipByNorm(0.3));
+```
+
 ### GradClipByValue
 
 Clips the gradient to a specified min and max.
 
 #### Constructors
 
- * `EarlyStopAtMinLoss(`_`min, max`_`)`
+ * `GradClipByValue(`_`min, max`_`)`
 
 #### Attributes
 
--- a/doc/callbacks.md
+++ b/doc/callbacks.md
@@ -150,7 +150,7 @@ One difficulty with optimization is that large parameter gradients can lead an
 optimizer to update the parameters strongly into a region where the loss
 function is much greater, effectively undoing much of the work done to get to
 the current solution. Such large updates during the optimization can cause a
-numerical overflow or underflow, often referred to as "exploding gradients." The
+numerical overflow or underflow, often referred to as "exploding gradients". The
 exploding gradient problem can be caused by: Choosing the wrong learning rate
 which leads to huge updates in the gradients.  Failing to scale a data set
 leading to very large differences between data points.  Applying a loss function
@@ -192,7 +192,7 @@ One difficulty with optimization is that large parameter gradients can lead an
 optimizer to update the parameters strongly into a region where the loss
 function is much greater, effectively undoing much of the work done to get to
 the current solution. Such large updates during the optimization can cause a
-numerical overflow or underflow, often referred to as "exploding gradients." The
+numerical overflow or underflow, often referred to as "exploding gradients". The
 exploding gradient problem can be caused by: Choosing the wrong learning rate
 which leads to huge updates in the gradients.  Failing to scale a data set
 leading to very large differences between data points.  Applying a loss function
--- a/doc/callbacks.md
+++ b/doc/callbacks.md
@@ -146,9 +146,21 @@ smorms3.Optimize(lrfTrain, coordinates, cb);
 
 ### GradClipByNorm
 
-Given the gradient, and a maximum norm value, the callback normalizes the
-gradient so that its L2-norm is less than or equal to the given maximum norm
-value.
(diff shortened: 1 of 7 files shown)


doc/callbacks.md
```diff
@@ -136,14 +136,73 @@ EarlyStopAtMinLoss cb(
...
 smorms3.Optimize(lrfTrain, coordinates, cb);
 ```
 
 </details>
 
+### GradClipByNorm
+
+Given the gradient, and a maximum norm value, the callback normalizes the
+gradient so that its L2-norm is less than or equal to the given maximum norm
+value.
```
rcurtin, line 151: It might be useful to add a sentence or two about where and when this is useful.  (Same for `GradClipByValue`.) :+1:

include/ensmallen_bits/callbacks/grad_clip_by_norm.hpp
```diff
@@ -0,0 +1,60 @@
...
+   */
+  template<typename OptimizerType, typename FunctionType, typename MatType>
+  void Gradient(OptimizerType& /* optimizer */,
+                FunctionType& /* function */,
+                const MatType& /* coordinates */,
+                MatType& gradient)
+  {
+    const double gradientNorm = arma::norm(gradient);
+    if (gradientNorm > maxNorm)
+      gradient = maxNorm * gradient / arma::norm(gradient);
```
rcurtin, line 50:
```suggestion
      gradient = maxNorm * gradient / gradientNorm;
```
Seems like you could save a little computation here? :)

include/ensmallen_bits/callbacks/grad_clip_by_value.hpp
```diff
@@ -0,0 +1,61 @@
...
+   * @param gradient Matrix that holds the gradient.
+   */
+  template<typename OptimizerType, typename FunctionType, typename MatType>
+  void Gradient(OptimizerType& /* optimizer */,
+                FunctionType& /* function */,
+                const MatType& /* coordinates */,
+                MatType& gradient)
+  {
+    for (size_t i = 0; i < gradient.n_elem; ++i)
+      gradient(i) = std::max(lower, std::min(gradient(i), upper));
```
rcurtin, line 48: Maybe you can use `clamp()` here instead?  http://arma.sourceforge.net/docs.html#clamp_member

tests/callbacks_test.cpp
```diff
@@ -783,3 +783,76 @@ TEST_CASE("ReportCallbackTest", "[CallbacksTest]")
...
+  arma::mat coordinates = f.GetInitialPoint();
+
+  StandardSGD s(0.0003, 1, 10, 1e-9, true);
+
+  std::stringstream stream;
+  s.Optimize(f, coordinates, GradClipByNorm(0.5), Report(0.1, stream));
+
+  // We don't store the gradient during the optimization process, so we use the
+  // output of the Report callback function to check if the gradient is
+  // clipped.
```
rcurtin, line 802: Nice, this is a clever test!

rcurtin (commented):
Awesome!  This is nice to add and the implementation is quite simple. :+1:

include/ensmallen_bits/callbacks/grad_clip_by_value.hpp
zoq: Ahh, yeah, I was looking if there is an existing armadillo function, but searched for the wrong keyword.

conradsnicta: Quick note on `.clamp()` vs `clamp()`.  The member function [`.clamp()`](http://arma.sourceforge.net/docs.html#clamp_member) is in Armadillo 10.5 onwards. The stand-alone function [`clamp()`](http://arma.sourceforge.net/docs.html#clamp) is in Armadillo 4.400 onwards.  It was expanded to handle cubes in 7.900, and sparse matrices in 9.600.

Currently the minimum Armadillo version required by ensmallen in 8.400.  I recommend bumping this to 9.800, which is in Ubuntu 20.04 LTS.  This will allow `clamp()` to be used by sparse matrices within ensmallen.

20.04 LTS has been out for more than a year, which is plenty of time for people to upgrade from 18.04.  In any case it's quite straightforward to manually update Armadillo to the latest version.

rcurtin: @conradsnicta had left a comment that pointed out that `.clamp()` wasn't available until recently, so we should probably use the standalone `clamp()` version.  But there is also another observation, which is that `clamp()` did not work on sparse matrices until 9.600.  I think it's not worth updating the minimum version of Armadillo for this (since I'd like to keep it in sync with mlpack's minimum required version of Armadillo), but, it might be worth special-casing this:

```
  #if ARMA_VERSION_MAJOR < 9 || ARMA_VERSION_MAJOR == 9 && ARMA_VERSION_MINOR < 600
    if (arma::is_arma_sparse_type<MatType>::value)
      // use manually implemented loop
    else
      // use clamp()
  #else
    // use clamp()
  #endif
```
  
and then when the time to upgrade the minimum version comes, we can remove this part.  (Alternately, we could add a caveat to the documentation pointing out that if you are on Armadillo < 9.600 and using sparse gradients, that this callback will not work.  The set of people realistically interested in doing that is probably vanishingly small.)

conradsnicta: I'm certainly not keen on having `#if`s throughout the codebase, especially with backported or manually implemented workarounds.  From first hand experience this type of approach is polluting and considerably increases the maintenance burden.  Little bit here and there quickly adds up, and requires tracking (either explicit or implicit).

I don't think we should aim to accommodate everybody's setup -- it's too much work in the end for something that is volunteer driven.  There are only so many hours in the day, and we all have other things on our plates.  It would be far less hassle to simply encourage people to upgrade their version of Armadillo if they want to use sparse gradients.

I'd wager that anybody sufficiently sophisticated to use sparse gradients would know how to update Armadillo on their system.

So overall, I'm strongly in favour of simply adding a caveat to the documentation, with the minimum suggested version of Armadillo at 9.800 (which is in Ubuntu 20.04 LTS).

rcurtin: Yeah, agreed---I think the note in the documentation is probably the way to go here then.

zoq: Sounds reasonable to me, I can open another PR with the necessary changes.

conradsnicta: Resolved by #318

conradsnicta (approved):

PR #318, opened 2021-08-25
Bump minimum Armadillo version to 9.800.

zoq:
As discussed here https://github.com/mlpack/ensmallen/pull/315 we like to bump the minimum Armadillo version to 9.800. I think once merged we should do another release `2.18.0`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ script:
   - if [ $ARMADILLO == "latest" ]; then
       curl https://ftp.fau.de/macports/distfiles/armadillo/`curl https://ftp.fau.de/macports/distfiles/armadillo/ -- | grep '.tar.xz' | sed 's/^.*<a href="\(armadillo-[0-9]*.[0-9]*.[0-9]*.tar.xz\)".*$/\1/' | tail -1` | tar xvJ && cd armadillo*;
     else
-      curl https://ftp.fau.de/macports/distfiles/armadillo/armadillo-8.400.0.tar.xz | tar -xvJ && cd armadillo*;
+      curl https://ftp.fau.de/macports/distfiles/armadillo/armadillo-9.800.1.tar.xz | tar -xvJ && cd armadillo*;
     fi
   - cmake . && make && sudo make install && cd ..
   - mkdir build && cd build && cmake .. && make ensmallen_tests -j2
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ script:
   - if [ $ARMADILLO == "latest" ]; then
       curl https://ftp.fau.de/macports/distfiles/armadillo/`curl https://ftp.fau.de/macports/distfiles/armadillo/ -- | grep '.tar.xz' | sed 's/^.*<a href="\(armadillo-[0-9]*.[0-9]*.[0-9]*.tar.xz\)".*$/\1/' | tail -1` | tar xvJ && cd armadillo*;
     else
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -23,7 +23,7 @@ script:
   - if [ $ARMADILLO == "latest" ]; then
       curl https://ftp.fau.de/macports/distfiles/armadillo/`curl https://ftp.fau.de/macports/distfiles/armadillo/ -- | grep '.tar.xz' | sed 's/^.*<a href="\(armadillo-[0-9]*.[0-9]*.[0-9]*.tar.xz\)".*$/\1/' | tail -1` | tar xvJ && cd armadillo*;
     else
-      curl https://ftp.fau.de/macports/distfiles/armadillo/armadillo-8.400.0.tar.xz | tar -xvJ && cd armadillo*;
+      curl https://ftp.fau.de/macports/distfiles/armadillo/armadillo-9.800.1.tar.xz | tar -xvJ && cd armadillo*;
```
conradsnicta, line 26: Suggest to use 9.800.6 instead, directly from Sourceforge:
https://sourceforge.net/projects/arma/files/armadillo-9.800.6.tar.xz
the .6 patch release is the last release in the 9.800.x series.

I'm not going to delete the above file (until at least when ensmallen updates its minimum armadillo version), so it's safe to use.  The previous situation which resulted in old armadillo releases being removed no longer applies.

README.md
```diff
@@ -105,8 +105,8 @@ Please cite the following paper if you use ensmallen in your research and/or
 software. Citations are useful for the continued development and maintenance of
 the library.
 
-* Ryan R. Curtin, Marcus Edel, Rahul Ganesh Prabhu, Suryoday Basak, Zhihao Lou, Conrad Sanderson.  
-  [The ensmallen library for flexible numerical optimization](https://jmlr.org/papers/volume22/20-416/20-416.pdf).  
+* Ryan R. Curtin, Marcus Edel, Rahul Ganesh Prabhu, Suryoday Basak, Zhihao Lou, Conrad Sanderson.
```
conradsnicta, line 108: Please keep the two blank spaces at the end of each line containing the author names and article name. Markdown interprets two spaces followed by a linebreak as a "forced linebreak".  In this case we need the forced linebreaks to ensure proper layout (to facilitate reading).

conradsnicta (changes requested):

rcurtin:
When this is merged, we should also bump it in mlpack, otherwise we can get a situation where mlpack's Armadillo version requirement is less than ensmallen's, so a user could feasible download mlpack, download Armadillo 8.400.0, think "I'm good to go!", then suddenly find out that the newest ensmallen (via the autodownloader) actually needs an even newer version.

shrit (approved):

shrit:
It is strange the error says it is unable to find Armadillo on windows

zoq:
I'll fix the windows build.

zoq:
The auto-downloader downloads the ensmallen master branch? If that is the case, we should change that behavior, it should always download the latest stable version. Anyway, I'll open another PR on mlpack. With the release of another ensmallen, I also open a second PR to update ensmallen.org.

rcurtin:
> The auto-downloader downloads the ensmallen master branch? If that is the case, we should change that behavior, it should always download the latest stable version.

Right, it just downloads the latest stable version.  So, we should just merge mlpack/mlpack#3043 before releasing another version of ensmallen. :+1:

PR #317, opened 2021-08-15
Add citation file.

zoq:
Add citation file to use GitHub's citation feature.

PR #319, opened 2021-09-06
Revert "Add citation file"

conradsnicta:
Reverts mlpack/ensmallen#317


conradsnicta:
GitHub is not using the JMLR article, so revert.  
See https://github.com/mlpack/ensmallen/pull/317#issuecomment-913252636

zoq (approved):
That is too bad.

PR #325, opened 2021-10-13
Fix for C++20 constructor template argument deprecation

### LINKED ISSUE #324: C++20 redudant template argument disallowed in ctor
#### Issue description

If you don't support C++20, I guess this is not an issue but as a heads up, code like that does not compile with C++20:

```c++
class EarlyStopAtMinLossType
{
 public:
  EarlyStopAtMinLossType<MatType>(...) :
  {  }
```
due to the following change that dissallow redundant template type declaration in the ctor and dtor declaration:

https://eel.is/c++draft/diff.cpp17.class#2

This is triggered by `EarlyStopAtMinLossType` by just including `ensmallen.hpp`, but possibly there other instances?

Fix is removing the `<MatType>` from the constructor and destructors.

#### Your environment

 * Tarball 2.17.0 from the website, also current git
 * Linux
 * g++ 11 with C++20 enabled
 * 10.7.1
 * CMake with `set(CMAKE_CXX_STANDARD 20)` in the cmakefile

#### Steps to reproduce

Compile with c++20 enabled, a cpp file that include `ensmallen.hpp`

#### Expected behavior

Compile.

#### Actual behavior

Does not compile

#### Fix

Confirmed that it is fixed by removing the redundant template type declaration to the ctor of EarlyStopAtMinLossType, resulting in:

```c++
namespace ens {

/**
 * Early stopping to terminate the optimization process early if the loss stops
 * decreasing.
 */
class EarlyStopAtMinLossType
{
 public:
  /**
   * Set up the early stop at min loss class, which keeps track of the minimum
   * loss and stops the optimization process if the loss stops decreasing.
   *
   * @param patienceIn The number of epochs to wait after the minimum loss has
   *    been reached or no improvement has been made (Default: 10).
   */
  EarlyStopAtMinLossType(const size_t patienceIn = 10) :
      callbackUsed(false), 
      patience(patienceIn), 
      bestObjective(std::numeric_limits<double>::max()),
      steps(0)
  { /* Nothing to do here */ }

  /**
   * Set up the early stop at min loss class, which keeps track of the minimum
   * loss and stops the optimization process if the loss stops decreasing.
   *
   * @param func, callback to return immediate loss evaluated by the function
   * @param patienceIn The number of epochs to wait after the minimum loss has
   *    been reached or no improvement has been made (Default: 10).
   */
  EarlyStopAtMinLossType(
      std::function<double(const MatType&)> func,
      const size_t patienceIn = 10)
    : callbackUsed(true), 
      patience(patienceIn), 
      bestObjective(std::numeric_limits<double>::max()),
      steps(0), 
      localFunc(func)
  {
    // Nothing to do here
  }
```

ElianeBriand:
Removes the redundant template type parameter from the constructors of `EarlyStopAtMinLossType` to allow inclusion and compilation of `ensmallen.hpp` in programs using C++20.

This is due to [the following change in C++20](https://eel.is/c++draft/diff.cpp17.class#2) (compared to C++17 and earlier) that disallows such non-necessary repeat of the parameter in constructors and destructors.

This does not affect compilation using earlier C++ standard. All `ensmallen_tests` still pass. This does not necessitate any change in user code. Fixes #324

### DIFF
--- a/include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
+++ b/include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp
@@ -32,7 +32,7 @@ class EarlyStopAtMinLossType
    * @param patienceIn The number of epochs to wait after the minimum loss has
    *    been reached or no improvement has been made (Default: 10).
    */
-  EarlyStopAtMinLossType<MatType>(const size_t patienceIn = 10) :
+  EarlyStopAtMinLossType(const size_t patienceIn = 10) :
       callbackUsed(false), 
       patience(patienceIn), 
       bestObjective(std::numeric_limits<double>::max()),
@@ -47,7 +47,7 @@ class EarlyStopAtMinLossType
    * @param patienceIn The number of epochs to wait after the minimum loss has
    *    been reached or no improvement has been made (Default: 10).
    */
-  EarlyStopAtMinLossType<MatType>(
+  EarlyStopAtMinLossType(
       std::function<double(const MatType&)> func,
       const size_t patienceIn = 10)
     : callbackUsed(true), 


ElianeBriand:
> Letting this sit for some hours in case you like to add your name to `COPYRIGHT.txt`.

It's fine, this is very minor.

PR #313, opened 2021-08-04
The BUILD_TESTS option was replaced with 'make ensmallen_tests'.

zoq:
Remove the `BUILD_TESTS` option in favor of `make ensmallen_tests` see https://github.com/mlpack/ensmallen/pull/303 for more details.

PR #326, opened 2021-10-15
Release version 2.18.0: "Fairmount Bagel"

zoq:
This automatically-generated pull request adds the commits necessary to make the 2.18.0 release.

Once the PR is merged, mlpack-bot will tag the release as HEAD~1 (so that it doesn't include the new HISTORY block) and publish it.

Or, well, hopefully that will happen someday.

When you merge this PR, be sure to merge it using a *rebase*.

### Changelog

 * Add gradient value clipping and gradient norm scaling callback ([#315](https://github.com/mlpack/ensmallen/pull/315)).

 * Remove superfluous CMake option to build the tests ([#313](https://github.com/mlpack/ensmallen/pull/313)).

 * Bump minimum Armadillo version to 9.800 ([#318](https://github.com/mlpack/ensmallen/pull/318)).

 * Update Catch2 to 2.13.7 ([#322](https://github.com/mlpack/ensmallen/pull/322)).

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,6 @@
+### ensmallen ?.??.?: "???"
+###### ????-??-??
+
 ### ensmallen 2.18.0: "Fairmount Bagel"
 ###### 2021-10-14
  * Add gradient value clipping and gradient norm scaling callback


rcurtin:
Were the bagels good? :smiley: 

Looks like there are some intermittent test failures; I see this in the AppVeyor logs:

```
-------------------------------------------------------------------------------
MOEADZDTONETest
-------------------------------------------------------------------------------
C:\projects\ensmallen\tests\moead_test.cpp(510)
...............................................................................
C:\projects\ensmallen\tests\moead_test.cpp(544): FAILED:
  REQUIRE( g == Approx(1.0).margin(0.99) )
with expansion:
  1.9990931495 == Approx( 1.0 )
===============================================================================
test cases:   292 |   291 passed | 1 failed
assertions: 13431 | 13430 passed | 1 failed
```

I can't see what the errors are for the Jenkins job:
http://ci.mlpack.org/job/pull-requests-ensmallen-mlpack-build/278/console
but I wouldn't be surprised if they're the same.

Do you think we should handle that issue before we release?  I'd be concerned that if this test does randomly fail sometimes, we'll get some user reports and confusion.  I ran it a few times locally and it looks like the failure rate is somewhere between 10% and 25%.  I think nothing is actually wrong with the method, so I'm working on a PR that will run the test multiple times to get the failure rate down.

zoq:
Agreed let's fix the test issue first before releasing a new version. I will also, fix the jenkins PR job, so it shown up again. 

> so I'm working on a PR that will run the test multiple times to get the failure rate down.

What I had in mind is to just run all the test multiple times, with a different seed. Some time ago I added an option to run the tests with a different seed, so I would just put a simple for loop around that. Since the ensmallen tests can be run relativity fast, I think we don't have to do some clever filtering to get the test time down. I can modify the jenkins job to do that.

zoq:
Exactly, to reduce the number of times we open another PR to adjust thresholds.

PR #327, opened 2021-10-16
Fix MOEAD test stability

rcurtin:
This is a simple PR related to #326 that makes the MOEAD tests much more stable by allowing multiple trials.  @jonpsy let me know what you think.

@zoq I saw also that you've left another comment on #326, so this may not be the best solution if you had something else in mind (we can continue that discussion there, I'll post a reply shortly).

I ran this 1000 times locally and all the tests in the `[MOEADTest]` suite passed.

### DIFF
--- a/tests/moead_test.cpp
+++ b/tests/moead_test.cpp
@@ -516,7 +516,6 @@ TEST_CASE("MOEADZDTONETest", "[MOEADTest]")
   ZDT1<> ZDT_ONE(100);
   const double lowerBound = 0;
   const double upperBound = 1;
-  double g = DBL_MAX;
 
   DefaultMOEAD opt(
       300, // Population size.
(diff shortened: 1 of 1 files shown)


zoq:

zoq:
@jonpsy see https://github.com/mlpack/ensmallen/pull/326#issuecomment-944933561

rcurtin:
> One thing, all these test were ran ~1000 times with various rng generator & it passes each time before this PR, hence the higher margin. So this might be unnecessary.

Hmm, maybe we did it in a different way?  I ran this:

```
$ i=0; while(true); do ./ensmallen_tests --rng-seed $i [MOEADTest] 2>&1 | grep -v 'version:' | grep -v 'Filters:' | grep -v '====' | grep -v 'All tests passed' | grep -v '^[ ]*$'; sleep 1; i=$(($i + 1)); echo $i; done
```

(Sorry that's a little long...)

I handled the other comments in df3a1cf and 23aab9f9.  @jonpsy are you sure that the state of the optimizer doesn't change when `Optimize()` is called?  The method isn't marked `const`, but I didn't dig deeper.

rcurtin:
> To keep the static analysis check happy.

Thanks!  At first I thought I'd need `g` outside the loop, so I defined it out there, but then I guess it wasn't actually necessary there.

PR #330, opened 2021-11-13
Accelerate SGD test time by skipping additional trials.

zoq:
Accelerate SGD test time by skipping additional trials.


rcurtin (approved):
Nice catch!

Issue #329, opened 2021-11-01
Debian build (test) failure on armel


barak:
#### Issue description

The Debian autobuilders are failing on the `armel` architecture for `ensmallen 2.18.0-1`, while succeeding for everything else!

#### Your environment

 * version of ensmallen: 2.18.0-1
 * operating system: Debian (unstable)

For other info, see the package versions installed in the build log.

#### Steps to reproduce

Visit https://buildd.debian.org/status/package.php?p=ensmallen and click on the `Build-Attempted` link with the red background in the `armel` row.

#### Expected behavior

Successful build

#### Actual behavior

Unsuccessful build

conradsnicta:
relevant extract from the rather noisy build log:
```
test 1
    Start 1: ensmallen_tests

1: Test command: /<<PKGBUILDDIR>>/obj-arm-linux-gnueabi/ensmallen_tests
1: Test timeout computed to be: 10000000
1: ensmallen version: 2.18.0 (Fairmount Bagel)
1: armadillo version: 10.6.2 (Keep Calm)
1: invalid fastbin entry (free)
1: 
1: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1: ensmallen_tests is a Catch v2.13.7 host application.
1: Run with -? for options
1: 
1: -------------------------------------------------------------------------------
1: SoftmaxRegressionFunctionPartialGradientTest
1: -------------------------------------------------------------------------------
1: ./tests/scd_test.cpp:187
1: ...............................................................................
1: 
1: ./tests/scd_test.cpp:187: FAILED:
1: due to a fatal error condition:
1:   SIGABRT - Abort (abnormal termination) signal
1: 
1: ===============================================================================
1: test cases:  252 |  251 passed | 1 failed
1: assertions: 8524 | 8523 passed | 1 failed
1: 
1/1 Test #1: ensmallen_tests ..................Subprocess aborted***Exception: 2377.02 sec
ensmallen version: 2.18.0 (Fairmount Bagel)
armadillo version: 10.6.2 (Keep Calm)
invalid fastbin entry (free)

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ensmallen_tests is a Catch v2.13.7 host application.
[...]

PR #332, opened 2021-11-18
SCD partial gradient test fix

zoq:
Use the correct number of classes to generate the label vector. This will fix the build issue reported in https://github.com/mlpack/ensmallen/issues/329.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,6 +6,8 @@
  * Fix potential infinite loop in CMAES
    ([#331](https://github.com/mlpack/ensmallen/pull/331)).
 
+ * Fix SCD partial gradient test
+   ([#332](https://github.com/mlpack/ensmallen/pull/332)).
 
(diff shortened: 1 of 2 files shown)


HISTORY.md
```diff
@@ -6,6 +6,8 @@
  * Fix potential infinite loop in CMAES
    ([#331](https://github.com/mlpack/ensmallen/pull/331)).
 
+ * Fix Softmax Regression function partial gradient test lables vector
```
conradsnicta, line 9: "lables vector" -> "label vector"
or perhaps simplify the entire line to: "Fix SCD partial gradient test"

conradsnicta:
PS. it's probably worth releasing a patch release (2.18.1) after this fix is committed

rcurtin (approved):
Nice catch!  Thanks for putting the time into debugging this---it sounds like it was tedious to reproduce everything.  All that work for a two-character fix :smile:

conradsnicta (approved):

PR #335, opened 2022-01-18
RAdam Update Rule

shubham1206agra:
Added RAdam Update Rule to Adam optimizers.
References:
- https://pytorch.org/docs/stable/generated/torch.optim.RAdam.html#torch.optim.RAdam
- https://arxiv.org/abs/1908.03265

### DIFF
--- a/include/ensmallen_bits/adam/radam_update.hpp
+++ b/include/ensmallen_bits/adam/radam_update.hpp
@@ -126,12 +126,10 @@ class RAdamUpdate
 			
 			if (pt > 5.0)
 			{
-				const GradType lt = std::sqrt(biasCorrection2) / 
-						arma::sqrt(v + parent.epsilon);
 				const double rt = std::sqrt((pt - 4.0) * (pt - 2.0) * parent.pinf / 
 						((parent.pinf - 4.0) * (parent.pinf - 2.0) * pt));
-				iterate -= (stepSize * rt / biasCorrection1) *
-						m * lt;
+				iterate -= (stepSize * rt * std::sqrt(biasCorrection2) / biasCorrection1) *
+						m / (arma::sqrt(v) + parent.epsilon);
 			}
 			else
 			{
--- a/include/ensmallen_bits/adam/radam_update.hpp
+++ b/include/ensmallen_bits/adam/radam_update.hpp
@@ -1,0 +1,172 @@
+/**
+ * @file radam_update.hpp
+ * @author Shubham Agrawal
+ *
+ * RAdam update rule.
+ *
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_ADAM_RADAM_UPDATE_HPP
+#define ENSMALLEN_ADAM_RADAM_UPDATE_HPP
+
+namespace ens {
+
+/**
+ * For more information, see the following.
+ *
+ * @code
+ * @article{Liu2019,
+ *   author  = {Liyuan Liu, Haoming Jiang, Pengcheng He, 
+ * Weizhu Chen, Xiaodong Liu, Jianfeng Gao, Jiawei Han},
+ *   title   = {On the Variance of the Adaptive Learning Rate and Beyond},
+ *   journal = {ICLR},
+ *   year    = {2019},
+ *   url     = {https://arxiv.org/abs/1908.03265}
+ * }
+ * @endcode
+ */
+class RAdamUpdate
+{
+ public:
+  /**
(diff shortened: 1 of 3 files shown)


include/ensmallen_bits/adam/radam_update.hpp
```diff
@@ -0,0 +1,170 @@
...
+   public:
+    /**
+     * This constructor is called by the SGD Optimize() method before the start
+     * of the iteration update process.
+     *
+     * @param parent AdamUpdate object.
+     * @param rows Number of rows in the gradient matrix.
+     * @param cols Number of columns in the gradient matrix.
+     */
+    Policy(RAdamUpdate& parent, const size_t rows, const size_t cols) :
```
conradsnicta, line 92: For clarity, suggest to use `n_rows` and `n_cols` instead of `rows` and `cols`.  
This will also make it consistent with Armadillo: http://arma.sourceforge.net/docs.html#attributes

include/ensmallen_bits/adam/radam_update.hpp
```diff
@@ -0,0 +1,170 @@
...
+    {
+      // Increment the iteration counter variable.
+      ++parent.iteration;
+
+      // And update the iterate.
+      m *= parent.beta1;
+      m += (1 - parent.beta1) * gradient;
+
+      v *= parent.beta2;
+      v += (1 - parent.beta2) * (gradient % gradient);
```
conradsnicta, line 118: suggest to use `arma::square(gradient)` instead of `gradient % gradient`

tests/adam_test.cpp
```diff
@@ -217,6 +217,15 @@ TEST_CASE("PadamLogisticRegressionTest", "[AdamTest]")
   LogisticRegressionFunctionTest(optimizer, 0.003, 0.006);
 }
 
+/**
+ * Run RAdam on logistic regression and make sure the results are acceptable.
+ */
+TEST_CASE("RAdamLogisticRegressionTest", "[AdamTest]")
+{
+  RAdam optimizer;
+  LogisticRegressionFunctionTest(optimizer, 0.003, 0.006);
```
conradsnicta, line 226: might be useful to provide a brief comment on why these settings (0.003, 0.006) are used.

include/ensmallen_bits/adam/radam_update.hpp
zoq: For ensmallen we basically followed the mlpack style guidelines where we use `cols, rows` but I can see that this creates some confusion, so I don't mind to adapt the armadillo style here.

include/ensmallen_bits/adam/radam_update.hpp
zoq: Ohh nice idea, we use `gradient % gradient` at some other places as well, so we should adapt it in another points as well.

include/ensmallen_bits/adam/radam_update.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+      const double biasCorrection1 = 1.0 - std::pow(parent.beta1,
+          parent.iteration);
+      const double biasCorrection2 = 1.0 - std::pow(parent.beta2,
+          parent.iteration);
+			const double pt = parent.pinf - 2.0 * parent.iteration * 
+					(1.0 - biasCorrection2) / biasCorrection2;
+			
+			if (pt > 5.0)
```
zoq, line 127: The reference implementation uses `>= 5`, the paper uses `> 4`, so I would suggest to make this configurable and use the default from the reference implementation.

include/ensmallen_bits/adam/radam_update.hpp
```diff
@@ -0,0 +1,170 @@
...
+      m += (1 - parent.beta1) * gradient;
+
+      v *= parent.beta2;
+      v += (1 - parent.beta2) * (gradient % gradient);
+
+      const double biasCorrection1 = 1.0 - std::pow(parent.beta1,
+          parent.iteration);
+      const double biasCorrection2 = 1.0 - std::pow(parent.beta2,
+          parent.iteration);
+			const double pt = parent.pinf - 2.0 * parent.iteration *
```
zoq, line 124: Looks like the indentation is off in a couple of places, can you make sure it aligns with the rest of the codebase.

include/ensmallen_bits/adam/radam_update.hpp
rcurtin: I think we should stick with the mlpack style here, since ensmallen is part of the mlpack "umbrella".  It's good to have consistency across those projects (I know that this comes at the price of consistency with Armadillo).

rcurtin (commented):
I think everyone else has already left substantive comments about the implementation, but one issue is, how will users find it?  Should we create a new optimizer called `Radam` and document it like the rest?  That seems like the best idea to me, personally.  Take a look at the implementation of, e.g., `Nadam`, and the associated documentation in `doc/optimizers.md` and `doc/function_types.md` for an example. :+1:

Also, don't forget to add a changelog entry. :)

Issue #244, opened 2020-12-22
Epoch time does not show elapsed time for more than 59 seconds.


shrit:
Hello everyone,

I have just noticed a small issue in the estimated time callback. I am training a neural network on a large dataset, the training time for each epoch is taking more than one minute, (several minutes per epoch). 
My understanding is that at the end of each epoch the ETA callback prints the elapsed training time for each epoche. If this is the case, the printed time does not show the minutes elapsed but only shows the seconds and not the seconds after `59`. In the following, each epoch took at least 3 to 4 minutes.
```
9563.88
Epoch 1
5819/5819 [====================================================================================================] 100% - 18s 34ms/step - loss: 40.7373
Validation loss: 770.42
Epoch 2
5819/5819 [====================================================================================================] 100% - 10s 43ms/step - loss: 34.7851
Validation loss: 759.774
```
To reproduce the issue, use a large size dataset `> 1M` lines and train it on the CPU. In the end, you should see the elapsed time only in seconds.

zoq:
Thanks for the report, the issue is here:

https://github.com/mlpack/ensmallen/blob/cdfae670da608ef8498dee1e649f79f1d3279ef9/include/ensmallen_bits/callbacks/progress_bar.hpp#L221

I'll fix it.

zeph1yr:
Hello @zoq, Can I submit a pull request, looks like a good first issue. Thanks!

zoq:
> Hello @zoq, Can I submit a pull request, looks like a good first issue. Thanks!

Ahh, yes that would be awesome.

conradsnicta:
Should be solved by #337. Please reopen if it doesn't work properly.

PR #337, opened 2022-02-06
Report the correct epoch time in seconds.

zoq:
Fix for https://github.com/mlpack/ensmallen/issues/244 now it shows the correct time in seconds:

```
Epoch 1/2
3/3 [====================================================================================================] 100% - 210s 70000ms/step - loss: 1224.1
Epoch 2/2
2/3 [==================================================================>.................................] 66% - ETA: 10s - loss: 439.535
```

See the `210s` at the end of the first epoch, which is the sum of each step.

### DIFF
--- a/include/ensmallen_bits/callbacks/progress_bar.hpp
+++ b/include/ensmallen_bits/callbacks/progress_bar.hpp
@@ -216,9 +216,8 @@ class ProgressBar
         output << ".";
       }
     }
-
     const size_t stepTime = epochTimer.toc() / (double) epochSize * 1000;
-    output << "] " << progress << "% - " << (size_t) epochTimer.toc() % 60
+    output << "] " << progress << "% - " << epochTimer.toc()
         << "s " << stepTime << "ms/step " << "- loss: " << objective  <<  "\n";
     output.flush();
   }
--- a/include/ensmallen_bits/callbacks/progress_bar.hpp
+++ b/include/ensmallen_bits/callbacks/progress_bar.hpp
@@ -218,7 +218,7 @@ class ProgressBar
     }
     const size_t stepTime = epochTimer.toc() / (double) epochSize * 1000;
     output << "] " << progress << "% - " << epochTimer.toc()
-        << "s/epoch; " << stepTime << "ms/step; " << " loss: "
+        << "s/epoch; " << stepTime << "ms/step; " << "loss: "
         << objective  <<  "\n";
     output.flush();
   }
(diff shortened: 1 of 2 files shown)


include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -216,9 +216,8 @@ class ProgressBar
         output << ".";
       }
     }
-
     const size_t stepTime = epochTimer.toc() / (double) epochSize * 1000;
-    output << "] " << progress << "% - " << (size_t) epochTimer.toc() % 60
+    output << "] " << progress << "% - " << epochTimer.toc()
         << "s " << stepTime << "ms/step " << "- loss: " << objective  <<  "\n";
```
conradsnicta, line 221: Suggest to demarcate the various fields explicitly, so they appear more clearly to users. For example, change:  
`<< "s " << stepTime << "ms/step " << "- loss: " << objective  <<  "\n";`
to:  
`<< "s;  " << stepTime << "ms/step;  " << "- loss: " << objective  <<  "\n";`

zoq: Good idea, I'll adapt the format.

zoq:
Updated output:

```
Epoch 1/2
3/3 [====================================================================================================] 100% - 210s/epoch; 70000ms/step;  loss: 1224.1
Epoch 2/2
2/3 [==================================================================>.................................] 66% - ETA: 10s - loss: 439.535
```

include/ensmallen_bits/callbacks/progress_bar.hpp
```diff
@@ -216,10 +216,10 @@ class ProgressBar
         output << ".";
       }
     }
-
     const size_t stepTime = epochTimer.toc() / (double) epochSize * 1000;
-    output << "] " << progress << "% - " << (size_t) epochTimer.toc() % 60
-        << "s " << stepTime << "ms/step " << "- loss: " << objective  <<  "\n";
+    output << "] " << progress << "% - " << epochTimer.toc()
```
conradsnicta, line 220: There's a subtle bug here.  `epochTimer.toc()` is called twice: here and on the preceding line.  Note that the timer is still running between the two calls to `.toc()`, so it's entirely possible that the second call to `epochTimer.toc()` will provide a different value than the first call to `epochTimer.toc()`.  A lot of stuff may get executed between the two calls, especially when streams are involved (ie. the `output << ...` part).

To fix this, need to store the value of `epochTimer.toc()` into a separate variable and use that in further calculations.  For example:
```
  const double epochTimer_elapsed = epochTimer.toc();
  const size_t stepTime = epochTimer_elapsed / (double) epochSize * 1000;
  output << "] " << progress << "% - " << epochTimer_elapsed
```

Ref: docs for [wall_clock](http://arma.sourceforge.net/docs.html#wall_clock)

conradsnicta (approved):

PR #340, opened 2022-03-23
Move `iteration` member into UpdatePolicy classes

rcurtin:
While testing some RNN support, I noticed that using `Adam` twice with `resetPolicy = true` did not produce the same results:

```
  ens::Adam optimizer(1e-5, 1, 0.9, 0.999, 1e-8, 200, 1e-8, false);

  ffn.Train(data.slice(0), responses.slice(0), optimizer);
  rnn.Train(data, responses, optimizer);
```

In this particular test case, that *should* produce the exact same results for `ffn` and `rnn`.  (It's easy to make an example that breaks in the same way.)

As it turns out, the issue is that even when `resetPolicy` is `true`, there is one member of `AdamUpdate` that doesn't get reset: the `iteration` counter.  Realistically, that should not be in the `AdamUpdate` class since it is specific to an optimization, but should instead be in the `AdamUpdate::Policy` class, which holds information specific to an individual optimization.

So, I went through all update policies with that pattern, and moved the `iteration` member into the child `Policy` class.

### DIFF
--- a/include/ensmallen_bits/ada_belief/ada_belief_update.hpp
+++ b/include/ensmallen_bits/ada_belief/ada_belief_update.hpp
@@ -49,8 +49,7 @@ class AdaBeliefUpdate
                   const double beta2 = 0.999) :
     epsilon(epsilon),
     beta1(beta1),
-    beta2(beta2),
-    iteration(0)
+    beta2(beta2)
   {
     // Nothing to do.
   }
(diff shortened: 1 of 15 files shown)


zoq:
Restarted the windows build to check if the failing FTML test was just a fluke.

rcurtin:
I also fixed a couple PVS-Studio issues, but one of the ones listed didn't make sense.  Let's see if it's any happier this time...

zoq (commented):
I think it would be a good idea to add a simple test to one optimizer, e.g. something simple like:

```c++
  Adam optimizer(0.5, 2, 0.7, 0.999, 1e-8, 50, 1e-3, false);
  optimizer.ResetPolicy() = true;

  SphereFunction f(2);

  arma::mat coordinatesA = f.GetInitialPoint();
  optimizer.Optimize(f, coordinatesA);

  arma::mat coordinatesB = f.GetInitialPoint();
  optimizer.Optimize(f, coordinatesB);

  CheckMatrices(coordinatesA, coordinatesB);
```

rcurtin:
Nice idea!  I added the test in ab4481a3.

zoq (approved):
I think, the remaining static check issues are false alarms, I'll add them to the ignore list.

PR #342, opened 2022-04-05
bump check for min supported version of armadillo

conradsnicta:
followup to #318

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -11,6 +11,9 @@
 
  * Add AdaSqrt optimizer
    ([#234](https://github.com/mlpack/ensmallen/pull/234)).
+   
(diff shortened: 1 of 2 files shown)


zoq (approved):
Looks good to me, let's do another release once this is merged.

rcurtin (approved):
Ahh, nice catch, I guess we forgot to update that check. :+1:

coatless:
Always nice to see a willingness to bump minimum dependencies. Though, I think this is the second in under a year or two. Is this related to an upstream change in MLPACK?

conradsnicta:
@coatless The bump was way back in 2021 via #318 (seems like a century ago). Unfortunately we missed updating the actual check for the version. This PR merely fixes the omission from 2021.

PR #344, opened 2022-05-19
Implementation of DTLZ1 Test Function on NSGA2 Optimizer

ahmedr2001:


zoq (commented):
Let me know if you need some help to figure out why the build fails.

ahmedr2001:
As in why the checks fail? Then yes, because it builds on my machine and passes the tests.

PR #347, opened 2022-06-28
avoid deprecation warnings in armadillo 11.2+

conradsnicta:
`ARMA_DONT_PRINT_ERRORS` was originally added to prevent runtime warnings coming from `arma::eig_sym()` within the `CheckPositiveSemiDefinite()` function within tests/sdp_primal_dual_test.cpp.

Since Armadillo 10.4, `ARMA_DONT_PRINT_ERRORS` doesn't do anything, as failed decompositions no longer emit warnings by default.  (The warning level can be adjusted via `ARMA_WARN_LEVEL`.)

Since Armadillo 11.2+, a compile-time deprecation notice is emitted if `ARMA_DONT_PRINT_ERRORS` is defined.

Since Armadillo 11.2+, a compile-time deprecation notice is emitted if `operator<<` is used for matrix and vector initialisation.  The much more efficient C++11 initializer_lists can be used instead.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,7 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
+* Avoid deprecation warning in Armadillo 11.2+
+   ([#347](https://github.com/mlpack/ensmallen/pull/347)).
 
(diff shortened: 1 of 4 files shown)


zoq (approved):
No comments from my side.

Issue #72, opened 2019-01-19
SDP should have a more general linear operator constraint


rcurtin:
This is a port of mlpack/mlpack#401.  There are some more details there.  This is the original description from @stephentu:

> Our SDP abstraction right now only supports sparse and dense matrices. This is sufficiently general, but for some certain classes of constraints, quite suboptimal.
>
> As an example, suppose I want to have the constraint that the sum of all elements in the matrix is constant. While we can express this as Tr((11^T) * X) = b, this is quite inefficient to evaluate for something that could be expressed as X.sum().
>
> Hence, it would be desirable to have a third type of constraint, "linearOperator", which is just a linear function which takes symmetric matrices to real numbers. The straightforward way to implement this would be a vector of std::function objects.

PR #350, opened 2022-12-10
Add more general linear operator constraint to sdp

SuvarshaChennareddy:
Hello everyone, 
This is related to issue #72. I've allowed the ability to store constraints as `std::function` objects.
I have not modified/added tests yet. However, if this PR looks alright, I'll be willing to do so.
Any comments would be greatly appreciated.

### DIFF
--- a/include/ensmallen_bits/sdp/lin_alg.hpp
+++ b/include/ensmallen_bits/sdp/lin_alg.hpp
@@ -15,6 +15,38 @@
 namespace ens {
 namespace math {
 
+/**
+* Matrix representation of a general Linear operator representated 
+* by a function/method.
+*
+* @param input Input function.
+* @param output The matrix representation.
+*/
+template<typename ElemType, typename MatBType>
+inline void convertToMatrix(std::function<ElemType(arma::Mat<ElemType>)> input, 
+                            MatBType& output)
+{
+    const size_t n = output.n_rows;
+
+    arma::Mat<ElemType> tempInputCoordinates;
+    tempInputCoordinates.zeros(n, n);
+
+    size_t idx = 0;
+    for (size_t i = 0; i < n; i++)
+    {
+        if (i > 0) tempInputCoordinates(i - 1, n - 1) = 0;
+        for (size_t j = i; j < n; j++)
+        {
+            if (j > i) tempInputCoordinates(i, j - 1) = 0;
+            tempInputCoordinates(i, j) = 1;
+            if (i == j)
+                output(i, j) = input(tempInputCoordinates);
+            else
+                output(i, j) = output(j, i) = 0.5 * arma::datum::sqrt2 * input(tempInputCoordinates);
+        }
+    }
+}
+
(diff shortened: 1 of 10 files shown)


include/ensmallen_bits/sdp/lin_alg.hpp
```diff
@@ -109,6 +141,8 @@ inline void Smat(const MatAType& input, MatBType& output)
   }
 }
 
+
+
```
rcurtin, line 145: No need to add this whitespace :)

include/ensmallen_bits/sdp/lrsdp_function.hpp
```diff
@@ -118,6 +120,12 @@ class LRSDPFunction
     return rrt.As<typename std::remove_reference<MatType>::type>();
   }
 
+  // Cache constarint matrices of the general linear operators
```
rcurtin, line 123:
```suggestion
  // Cache constraint matrices of the general linear operators 
```

include/ensmallen_bits/sdp/lin_alg.hpp
```diff
@@ -15,6 +15,38 @@
...
+/**
+* Matrix representation of a general Linear operator representated 
+* by a function/method.
+*
+* @param input Input function.
+* @param output The matrix representation.
+*/
+template<typename ElemType, typename MatBType>
+inline void convertToMatrix(std::function<ElemType(arma::Mat<ElemType>)> input, 
+                            MatBType& output)
```
rcurtin, line 27: It looks like the approach you are taking is to turn a `std::function` that the user gives into a matrix, and then use that as a constraint.  But this is the opposite of what we are looking for in #72: take the given example of the constraint that the sum of elements is a constant.  You can express this very tediously as the constraint `trace(1 1^T X) = b`, but the much easier way to represent this is `sum(X) = b`.  So, instead of extracting the matrix `1 1^T`, it would be better to just store the `std::function`s in the SDP objects and evaluate them directly to get the value.

rcurtin (commented):
Hey @SuvarshaChennareddy, thanks for taking the time to look into this.  I have some basic comments before you proceed further.  I think that tests are definitely necessary both to ensure that the implementation is correct, and also that it might behave like a user expects.

If I were you, I would start by assembling a couple reasonable examples of how a user might want to express a constraint in this way, and then designing to make those cases easy. :+1:

include/ensmallen_bits/sdp/lin_alg.hpp
SuvarshaChennareddy: First of all, thank you for taking the time to look into this. I'm glad you asked this. I honestly didn't know if this would be alright.

However, (from what I understand) in the interior point method, the constraints need to be expressed as matrices (as matrix multiplication and transposing is involved).

Besides this, in the LRSDP implementation, I have used these `std::function`s wherever the matrix conversion isn't necessary (again like in the case of matrix multiplication or transposing).

SuvarshaChennareddy:
.

PR #351, opened 2023-01-02
An attempt to fix the current CMAES inconsistencies

SuvarshaChennareddy:
I've attempted to fix the inconsistencies listed out in issue #70 with a couple of other things:
1. I noticed that this implementation of CMAES does not handle parameters (decision variables) that are represented differently, in a similar fashion. More specifically, parameters that are represented as a column vector are treated differently from the SAME parameters that are represented as a row vector (with appropriate changes in the fitness/loss function of course).  That is to say, if there existed `P_row` and `P_col` (which are equivalent except in the form that they are represented in) and `L_row` and `L_col` (which are respective fitness functions where `L_row(P_row) =  L_col(P_col)`), they do not converge to the same result with the current implementation of CMAES. This wasn't very difficult to solve considering there was a transposed space (for example, row-wise representation) along with the space (column wise representation) itself.
2. Another error I noticed was the update step for `p_sigma`. Here it was considered that `C^(-1/2)`, where C is the covariance matrix, was equivalent to `L^T` where L is the matrix satisfying `C = L*L^T` (Cholesky  Decomposition). I'm not sure the way I fixed it is the best way to go about it (but I guess it works for now?).
3. I added Transformation policies (Boundary Box Constraint and Empty Transformation). I used Pull Request #193 by @gaurav-singh1998 as a foundation.

### DIFF
--- a/include/ensmallen_bits/cmaes/cmaes.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes.hpp
@@ -17,6 +17,8 @@
 
 #include "full_selection.hpp"
 #include "random_selection.hpp"
+#include "transformation_policies/empty_transformation.hpp"
+#include "transformation_policies/boundary_box_constraint.hpp"
 
 namespace ens {
 
@@ -46,8 +48,11 @@ namespace ens {
  * ensmallen website.
  *
  * @tparam SelectionPolicy The selection strategy used for the evaluation step.
+ * @tparam transformationPolicy The transformation strategy used to 
+ *       map cooridnates to the desired domain.
  */
-template<typename SelectionPolicyType = FullSelection>
+template<typename SelectionPolicyType = FullSelection,
+         typename TransformationPolicyType = EmptyTransformation<>>
 class CMAES
 {
  public:
@@ -60,8 +65,8 @@ class CMAES
    * equal one pass over the dataset).
    *
    * @param lambda The population size (0 use the default size).
-   * @param lowerBound Lower bound of decision variables.
-   * @param upperBound Upper bound of decision variables.
+   * @param transformationPolicy Instantiated transformation policy used to 
+   *     map the cooridnates to the desired domain.
    * @param batchSize Batch size to use for the objective calculation.
    * @param maxIterations Maximum number of iterations allowed (0 means no
    *     limit).
@@ -70,8 +75,8 @@ class CMAES
    *     objective.
    */
   CMAES(const size_t lambda = 0,
-        const double lowerBound = -10,
-        const double upperBound = 10,
+        const TransformationPolicyType& 
+              transformationPolicy = TransformationPolicyType(),
         const size_t batchSize = 32,
         const size_t maxIterations = 1000,
         const double tolerance = 1e-5,
@@ -87,30 +92,22 @@ class CMAES
    * @tparam CallbackTypes Types of callback functions.
    * @param function Function to optimize.
    * @param iterate Starting point (will be modified).
+   * @param stepSize Starting sigma/step size (will be modified).
    * @param callbacks Callback functions.
    * @return Objective value of the final point.
    */
   template<typename SeparableFunctionType,
-           typename MatType,
-           typename... CallbackTypes>
-  typename MatType::elem_type Optimize(SeparableFunctionType& function,
-                                       MatType& iterate,
-                                       CallbackTypes&&... callbacks);
+      typename MatType,
+      typename... CallbackTypes>
+      typename MatType::elem_type Optimize(SeparableFunctionType& function,
+          MatType& iterate,
+          double stepSize = 0.6,
+          CallbackTypes&&... callbacks);
 
   //! Get the population size.
   size_t PopulationSize() const { return lambda; }
   //! Modify the population size.
   size_t& PopulationSize() { return lambda; }
-
-  //! Get the lower bound of decision variables.
-  double LowerBound() const { return lowerBound; }
-  //! Modify the lower bound of decision variables.
-  double& LowerBound() { return lowerBound; }
-
-  //! Get the upper bound of decision variables
-  double UpperBound() const { return upperBound; }
-  //! Modify the upper bound of decision variables
-  double& UpperBound() { return upperBound; }
 
   //! Get the batch size.
   size_t BatchSize() const { return batchSize; }
@@ -132,15 +129,16 @@ class CMAES
   //! Modify the selection policy.
   SelectionPolicyType& SelectionPolicy() { return selectionPolicy; }
 
+  //! Get the transformation policy.
+  const TransformationPolicyType& TransformationPolicy() const
+  { return transformationPolicy; }
+  //! Modify the transformation policy.
+  TransformationPolicyType& TransformationPolicy() 
+  { return transformationPolicy; }
+
  private:
   //! Population size.
   size_t lambda;
-
-  //! Lower bound of decision variables.
-  double lowerBound;
-
-  //! Upper bound of decision variables
-  double upperBound;
 
   //! The batch size for processing.
   size_t batchSize;
@@ -153,13 +151,16 @@ class CMAES
 
   //! The selection policy used to calculate the objective.
   SelectionPolicyType selectionPolicy;
+
+  TransformationPolicyType transformationPolicy;
 };
 
 /**
  * Convenient typedef for CMAES approximation.
  */
-template<typename SelectionPolicyType = RandomSelection>
-using ApproxCMAES = CMAES<SelectionPolicyType>;
+template<typename TransformationPolicyType = EmptyTransformation<>,
+         typename SelectionPolicyType = RandomSelection>
+using ApproxCMAES = CMAES<SelectionPolicyType, TransformationPolicyType>;
 
 } // namespace ens
 
--- a/include/ensmallen_bits/cmaes/cmaes.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes.hpp
@@ -194,6 +194,7 @@ class CMAES
   //! has completed.
   TransformationPolicyType transformationPolicy;
 
+  //! The step size.
   double stepSize;
 };
 
--- a/include/ensmallen_bits/cmaes/cmaes.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes.hpp
@@ -105,14 +105,14 @@ class CMAES
    * objective.
    * @param stepSize Starting sigma/step size (will be modified).
    */
-  CMAES(const size_t lambda = 0,
-        const double lowerBound = -10,
-        const double upperBound = 10,
-        const size_t batchSize = 32,
-        const size_t maxIterations = 1000,
-        const double tolerance = 1e-5,
-        const SelectionPolicyType & selectionPolicy = SelectionPolicyType(),
-        double stepSize = 1);
+  ens_deprecated CMAES(const size_t lambda = 0,
+                       const double lowerBound = -10,
+                       const double upperBound = 10,
+                       const size_t batchSize = 32,
+                       const size_t maxIterations = 1000,
+                       const double tolerance = 1e-5,
+                       const SelectionPolicyType& selectionPolicy = SelectionPolicyType(),
+                       double stepSize = 1);
 
   /**
    * Optimize the given function using CMA-ES. The given starting point will be
@@ -190,7 +190,8 @@ class CMAES
   SelectionPolicyType selectionPolicy;
 
   //! The transformationPolicy used to map coordinates to the suitable domain
-  //! while evaluating fitness. 
+  //! while evaluating fitness. This mapping is also done after optimization 
+  //! has completed.
   TransformationPolicyType transformationPolicy;
 
   double stepSize;
--- a/include/ensmallen_bits/cmaes/cmaes.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes.hpp
@@ -48,8 +48,9 @@ namespace ens {
  * ensmallen website.
  *
  * @tparam SelectionPolicy The selection strategy used for the evaluation step.
- * @tparam transformationPolicy The transformation strategy used to 
- *       map cooridnates to the desired domain.
+ * @tparam TransformationPolicy The transformation strategy used to 
+ *       map decision variables to the desired domain during fitness evaluation
+ *       and termination. Use EmptyTransformation if the domain isn't bounded.
  */
 template<typename SelectionPolicyType = FullSelection,
          typename TransformationPolicyType = EmptyTransformation<>>
@@ -66,13 +67,14 @@ class CMAES
    *
    * @param lambda The population size (0 use the default size).
    * @param transformationPolicy Instantiated transformation policy used to 
-   *     map the cooridnates to the desired domain.
+   *     map the coordinates to the desired domain.
    * @param batchSize Batch size to use for the objective calculation.
    * @param maxIterations Maximum number of iterations allowed (0 means no
    *     limit).
    * @param tolerance Maximum absolute tolerance to terminate algorithm.
    * @param selectionPolicy Instantiated selection policy used to calculate the
    *     objective.
+   * @param stepSize Starting sigma/step size (will be modified).
    */
   CMAES(const size_t lambda = 0,
         const TransformationPolicyType& 
@@ -80,7 +82,37 @@ class CMAES
         const size_t batchSize = 32,
         const size_t maxIterations = 1000,
         const double tolerance = 1e-5,
-        const SelectionPolicyType& selectionPolicy = SelectionPolicyType());
+        const SelectionPolicyType& selectionPolicy = SelectionPolicyType(),
+        double stepSize = 0);
+
+  /**
+   * Construct the CMA-ES optimizer with the given function and parameters 
+   * (including lower and upper bounds). The defaults here are not necessarily 
+   * good for the given problem, so it is suggested that the values used be 
+   * tailored to the task at hand.  The maximum number of iterations refers to 
+   * the maximum number of points that are processed (i.e., one iteration 
+   * equals one point; one iteration does not equal one pass over the dataset). 
+   * This constructor is deprecated.
+   * 
+   * @param lambda The population size(0 use the default size).
+   * @param lowerBound Lower bound of decision variables.
+   * @param upperBound Upper bound of decision variables.
+   * @param batchSize Batch size to use for the objective calculation.
+   * @param maxIterations Maximum number of iterations allowed(0 means no
+      limit).
+   * @param tolerance Maximum absolute tolerance to terminate algorithm.
+   * @param selectionPolicy Instantiated selection policy used to calculate the
+   * objective.
+   * @param stepSize Starting sigma/step size (will be modified).
+   */
+  CMAES(const size_t lambda = 0,
+        const double lowerBound = -10,
+        const double upperBound = 10,
+        const size_t batchSize = 32,
+        const size_t maxIterations = 1000,
+        const double tolerance = 1e-5,
+        const SelectionPolicyType & selectionPolicy = SelectionPolicyType(),
+        double stepSize = 1);
 
   /**
    * Optimize the given function using CMA-ES. The given starting point will be
@@ -92,7 +124,6 @@ class CMAES
    * @tparam CallbackTypes Types of callback functions.
    * @param function Function to optimize.
    * @param iterate Starting point (will be modified).
-   * @param stepSize Starting sigma/step size (will be modified).
    * @param callbacks Callback functions.
    * @return Objective value of the final point.
    */
@@ -101,7 +132,6 @@ class CMAES
       typename... CallbackTypes>
       typename MatType::elem_type Optimize(SeparableFunctionType& function,
           MatType& iterate,
-          double stepSize = 0.6,
           CallbackTypes&&... callbacks);
 
   //! Get the population size.
@@ -136,6 +166,13 @@ class CMAES
   TransformationPolicyType& TransformationPolicy() 
   { return transformationPolicy; }
 
+  //! Get the step size.
+  double StepSize() const
+  { return stepSize; }
+  //! Modify the step size.
+  double& StepSize()
+  { return stepSize; }
+
  private:
   //! Population size.
   size_t lambda;
@@ -152,7 +189,11 @@ class CMAES
   //! The selection policy used to calculate the objective.
   SelectionPolicyType selectionPolicy;
 
+  //! The transformationPolicy used to map coordinates to the suitable domain
+  //! while evaluating fitness. 
   TransformationPolicyType transformationPolicy;
+
+  double stepSize;
 };
 
 /**
--- a/include/ensmallen_bits/cmaes/cmaes_impl.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes_impl.hpp
@@ -24,6 +24,8 @@
  This partial specialization is used to throw an exception when the
  TransformationPolicyType is EmptyTransformation and call a
  constructor with paramters 'lowerBound' and 'upperBound' otherwise.
+ This shall be removed when the deprecated constructor is removed in
+ the next major version of ensmallen.
 */
 template<typename T1, typename T2>
 struct notEmptyTransformation : std::true_type {
--- a/include/ensmallen_bits/cmaes/cmaes_impl.hpp
+++ b/include/ensmallen_bits/cmaes/cmaes_impl.hpp
@@ -200,6 +200,12 @@ typename MatType::elem_type CMAES<SelectionPolicyType,
   // Now iterate!
   terminate |= Callback::BeginOptimization(*this, function, 
       transformedIterate, callbacks...);
+
+  // The number of generations to wait after the minimum loss has
+  // been reached or no improvement has been made before terminating.
+  size_t patience = 10 + (30 * iterate.n_elem / lambda) + 1;
+  size_t steps = 0;
+
   for (size_t i = 1; (i != maxIterations) && !terminate; ++i)
   {
     // To keep track of where we are.
@@ -219,19 +225,19 @@ typename MatType::elem_type CMAES<SelectionPolicyType,
       if (iterate.n_rows > iterate.n_cols)
       {
         pStep[idx(j)] = covLower *
-            arma::randn<BaseMatType>(iterate.n_rows, iterate.n_cols);
+          arma::randn<BaseMatType>(iterate.n_rows, iterate.n_cols);
       }
       else
       {
         pStep[idx(j)] = arma::randn<BaseMatType>(iterate.n_rows, iterate.n_cols)
-            * covLower.t();
+          * covLower.t();
       }
 
       pPosition[idx(j)] = mPosition[idx0] + sigma(idx0) * pStep[idx(j)];
 
       // Calculate the objective function.
       pObjective(idx(j)) = selectionPolicy.Select(function, batchSize,
-          transformationPolicy.Transform(pPosition[idx(j)]), callbacks...);
+        transformationPolicy.Transform(pPosition[idx(j)]), callbacks...);
     }
 
     // Sort population.
@@ -245,7 +251,7 @@ typename MatType::elem_type CMAES<SelectionPolicyType,
 
     // Calculate the objective function.
     currentObjective = selectionPolicy.Select(function, batchSize,
-        transformationPolicy.Transform(mPosition[idx1]), callbacks...);
+      transformationPolicy.Transform(mPosition[idx1]), callbacks...);
 
     // Update best parameters.
     if (currentObjective < overallObjective)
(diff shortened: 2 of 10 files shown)


SuvarshaChennareddy:
I haven't updated `callbacks_test.cpp` yet. I'm not sure if I should take the initial step size as a parameter in `Optimize(...)`.

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -60,8 +65,8 @@ class CMAES
    * equal one pass over the dataset).
    *
    * @param lambda The population size (0 use the default size).
-   * @param lowerBound Lower bound of decision variables.
-   * @param upperBound Upper bound of decision variables.
+   * @param transformationPolicy Instantiated transformation policy used to 
+   *     map the cooridnates to the desired domain.
```
rcurtin, line 69:
```suggestion
   *     map the coordinates to the desired domain.
```

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -46,8 +48,11 @@ namespace ens {
  * ensmallen website.
  *
  * @tparam SelectionPolicy The selection strategy used for the evaluation step.
+ * @tparam transformationPolicy The transformation strategy used to 
+ *       map cooridnates to the desired domain.
```
rcurtin, line 52:
```suggestion
 * @tparam TransformationPolicy The transformation strategy used to 
 *       map coordinates to the desired domain.
```

Do you think you can provide a little more documentation on what is expected of a `TransformationPolicyType`?

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -70,8 +75,8 @@ class CMAES
    *     objective.
    */
   CMAES(const size_t lambda = 0,
-        const double lowerBound = -10,
-        const double upperBound = 10,
+        const TransformationPolicyType& 
+              transformationPolicy = TransformationPolicyType(),
```
rcurtin, line 79: This breaks reverse compatibility---if the default transformation policy type is the bounding box (which takes parameters `lowerBound` and `upperBound` anyway), one option could be to preserve a deprecated constructor that takes `lowerBound` and `upperBound` and throws an exception if `TransformationPolicyType` is not `BoundaryBoxConstraint`.  We could remove that deprecated constructor for ensmallen 3.x, but for now we would need to preserve it so we satisfy the guarantees in `UPDATING.txt`.

We would also need to update the documentation to reflect the new API.

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -87,31 +92,23 @@ class CMAES
...
-           typename MatType,
-           typename... CallbackTypes>
-  typename MatType::elem_type Optimize(SeparableFunctionType& function,
-                                       MatType& iterate,
-                                       CallbackTypes&&... callbacks);
+      typename MatType,
+      typename... CallbackTypes>
+      typename MatType::elem_type Optimize(SeparableFunctionType& function,
+          MatType& iterate,
+          double stepSize = 0.6,
```
rcurtin, line 104: Parameters like the step size are usually parameters to the constructor itself---do you want to move that there, then add `StepSize()` functions to access and change it?

rcurtin: Also, might be worth using a step size of `0.0` (the default) for "automatic", which can use the existing behavior of `0.3 * (upperBound - lowerBound)`.  (I guess you'd have to create some "default step size" function for each `TransformationPolicyType`.)

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -153,13 +151,16 @@ class CMAES
 
   //! The selection policy used to calculate the objective.
   SelectionPolicyType selectionPolicy;
+
+  TransformationPolicyType transformationPolicy;
```
rcurtin, line 155: Want to add a comment here on what this is for?  Maybe

```
//! The transformation policy maps candidate points back into a suitable domain.
```

but I am not sure if that fully reflects your intention, it's just a suggestion.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -99,8 +100,7 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
 
   std::vector<BaseMatType> mPosition(2, BaseMatType(iterate.n_rows,
       iterate.n_cols));
-  mPosition[0] = lowerBound + arma::randu<BaseMatType>(
-      iterate.n_rows, iterate.n_cols) * (upperBound - lowerBound);
+  mPosition[0] = iterate;
```
rcurtin, line 103: Should we apply the transformation policy here, just in case the initial iterate is outside the bounds?

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -110,11 +110,13 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
     const size_t effectiveBatchSize = std::min(batchSize, numFunctions - f);
-    const ElemType objective = function.Evaluate(mPosition[0], f,
+    const ElemType objective = 
+        function.Evaluate(transformationPolicy.Transform(mPosition[0]), f,
         effectiveBatchSize);
     currentObjective += objective;
 
-    Callback::Evaluate(*this, function, mPosition[0], objective,
+    Callback::Evaluate(*this, function, 
+      transformationPolicy.Transform(mPosition[0]), objective,
```
rcurtin, line 119:
```suggestion
        transformationPolicy.Transform(mPosition[0]), objective,
```
Minor style fix: wrapped lines should be doubly indented.

rcurtin: Also worth pointing out is that you call `transformationPolicy.Transform()` twice here for each inner iteration of the loop.  Does it perhaps make more sense to call `Transform()` once, before the loop?  That may require some further refactoring, but it seems to me to be conceptually simpler to simply use `Transform()` to map each new proposed iterate back into the required range right after a step is taken, only once.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -171,14 +177,14 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
       else
       {
         pStep[idx(j)] = arma::randn<BaseMatType>(iterate.n_rows, iterate.n_cols)
-            * covLower;
+            * covLower.t();
```
rcurtin, line 180: Can you explain why this change is needed?  It isn't immediately obvious to me.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -192,27 +198,31 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
 
     // Calculate the objective function.
     currentObjective = selectionPolicy.Select(function, batchSize,
-        mPosition[idx1], callbacks...);
+      transformationPolicy.Transform(mPosition[idx1]), callbacks...);
```
rcurtin, line 201:
```suggestion
        transformationPolicy.Transform(mPosition[idx1]), callbacks...);
```
(small style fix)

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
```diff
@@ -0,0 +1,166 @@
...
+ *
+ *
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+
+namespace ens {
+  /**
```
rcurtin, line 15:
```suggestion
namespace ens {

  /**
```
Some more style pedantry.  Also, it seems like this comment is indented when it doesn't need to be?

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
```diff
@@ -0,0 +1,166 @@
...
+  * Construct the boundary box constraint poilcy.
+  *
+  * @param lowerBound The lower bound of the coordinates.
+  * @param upperBound The upper bound of the coordinates.
+  */
+  BoundaryBoxConstraint(const MatType& lowerBound,
+    const MatType& upperBound):
+    lowerBound(lowerBound),
+    upperBound(upperBound)
```
rcurtin, line 46:
```suggestion
  BoundaryBoxConstraint(const MatType& lowerBound,
                        const MatType& upperBound) :
      lowerBound(lowerBound),
      upperBound(upperBound)
```

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
```diff
@@ -0,0 +1,166 @@
...
+  {}
+  
+  /**
+  * Map the given coordinates to the range 
+  * [lowerBound, upperBound]
+  *
+  * @param x Given coordinates.
+  * @return Transformed coordinates.
```
rcurtin, line 54: It seems to me like it make more sense to do this as an in-place transformation, so `x` is simply modified and the function returns `void`.  What do you think?  Did I overlook something?

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
```diff
@@ -0,0 +1,166 @@
...
+{
+public:
+  /**
+  * Construct the boundary box constraint poilcy.
+  *
+  * @param lowerBound The lower bound of the coordinates.
+  * @param upperBound The upper bound of the coordinates.
+  */
+  BoundaryBoxConstraint(const MatType& lowerBound,
+    const MatType& upperBound):
```
rcurtin, line 44: So, here you are allowing a different lower and upper bound for each element of the input?  My opinion is we should also provide a `double` constructor that reflects the previous functionality (so every dimension's limits are `[lowerBound, upperBound]`).

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
```diff
@@ -0,0 +1,166 @@
...
+  
+  /**
+  * Map the given coordinates to the range 
+  * [lowerBound, upperBound]
+  *
+  * @param x Given coordinates.
+  * @return Transformed coordinates.
+  */
+  MatType Transform(const MatType& x)
+  {
```
rcurtin, line 57: It might make sense as a sanity check to ensure that `lowerBound` and `upperBound` have the same size as `x`.

tests/cmaes_test.cpp
```diff
@@ -53,6 +65,10 @@ TEST_CASE("CMAESLogisticRegressionFMatTest", "[CMAESTest]")
  */
 TEST_CASE("ApproxCMAESLogisticRegressionFMatTest", "[CMAESTest]")
 {
-  ApproxCMAES<> cmaes(0, -1, 1, 32, 200, 1e-3);
+  arma::fmat lb(1, 1, arma::fill::value(-1));
+  arma::fmat ub(1, 1, arma::fill::value(1));
+  BoundaryBoxConstraint b(lb, ub);
+
+  ApproxCMAES<BoundaryBoxConstraint<arma::fmat>> cmaes(0, b, 32, 200, 1e-3);
```
rcurtin, line 72: When you add the deprecated constructor, we should at least add one test that uses the deprecated constructor (and we can point out in a comment that the test itself can be removed when ensmallen 3.x is released, as the constructor will go away).

tests/cmaes_test.cpp
```diff
@@ -23,7 +23,11 @@ using namespace ens::test;
  */
 TEST_CASE("CMAESLogisticRegressionTest", "[CMAESTest]")
 {
-  CMAES<> cmaes(0, -1, 1, 32, 200, 1e-3);
+  arma::mat lb(1, 1, arma::fill::value(-1));
+  arma::mat ub(1, 1, arma::fill::value(1));
+  BoundaryBoxConstraint b(lb, ub);
+
+  CMAES cmaes(0, b, 32, 200, 1e-3);
```
rcurtin, line 30: A test for `EmptyTransformation` would be a good idea too. :+1:

include/ensmallen_bits/cmaes/transformation_policies/empty_transformation.hpp
```diff
@@ -0,0 +1,45 @@
...
+/**
+* Inverse transformtion (effectively no transformation).
+*
+* @param y Input coordinates.
+* @return Transformed coordinates (the coordinates themselves).
+*/
+MatType Inverse(const MatType& y) { return y; }
+};
+
+} // namespace ens
```
rcurtin, line 45: Mind adding a newline at the end of the file? :)

include/ensmallen_bits/cmaes/transformation_policies/empty_transformation.hpp
```diff
@@ -0,0 +1,45 @@
+/**
+ * @file empty_transformation.hpp
+ * @author Suvarsha Chennareddy
+ *
+ * Empty Transformation, can also be called an Indentiy Transformation.
+ *
```
rcurtin, line 6:
```suggestion
 * Empty Transformation, can also be called an Identity Transformation.
```
:)

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
```diff
@@ -0,0 +1,166 @@
...
+  * In a particular case, the domain of a specific function is restricted 
+  * by boundaries.
+  * The implemented transformation transforms given coordinates into a region 
+  * bounded by the given lower and upper bounds (a box). First, the 
+  * coordinates are shifted into a feasible preimage bounded by lower_bound - al 
+  * and upper_bound + au where al and au and calculated internally. 
+  * These shifted coordinates are then transformed into coordinates bounded by 
+  * lower_bound and upper_bound. It is an identity transformation in between 
+  * the lower and upper bounds.
+  * The inverse transformation transforms the coordinates back to the preimage.
```
rcurtin, line 26: The inverse transformation is never used in these changes---do you intend to use it, or should we remove the implementation?  (Not a great idea to keep unused code)

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -192,27 +198,31 @@ typename MatType::elem_type CMAES<SelectionPolicyType>::Optimize(
...
+      transformedIterate = transformationPolicy.Transform(iterate);
+      terminate |= Callback::StepTaken(*this, function, 
+        transformedIterate, callbacks...);
     }
 
     // Update Step Size.
     if (iterate.n_rows > iterate.n_cols)
     {
       ps[idx1] = (1 - cs) * ps[idx0] + std::sqrt(
-          cs * (2 - cs) * muEffective) * covLower.t() * step;
```
rcurtin, line 210: If the goal is the compute `C^{-1/2}`, it should suffice to use `inv(covLower)`: https://math.stackexchange.com/questions/1230051/inverse-square-root-of-matrix

I think Armadillo will solve the system more quickly too if you specify `trimatl()` (since it is a lower triangular matrix): `inv(trimatl(covLower))`.

I haven't checked this exactly, but it should at least be in the right direction.  You should verify my comment here instead of trusting it to be correct... :smile:

rcurtin (commented):
Thanks @SuvarshaChennareddy for looking into these things!  I took a look through and left a bunch of comments. :)

> I noticed that this implementation of CMAES does not handle parameters (decision variables) that are represented differently, in a similar fashion. More specifically, parameters that are represented as a column vector are treated differently from the SAME parameters that are represented as a row vector (with appropriate changes in the fitness/loss function of course). That is to say, if there existed P_row and P_col (which are equivalent except in the form that they are represented in) and L_row and L_col (which are respective fitness functions where L_row(P_row) =  L_col(P_col)), they do not converge to the same result with the current implementation of CMAES. This wasn't very difficult to solve considering there was a transposed space (for example, row-wise representation) along with the space (column wise representation) itself.

This might be a good idea to add as a test case, to ensure that you get the same result either way.

I didn't see where exactly this was fixed in the code, though---could you point me towards where those changes are?  Maybe I overlooked it.

> I haven't updated callbacks_test.cpp yet. I'm not sure if I should take the initial step size as a parameter in Optimize(...).

I left a comment in the PR, but step size should be a constructor parameter to match all the other optimizers we have implemented.  I suspect the adaptation of callbacks_test.cpp will be basically trivial, like the other test changes you made, once that's changed. :+1:

Let me know if I can clarify any of my comments.  I haven't gone through the exact changes completely in detail yet, so I still need to check the mathematical details.  I figured let's get the general design right first. :)

include/ensmallen_bits/cmaes/cmaes.hpp
SuvarshaChennareddy: The default transformation policy is `EmptyTransformation`. However, I could preserve a constructor that takes `lowerBound` and `upperBound` as parameters of type `double`. This would then throw an exception if `TransformationPolicyType` is not `BoundaryBoxConstraint` like you said.

include/ensmallen_bits/cmaes/cmaes.hpp
SuvarshaChennareddy: Alright! I will add the initial step size as a parameter of the constructor instead. I also agree that adding a default step size method for each TransformationPolicyType would work.

include/ensmallen_bits/cmaes/cmaes.hpp
SuvarshaChennareddy: Alright!

include/ensmallen_bits/cmaes/transformation_policies/empty_transformation.hpp
SuvarshaChennareddy: No worries, sorry about that!

include/ensmallen_bits/cmaes/transformation_policies/empty_transformation.hpp
SuvarshaChennareddy: 👍

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
SuvarshaChennareddy: Yeah...I was a bit hesitant on adding it, we really don't need it at all in this case. Is commenting out fine?

include/ensmallen_bits/cmaes/cmaes_impl.hpp
SuvarshaChennareddy: We certainly could, but it isn't required. I request you to refer the comment above this one.

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
SuvarshaChennareddy: I think my comment on the relation between CMA-ES and the transformation policies (them being independent) explains why this was not done.

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
SuvarshaChennareddy: 👍

include/ensmallen_bits/cmaes/transformation_policies/empty_transformation.hpp
```diff
@@ -0,0 +1,45 @@
...
+{
+public:
+
+/**
+* Transforms coordinates to themselves (effectively no transformation).
+*
+* @param x Input coordinates.
+* @return Transformed coordinates (the coordinates themselves).
+*/
+MatType Transform(const MatType& x) { return x; }
```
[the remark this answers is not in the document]
SuvarshaChennareddy: I don't this is required though.

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
SuvarshaChennareddy: Alright! This would work well with preserving a constructor that takes `lowerBound` and `upperBound` as parameters of type `double.`

tests/cmaes_test.cpp
SuvarshaChennareddy: Yes, makes sense.

tests/cmaes_test.cpp
SuvarshaChennareddy: Alright!

include/ensmallen_bits/cmaes/cmaes_impl.hpp
SuvarshaChennareddy: In Nikolaus Hansen's [The CMA Evolution Strategy: A Tutorial](https://arxiv.org/pdf/1604.00772v1.pdf) `C^{-1/2}` is defined to be `B * D^{-1} * B^T` which isn't always equal to `covLower = B * D^{-1}`

include/ensmallen_bits/cmaes/cmaes_impl.hpp
SuvarshaChennareddy: This is the case when the parameters are represented as a row vector. In the case of a column vector, `pStep[idx(j)] = covLower * arma::randn<BaseMatType>(iterate.n_rows, iterate.n_cols)`. The same parameters represented as a row vector would use the trasposed form: `pStep[idx(j)] = arma::randn<BaseMatType>(iterate.n_rows, iterate.n_cols) * covLower.t()`. I've explained this inconsistency in point one of this PR.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
SuvarshaChennareddy: Ahh yes! Sorry and thank you for pointing this out! I will fix this very soon.

SuvarshaChennareddy: > That may require some further refactoring, but it seems to me to be conceptually simpler to simply use `Transform()` to map each new proposed iterate back into the required range right after a step is taken, only once.

This would make the update steps depend on the transformation policy, which isn't really my intention here.

include/ensmallen_bits/cmaes/cmaes.hpp
SuvarshaChennareddy: I'll definitely add a comment.

SuvarshaChennareddy:
Thank you @rcurtin for the comments, I'll definitely look into them. Please let me know if anything doesn't make sense or needs further clarification.

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
SuvarshaChennareddy: Actually, `lowerBound` and `UpperBound` do not need to be the same size as x. This is possible due to `Bi` and `Bj`.

include/ensmallen_bits/cmaes/transformation_policies/boundary_box_constraint.hpp
rcurtin: I would just remove it.  If it's left in but commented out, my experience is that nobody will ever come back and uncomment it; it'll just be dead code.

include/ensmallen_bits/cmaes/cmaes.hpp
rcurtin: Yeah, that would be nice. Ideally this should preserve reverse compatibility completely, so that if a user still used the old constructor, the behavior would be identical.  That might result in some amount of slightly ugly code, but you can mark the relevant functions as deprecated and leave a comment that they should be removed for the next major version of ensmallen. :+1:

SuvarshaChennareddy:
@rcurtin @zoq Let me know if there is anything else that that you all would like me to change or add.

conradsnicta:
@rcurtin is this ok to merge?

rcurtin:
I haven't had a chance to come back to it---it still needs a pretty complete review (sorry @SuvarshaChennareddy!  Life has been busy).

SuvarshaChennareddy:
No worries!

SuvarshaChennareddy:
Keep open

SuvarshaChennareddy:
Can we keep this open?

SuvarshaChennareddy:
@rcurtin @zoq

zoq:
Reopened.

zoq (approved):
Looks good to me.

rcurtin (changes requested):
Sorry that I have been so slow to respond on this.  It can't be fun on your side waiting for these comments that take so long.

The implementation itself seems fine to me (or at least I have run out of time to check it more deeply), but there are a couple of things that we do have to handle before merge, which should be easy:

 * The documentation in `doc/optimizers.md` needs to be updated.  Take a look at the `CMAES` section; it should be fairly easy to adapt.  You don't have to worry about reverse compatibility here---that's only so that old code compiles; we don't need to advertise deprecated support.
 * Add an update to `HISTORY.md` so that something about this PR is in the release notes. :)
 * It seems like the CI workers are not working correctly right now, but I am not able to build this branch on my system.  When I try to build the tests, I get several compilation errors.  Most of them look to be relatively simple to fix.

I think that's it.  I had a few other little comments throughout.

rcurtin:
Also this warning should be fixed:

```
In file included from /home/ryan/src/ensmallen/include/ensmallen_bits/cmaes/cmaes.hpp:209,
                 from /home/ryan/src/ensmallen/include/ensmallen.hpp:99,
                 from /home/ryan/src/ensmallen/tests/main.cpp:12:
/home/ryan/src/ensmallen/include/ensmallen_bits/cmaes/cmaes_impl.hpp:37:32: warning: ISO C++ forbids typename key in template template parameter; use ‘-std=c++17’ or ‘-std=gnu++17’ [-Wc++17-extensions]
   37 | template<template<typename...> typename T, typename... A, typename... B>
```

You can just change it to `template<typename...> class T`, it means the same thing and is compatible with older C++ standards.

rcurtin:
Maybe removing the label manually will make @mlpack-bot happy :)

SuvarshaChennareddy:
@rcurtin Thank you for the comments. I've made the necessary changes. If there's anything more, let me know.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
SuvarshaChennareddy: I understand and agree. I haven't had much time to think about this either but I'll try to figure something out soon.

rcurtin:
I pushed a couple of little style changes, and a few fixes for compilation that I found when I tried to run locally.  Everything looks good to me, but I noticed that all the CMAES tests are failing:

```
-------------------------------------------------------------------------------
CMAESDeprecatedConstructorLogisticRegressionTest
-------------------------------------------------------------------------------
/home/ryan/src/ensmallen/tests/cmaes_test.cpp:26
...............................................................................

/home/ryan/src/ensmallen/tests/test_function_tools.hpp:215: FAILED:
  REQUIRE( acc == Approx(100.0).epsilon(trainAccuracyTolerance) )
with expansion:
  61.3 == Approx( 100.0 )

-------------------------------------------------------------------------------
CMAESLogisticRegressionTest
-------------------------------------------------------------------------------
/home/ryan/src/ensmallen/tests/cmaes_test.cpp:36
...............................................................................

/home/ryan/src/ensmallen/tests/test_function_tools.hpp:215: FAILED:
  REQUIRE( acc == Approx(100.0).epsilon(trainAccuracyTolerance) )
with expansion:
  88.0 == Approx( 100.0 )

-------------------------------------------------------------------------------
ApproxCMAESLogisticRegressionTest
-------------------------------------------------------------------------------
/home/ryan/src/ensmallen/tests/cmaes_test.cpp:47
...............................................................................

/home/ryan/src/ensmallen/tests/test_function_tools.hpp:215: FAILED:
  REQUIRE( acc == Approx(100.0).epsilon(trainAccuracyTolerance) )
with expansion:
  87.4 == Approx( 100.0 )

-------------------------------------------------------------------------------
CMAESLogisticRegressionFMatTest
-------------------------------------------------------------------------------
/home/ryan/src/ensmallen/tests/cmaes_test.cpp:58
...............................................................................

/home/ryan/src/ensmallen/tests/test_function_tools.hpp:215: FAILED:
  REQUIRE( acc == Approx(100.0).epsilon(trainAccuracyTolerance) )
with expansion:
  76.9 == Approx( 100.0 )

-------------------------------------------------------------------------------
ApproxCMAESLogisticRegressionFMatTest
-------------------------------------------------------------------------------
/home/ryan/src/ensmallen/tests/cmaes_test.cpp:69
...............................................................................

/home/ryan/src/ensmallen/tests/test_function_tools.hpp:215: FAILED:
  REQUIRE( acc == Approx(100.0).epsilon(trainAccuracyTolerance) )
with expansion:
  81.3 == Approx( 100.0 )
```

rcurtin:
From my side everything else looks good, so as soon as we have a green build I'm excited for merge (and then a release with the fixes). :+1:

SuvarshaChennareddy:
Thank you, @rcurtin, for looking into this. I made a few quick modifications (including a bug fix) and ran the tests, all of which appear to pass successfully (when executed locally).

SuvarshaChennareddy:
@zoq 
Upon running the tests for multiple iterations (with a different random seed every time), I've noticed that for some iterations, one or two tests fail at most. I don't know the reason yet, but I'm guessing it's due to the hyperparameters set (such as the boundary box constraint domain and the population size). This assumption is based on the fact that the last test, which uses the Empty Transformation policy, never failed.

zoq:
I'll check the implementation again, and maybe we can finetune the hyperparameters to make the tests more stable. In the past, we adapted some thresholds, but we should make sure, the failing test is not pointing torwards an issue with the implementation.

SuvarshaChennareddy:
Alright! Thanks for looking into this, @zoq.

zoq:

zoq:
To trigger the CPU test suite.

SuvarshaChennareddy:
That's weird, the accuracy received from the failed test was shown to be 50 percent. It's almost as if CMAES couldn't find a path to a local/global minimum during optimization.
This wasn't the case with other test cases tho :/.

zoq:
> Details

It's not the first time I have seen this, perhaps it just needs some more steps to find a solution. Let me rerun the test.

zoq:

SuvarshaChennareddy:
Hmm...This time it wasn't CMAES that failed.

zoq:
Right, and we test it 8 times in a row with different random seeds, do you mind to increase the number of iterations for the `CMAESLogisticRegressionFMatTest` test?

zoq:
@SuvarshaChennareddy see comment above, let me know, I can quickly push a patch as well.

SuvarshaChennareddy:
@zoq, I'll update the hyper-parameters once more. I don't think just changing the number of iterations will affect the results.

zoq:
I still like to keep the test time low, if we set the maximum iterations to zero, we could potentially run for a long time. Let's increase the population size. The main goal is to make sure the tests actually test that the implementation is correct, if the test fails with the hyp parameters a few times, that's fine but I like to reduce the cases it fails.

SuvarshaChennareddy:
Alright I'll increase the population size to 256 then.

zoq:

SuvarshaChennareddy:
Alright, I have an idea on what might be going on. I believe the step size is diverging to a relatively large value ( > 10^14). I'll add a termination condition for this. I'll also add a patience (similar to the one provided by the [EarlyStopAtMinLossType](https://github.com/mlpack/ensmallen/blob/master/include/ensmallen_bits/callbacks/early_stop_at_min_loss.hpp) Callback) to allow the optimizer to explore the search space even if the overallObjective doesn't improve immediately after a generation.

zoq:
Sounds good, thanks.

SuvarshaChennareddy:

SuvarshaChennareddy:
@zoq the tests passed

rcurtin (approved):
Nothing else from my side, sorry for the slight holdup! Great to see this finally get across the finish line :+1:

zoq:
Thanks for putting this together.

rcurtin:
Do we want to make a release now?  (also, awesome work @SuvarshaChennareddy, great to have this in :))

zoq:
> Do we want to make a release now? (also, awesome work @SuvarshaChennareddy, great to have this in :))

I'll open a PR to fix some compiler warnings, I think afterwards we can do another release.

PR #352, opened 2023-01-03
Bandicoot adaption

zoq:
Initial Bandicoot adaption.

### DIFF
--- a/include/ensmallen_bits/eve/eve.hpp
+++ b/include/ensmallen_bits/eve/eve.hpp
@@ -106,8 +106,8 @@ class Eve
            typename MatType,
            typename GradType,
            typename... CallbackTypes>
-  typename std::enable_if<IsArmaType<GradType>::value,
-      typename MatType::elem_type>::type
+  typename std::enable_if<IsArmaType<GradType>::value ||
+      coot::is_coot_type<GradType>::value, typename MatType::elem_type>::type
   Optimize(SeparableFunctionType& function,
            MatType& iterate,
            CallbackTypes&&... callbacks);
--- a/include/ensmallen_bits/utility/randn.hpp
+++ b/include/ensmallen_bits/utility/randn.hpp
@@ -1,0 +1,72 @@
+/**
+ * @file randn.hpp
+ * @author Marcus Edel
+ *
+ * A simple `randn` wrapper that based on the data type forwards to
+ * `coot::randn` or `arma::randi`.
+ *
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_UTILITY_RANDN_HPP
+#define ENSMALLEN_UTILITY_RANDN_HPP
+
+namespace ens {
+
+template<typename ElemType>
+typename std::enable_if<coot::is_coot_type<ElemType>::value, ElemType>::type
+randn(const size_t rows, const size_t cols)
+{
+  /* return coot::randu<ElemType>(rows, cols); */
+  return coot::randn<ElemType>(rows, cols);
+  /* ElemType foo; */
+  /* return foo; */
+}
+
+template<typename ElemType>
+typename std::enable_if<!coot::is_coot_type<ElemType>::value, ElemType>::type
+randn(const size_t rows, const size_t cols)
+{
+  return arma::randn<ElemType>(rows, cols);
+}
+
+/* template<typename ElemType> */
+/* typename std::enable_if<coot::is_coot_type<ElemType>::value, ElemType>::type */
+/* randi(const size_t size, const arma::distr_param& param) */
+/* { */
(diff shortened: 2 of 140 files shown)


.appveyor.yml
```diff
@@ -1,5 +1,5 @@
 environment:
-  ARMADILLO_DOWNLOAD: "https://sourceforge.net/projects/arma/files/armadillo-9.800.6.tar.xz"
+  ARMADILLO_DOWNLOAD: "http://ftp.fau.de/macports/distfiles/armadillo/armadillo-8.400.0.tar.xz"
```
conradsnicta, line 2: @zoq is that version downgrade deliberate?
recent releases of ensmallen require armadillo 9.800+

include/ensmallen_bits/eve/eve.hpp
```diff
@@ -106,8 +106,8 @@ class Eve
            typename MatType,
            typename GradType,
            typename... CallbackTypes>
-  typename std::enable_if<IsArmaType<GradType>::value,
-      typename MatType::elem_type>::type
+  typename std::enable_if<IsArmaType<GradType>::value ||
+      coot::is_coot_type<GradType>::value, typename MatType::elem_type>::type
```
rcurtin, line 110: One alternate idea here would be to make a combined `IsArmaOrCootType` traits class and use that instead.

include/ensmallen_bits/utility/randn.hpp
```diff
@@ -0,0 +1,72 @@
...
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_UTILITY_RANDN_HPP
+#define ENSMALLEN_UTILITY_RANDN_HPP
+
+namespace ens {
+
+template<typename ElemType>
+typename std::enable_if<coot::is_coot_type<ElemType>::value, ElemType>::type
+randn(const size_t rows, const size_t cols)
```
rcurtin, line 20: I think that the RNGs I just merged to bandicoot should match the Armadillo API, so I *think* you should be able to remove these functions and just call `randn<MatType>(...)` directly, etc.

rcurtin:
This is super awesome!  I'm still working through the related Bandicoot changes and opening PRs, but little by little I'll get everything opened in MRs there so that this will work.  I only took a quick glance for now, but once I can properly test with Bandicoot I'll do a more detailed review. :rocket:

shrit:
@zoq I tried to review it, but there are massive amount of things that needs to be completed, this looks to me more of a draft at this stage, let me know if you need any help

zoq:
I think all of the SGD based optimizers are working (let me know if you miss one), which I think is what you will need, to move forward with the mlpack part? Some of the optimizers will have to wait for downstream implementations of e.g. `qr`.

So not sure what you mean by "massive amount of things that needs to be completed"?

rcurtin (commented):
Awesome to finally see this come together.  I reviewed most of the optimizer implementations, but still have to look at the CMake configuration and the tests.

For the tests, I think that the only function really worth testing with Bandicoot is the `LogisticRegressionTestFunction`; almost all of the other ones compute the objective via individual element access, or operate on matrices so small that there is no hope of a GPU ever being fast enough (just the communication costs alone will be way too high).

Ideally I would hope that for at least this initial PR, we can get to a state where:

 * SGD and L-BFGS are reasonably fast enough that we can run the logistic regression tests.  (I may need to do a little Bandicoot tuning and implementation behind-the-scenes.)
 
  * Other optimizers will at least compile with Bandicoot matrices, but we don't actually need to run the tests if they are slow.  It should be straightforward enough to run with timing reports to see the optimizers where Bandicoot is obscenely slow (probably due to element access).  In those cases, we can just add a test that creates the optimizer and either does 1 iteration only, or otherwise terminates immediately---then, the test is more of a check that that optimizer will at least compile with Bandicoot matrices.

I left some comments throughout, but often times the comment could be applied in many places but I only left it once.  Let me know what you think, happy to help out with implementation or if there are other missing parts of Bandicoot that would make life easier. :smile:

rcurtin:
I took a look into what to do about the `FindBandicoot.cmake` file.  I think the result I like this most is this one: https://stackoverflow.com/questions/6580856/how-to-distribute-findxxx-cmake

Basically the idea would be to put `FindBandicoot.cmake` (and related files) in the Bandicoot repository, try to install them to `${CMAKE_ROOT}/Modules/` when the library is installed, and then also it would be a good idea to add a note to the README indicating that `FindBandicoot.cmake` can be used directly by copying it to another project.  (And I guess we should do the same with `Findensmallen.cmake` and `Findmlpack.cmake` eventually.)

include/ensmallen_bits/utility/proxies.hpp
```diff
@@ -0,0 +1,148 @@
...
+#define ENSMALLEN_UTILITY_PROXIES_HPP
+
+namespace ens {
+
+template<typename ElemType>
+typename std::enable_if<IsCootType<ElemType>::value, ElemType>::type
+randu(const size_t rows, const size_t cols)
+{
+  #ifdef USE_COOT
+  return coot::randu<ElemType>(rows, cols);
```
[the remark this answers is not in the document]
shrit: Agreed, I will remove them and replace it with using

PR #363, opened 2023-03-19
Added Generation MOEA-d (Extended version of MOEA-d)Algorithm

Sainithin-bit:
Right now ensmallen have MOEA/D algorithm however,  MOEA/D algorithm has been shown to work well on many multi-objective optimization problems, particularly those with convex Pareto fronts .But It has some drawbacks like premature convergence, where the diversity of the population is reduced as the algorithm progresses, and it can become stuck in local optima.

The Generation MOEA/D algorithm was developed as an extension to the MOEA/D algorithm to address this issue, by using a different approach to maintain diversity in the population.

The code I have added is  backbone algorithm , modifications are required to align it with coding style of ensmallen. I am just creating this pull request for the reviewers to check and say go ahead for making further changes. 

TODO:
Remove existing dependencies and replace with ensmallen dependencies 
Align it with Ensmallen Coding style 
Add Test Cases 
Add Documentation


zoq:
Thanks for opening the PR, please go ahead with the changes, let us know if there is anything that needs clarification.

Sainithin-bit:
Thanks for appreciating the effort and quick reply. I will let you know if there's need of any clarification.

PR #366, opened 2023-06-06
add short list of example optimisers

conradsnicta:
Add example optimisers to increase relevant keywords.

This increases the likelihood of ensmallen being found via web search (eg. users typing specific keywords into google).

The corresponding PR for ensmallen.org is at https://github.com/mlpack/ensmallen.org/pull/15


zoq (approved):
Good idea.

coatless (approved):

coatless:
👍 

In we go.

PR #367, opened 2023-06-09
Implementation of Active CMAES

SuvarshaChennareddy:
This pull request is an initial implementation of Active CMAES, which is part of my Google Summer of Code (GSoC) project. Tests have not been added yet, but they will be included soon.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,7 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
+* Avoid deprecation warnings in Armadillo 11.2+
+   ([#347](https://github.com/mlpack/ensmallen/pull/347)).
 
 ### ensmallen 2.19.0: "Eight Ball Deluxe"
 ###### 2022-04-06
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,9 +2,6 @@
 ###### ????-??-??
  * Implementation of Active CMAES
    ([#367](https://github.com/mlpack/ensmallen/pull/367)).
-   
- * Fix the current CMAES inconsistencies
-   ([#351](https://github.com/mlpack/ensmallen/pull/351)).
 
  * LBFGS: avoid generation of NaNs, and add checks for finite values
    ([#368](https://github.com/mlpack/ensmallen/pull/368)).
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,7 +1,13 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
+ * Implementation of Active CMAES
+   ([#367](https://github.com/mlpack/ensmallen/pull/367)).
+   
  * Fix the current CMAES inconsistencies
    ([#351](https://github.com/mlpack/ensmallen/pull/351)).
+
+ * LBFGS: avoid generation of NaNs, and add checks for finite values
+   ([#368](https://github.com/mlpack/ensmallen/pull/368)).
 
  * Fix CNE test tolerances
    ([#360](https://github.com/mlpack/ensmallen/pull/360)).
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,466 +1,1036 @@
-### ensmallen ?.??.?: "???"
+### mlpack ?.?.?
 ###### ????-??-??
-* Avoid deprecation warnings in Armadillo 11.2+
-   ([#347](https://github.com/mlpack/ensmallen/pull/347)).
+  * Fix mapping of categorical data for Julia bindings (#3305).
 
-### ensmallen 2.19.0: "Eight Ball Deluxe"
-###### 2022-04-06
-* Added DemonSGD and DemonAdam optimizers
-   ([#211](https://github.com/mlpack/ensmallen/pull/211)).
+### mlpack 4.0.0
+###### 2022-10-23
+  * Bump C++ standard requirement to C++14 (#3233).
 
- * Fix bug with Adam-like optimizers not resetting when `resetPolicy` is `true`.
-   ([#340](https://github.com/mlpack/ensmallen/pull/340)).
+  * Fix `Perceptron` to work with cross-validation framework (#3190).
 
- * Add Yogi optimizer
-   ([#232](https://github.com/mlpack/ensmallen/pull/232)).
+  * Migrate from boost tests to Catch2 framework (#2523), (#2584).
 
- * Add AdaBelief optimizer
-   ([#233](https://github.com/mlpack/ensmallen/pull/233)).
+  * Bump minimum armadillo version from 8.400 to 9.800 (#3043), (#3048).
 
- * Add AdaSqrt optimizer
-   ([#234](https://github.com/mlpack/ensmallen/pull/234)).
-   
- * Bump check for minimum supported version of Armadillo
-   ([#342](https://github.com/mlpack/ensmallen/pull/342)).
+  * Adding a copy constructor in the Convolution layer (#3067).
 
-### ensmallen 2.18.2: "Fairmount Bagel"
-###### 2022-02-13
- * Update Catch2 to 2.13.8
-  ([#336](https://github.com/mlpack/ensmallen/pull/336)).
+  * Replace `boost::spirit` parser by a local efficient implementation (#2942).
 
- * Fix epoch timing output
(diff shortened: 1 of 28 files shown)


SuvarshaChennareddy:
It's just the CMAES tests that are failing.

include/ensmallen_bits/cmaes/active_cmaes_impl.hpp
```diff
@@ -0,0 +1,378 @@
...
+#ifndef ENSMALLEN_ACTIVE_CMAES_CMAES_IMPL_HPP
+#define ENSMALLEN_ACTIVE_CMAES_CMAES_IMPL_HPP
+
+// In case it hasn't been included yet.
+#include "active_cmaes.hpp"
+
+#include <ensmallen_bits/function.hpp>
+
+#ifndef NOT_EMPTY_TRANSFORMATION
+#define NOT_EMPTY_TRANSFORMATION
```
conradsnicta, line 24: not sure what this define is for

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -20,31 +20,82 @@
 
 #include <ensmallen_bits/function.hpp>
 
+#ifndef NOT_EMPTY_TRANSFORMATION
+#define NOT_EMPTY_TRANSFORMATION
```
conradsnicta, line 24: not sure what this define is for

HISTORY.md
```diff
@@ -1,5 +1,8 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
+ * Fix the current CMAES inconsistencies
+   ([#351](https://github.com/mlpack/ensmallen/pull/351)).
```
conradsnicta, line 4: this doesn't look right.  the present PR is 367, and has the description of "Implementation of Active CMAES".  why mention 351 here?

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -423,9 +423,9 @@ inline bool NSGA2::Dominates(
 //! Assign crowding distance to the population.
 template <typename MatType>
 inline void NSGA2::CrowdingDistanceAssignment(
-	const std::vector<size_t>& front,
-	std::vector<arma::Col<typename MatType::elem_type>>& calculatedObjectives,
-	std::vector<typename MatType::elem_type>& crowdingDistance)
+    const std::vector<size_t>& front,
+    std::vector<arma::Col<typename MatType::elem_type>>& calculatedObjectives,
+    std::vector<typename MatType::elem_type>& crowdingDistance)
```
conradsnicta, line 428: these are just changes in whitespace. what's the reason for this?  if the changes are not related to Active CMAES, please omit them

conradsnicta (commented):
minor comments

include/ensmallen_bits/cmaes/active_cmaes_impl.hpp
SuvarshaChennareddy: Thank you for pointing this out! Now that I look at it, I don't think this define is required. The `ACTIVE_CMAES_CMAES_IMPL_HPP` define should be enough. I'll make changes.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
SuvarshaChennareddy: I'll be sure to update this too.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
rcurtin: It fixes the style to only indent twice on a wrapped line---personally I think it's just fine.

HISTORY.md
SuvarshaChennareddy: Sorry about that! The branch of this PR (367) is actually a branch of PR 351's branch 
😅. I was trying to find a cleaner way to open this PR, but couldn't really find one (that being said, is there a way?). 
I know I'll have to deal with the merge conflicts if PR 351 gets merged.

SuvarshaChennareddy: I'll update this soon too.

conradsnicta: suggest to keep PR 351 and this PR as separate (ie. not inter-dependent).  if this PR is merged first, PR 351 may then need to be updated (and vice-versa).

include/ensmallen_bits/nsga2/nsga2_impl.hpp
conradsnicta: The whitespace changes are not a problem by themselves.  The issue is that this PR seems to be touching stuff that's not directly related to it.  I'm worried that there may be other irrelevant files that are also changed, possibly introducing regressions.

HISTORY.md
SuvarshaChennareddy: I understand, I believe (between the two if any) PR 351 will get merged first. So I would have to update this PR soon (resolving merge and conflicts and such).

include/ensmallen_bits/cmaes/active_cmaes_impl.hpp
SuvarshaChennareddy: I'm looking at this once more and I realize that if the define is removed and the user uses both Vanilla and Active CMAES, there will be double inclusion (which is why I added the include guard in the first place). The `ACTIVE_CMAES_CMAES_IMPL_HPP` define is indeed not enough.

rcurtin: It sounds like in that case that either `NotEmptyTransfornation` should be moved to its own file and included by both versions, or if the functionality differs here then perhaps it just needs a new name. I think I have understood the problem, let me know if not.

SuvarshaChennareddy: Alright, I'll move `NotEmptyTransfornation` to its own file.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
SuvarshaChennareddy: These changes were pushed with the last PR [351](https://github.com/mlpack/ensmallen/pull/367).

zoq: Thanks

zoq:

zoq (approved):
No more comments from my side, thanks for the contribution.

SuvarshaChennareddy:
Thank you, @zoq!
There is one thing I would like to address. Wouldn't removing the deprecation go against [this](https://github.com/mlpack/ensmallen/pull/351#discussion_r1064832640). I also think we should make changes (fixing style and changing element types) the to Vanilla CMAES for consistency.

SuvarshaChennareddy:
> If we use `BoundaryBoxConstraint` as the default transformation policy wouldn't that be the same as using the other constructor?

Yes, but the default is `EmptyTransformation`.

zoq:
Do you see any problem with using `BoundaryBoxConstraint` as the default for the constructor?

zoq:
We could even provide another constructor for thee empty transformation.

zoq:
Looking at it again, if we keep the constructor, it's backward compatible, since I can still use the "old" constructor.

SuvarshaChennareddy:
> Looking at it again, if we keep the constructor, it's backward compatible, since I can still use the "old" constructor.

Yes, that's why it was marked as deprecated instead of being removed altogether. We could make `BoundaryBoxConstraint` the default constructor; however, CMA-ES itself isn't bounded. So, I think it would make more sense to keep `EmptyTransformation` as the default. Since it would be the default, keeping it deprecated would also make sense, wouldn't it? These are just my thoughts.

zoq:
I think both interface are fine, the old interface is closer to what we provide for the other optimizers, so I like to keep it, the new one offers the ability to support more transformations so I like to keep it as well.

SuvarshaChennareddy:
Alright. We should probably modify Vanilla CMA-ES as well then (the element types should be changed as well for consistency).

zoq:
Makes sense, I'll resolve the failing test and merge this PR.

zoq:
Thanks for the great contribution.

Issue #365, opened 2023-05-11
During AugLagrangian's optimization, L_BFGS sets coordinates to nan


olgavrou:
#### Issue description

During `AugLagrangian`'s optimization, `L_BFGS` sets coordinates to `nan`

I believe the issue is [here](https://github.com/mlpack/ensmallen/blob/master/include/ensmallen_bits/lbfgs/lbfgs_impl.hpp#L89)

```
scalingFactor = dot(sMat, yMat) / dot(yMat, yMat);
```

if the two dot products are very small this leads to the scaling factor becoming a `nan`

In my experiment, if I replace that line with the code pasted below:

```
scalingFactor = dot(sMat, yMat) / std::max(1e-12, dot(yMat, yMat));
```

(EDITED)

or

```
double max_y = arma::abs(yMat).max();

if (max_y == 0.0)
{
  scalingFactor = 0.0;
}
else
{
  auto z = yMat / max_y;
  scalingFactor = (dot(sMat, z) / max_y) / dot(z, z);
}
```

which results in the optimizer working as expected.

Please let me know if you think this calculation is ok for the library and if you would be open to either patching it or accepting a patch.

#### Your environment

 * version of ensmallen: 2.19.1
 * Ubuntu 20.04:
 * compiler: gcc 9.4.0
 * version of Armadillo: 12.1.90
 * any other environment information you think is relevant:

#### Steps to reproduce

I have been iterating over the implementation of my constraint optimization problem you can see the latest version [here](https://github.com/VowpalWabbit/vowpal_wabbit/pull/4581/files) and the behaviour is seen when [this](https://github.com/VowpalWabbit/vowpal_wabbit/pull/4581/files#diff-77e8284a33c61d33a5de316722e95a09a60773df3574240f7a21dd6e90a34cf8L593) test (`CheckUpdateRule500WIterations`) is [run](https://github.com/VowpalWabbit/vowpal_wabbit/actions/runs/4942865717/jobs/8836792141) 

but unfortunately I don't have a smaller repro in hand

#### Expected behavior

Coordinates not getting set to `nan`

#### Actual behavior

Coordinates get set to `nan`

olgavrou:
OK will do :) I think the first solution is probably working better, I got another `nan` with the second one at some point

conradsnicta:
> `scalingFactor = dot(sMat, yMat) / std::max(1e-12, dot(yMat, yMat));`

[...]

PR #368, opened 2023-06-21
LBFGS: avoid generation of NaNs, and add checks for finite values

conradsnicta:
related to #365

### DIFF
--- a/include/ensmallen_bits/lbfgs/lbfgs_impl.hpp
+++ b/include/ensmallen_bits/lbfgs/lbfgs_impl.hpp
@@ -86,11 +86,17 @@ double L_BFGS::ChooseScalingFactor(const size_t iterationNum,
     // Get s and y matrices once instead of multiple times.
     const arma::Mat<CubeElemType>& sMat = s.slice(previousPos);
     const arma::Mat<CubeElemType>& yMat = y.slice(previousPos);
-    scalingFactor = dot(sMat, yMat) / dot(yMat, yMat);
+    
+    const CubeElemType tmp   = arma::dot(yMat, yMat);
+    const CubeElemType denom = (tmp != CubeElemType(0)) ? tmp : CubeElemType(1);
+    
+    scalingFactor = arma::dot(sMat, yMat) / denom;
   }
   else
   {
-    scalingFactor = 1.0 / sqrt(dot(gradient, gradient));
+    const CubeElemType tmp = arma::norm(gradient, "fro");
+    
+    scalingFactor = (tmp != CubeElemType(0)) ? (1.0 / tmp) : 1.0;
   }
 
   return scalingFactor;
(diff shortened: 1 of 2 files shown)


include/ensmallen_bits/lbfgs/lbfgs_impl.hpp
```diff
@@ -212,6 +224,13 @@ bool L_BFGS::LineSearch(FunctionType& function,
   double stepSize = 1.0;
   finalStepSize = 0.0; // Set only when we take the step.
 
+  if (searchDirection.is_finite() == false)
+  {
+    Warn << "L-BFGS line search direction vector is not finite "
+        << "(terminating)!" << std::endl;
+    return false;
+  }
```
rcurtin, line 232: Personally I think that the better check would be on `initialSearchDirectionDotGradient`; if `searchDirection` isn't finite, then that won't be either.  And if `gradient` isn't finite, the same thing will happen, so checking the scalar `initialSearchDirectionDotGradient` seems like it is basically the same level of error detection for much cheaper.

include/ensmallen_bits/lbfgs/lbfgs_impl.hpp
```diff
@@ -86,11 +86,17 @@ double L_BFGS::ChooseScalingFactor(const size_t iterationNum,
     // Get s and y matrices once instead of multiple times.
     const arma::Mat<CubeElemType>& sMat = s.slice(previousPos);
     const arma::Mat<CubeElemType>& yMat = y.slice(previousPos);
-    scalingFactor = dot(sMat, yMat) / dot(yMat, yMat);
+    
+    const CubeElemType tmp   = arma::dot(yMat, yMat);
+    const CubeElemType denom = (tmp != CubeElemType(0)) ? tmp : CubeElemType(1);
```
rcurtin, line 91: Why divide by 1 here and not some small number that won't cause overflow?  I would worry that this strategy would not capture the intuition of setting the scaling factor to be large when `dot(yMat, yMat)` is small.

rcurtin (commented):
Thanks for handling this---just some small comments. :+1:

include/ensmallen_bits/lbfgs/lbfgs_impl.hpp
conradsnicta: One problem is that in relation to the numerator `dot(sMat, yMat)`, we don't know which "small" number to use as the denominator, or whether a "small" number would be appropriate at all.

If the denominator is 0, this would imply that we're in all sorts of trouble with the provided data.

Given the lack of sufficient information, the best "non-informative" choice to avoid the generation of NaNs seems to be the mathematical equivalent of a no-op in the division, ie. denominator = 1.

This has parallels with using 1 on the diagonals of covariance matrices when encountering problematic data.  Occam's razor may also be applicable here.

include/ensmallen_bits/lbfgs/lbfgs_impl.hpp
conradsnicta: good point; implemented in https://github.com/mlpack/ensmallen/pull/368/commits/49ad90c6193dfdd86c9be7b155473ee738d07685

include/ensmallen_bits/lbfgs/lbfgs_impl.hpp
rcurtin: Yes, I see what you mean now.  I think there's a case to be made to just take `scalingFactor` as "some large value that's not quite inf", but that case is easier to poke holes in than yours (e.g. if `dot(sMat, yMat)` is actually smaller than `dot(yMat, yMat)`, but both are represented in floating point as 0).  So, I agree with your reasoning.

rcurtin (approved):
Thanks, no more comments from my side. :+1:

zoq (approved):
Looks good to me.

Issue #378, opened 2023-09-15
SCD isn't stochastic and perhaps should be called CD?


rcurtin:
I was looking for a coordinate descent implementation and found that we do have SCD, but the actual implementation itself isn't "stochastic" in the sense of "stochastic gradient descent".  Instead, in the setting of machine learning, this would operate on all of the points at once.

My proposal is that it would be a good idea to rename this to `CD` instead of `SCD`, and maintain `SCD` (and the current typedefs) as deprecated.  Perhaps at another time in the future we can add an actually stochastic version that operates on only a few points in the dataset at a time.

Thoughts?  Did I misunderstand why the word "stochastic" was used here?

CC: @zoq, @shikharbhardwaj

conradsnicta:
No issues from my side.  We should probably release ensmallen 2.20 after this is resolved.  The last "proper" release was way back in April 2022, and enough notable changes have accumulated since then.

rcurtin:
Solved by #379.

PR #379, opened 2023-09-26
Rename `SCD` to `CD`

rcurtin:
This is a follow-up to #378.  I renamed `SCD`, now calling it `CD`.  `SCD` was retained as a deprecated name, and can be removed in ensmallen 3.  I also noticed in the documentation that the `GreedySCD` typedef and other typedefs didn't exist---so I added them.


zoq (approved):
Looks good to me.

coatless (approved):
LGTM 👍

PR #382, opened 2023-11-06
Fixed bool form signature of several callback methods

DiscreteLogarithm:

Issue #377, opened 2023-09-02
Custom early stop callback


Kaltxi:
#### Issue description

I am trying to make a custom callback to be able to stop optimization asynchronously. Documentation says that the callback should return true to end the process, so I made the following callback:

```C++
class OptimizationEarlyStop {
public:
  template <typename OptimizerType, typename FunctionType, typename MatType>
  auto Evaluate(OptimizerType& /* optimizer */,
                FunctionType& /* function */,
                const MatType& /* coordinates */,
                const double /* objective */) -> bool {
    return shouldStop_;
  }

  void stop() {
    const std::lock_guard lock(stopMutex_);
    shouldStop_ = true;
  }

private:
  bool shouldStop_{false};
  std::mutex stopMutex_;
};
```

That said the callback doesn't seem to run. If I make the `Evaluate` return `void` the callback runs, but I can't return `bool` in this case.

#### Your environment

 * version of ensmallen: 2.19.0
 * operating system: Manjaro, 6.1.44-1-MANJARO  kernel
 * compiler: clang 15.0.7
 * version of Armadillo: 12.4.1

#### Steps to reproduce

Defining the `Evaluate()` callback with `bool` return type makes callback unable to be run.

#### Expected behavior

I expect to be able to define `bool Evaluate()` and be able to stop optimization with it.

rcurtin:
Hey there @Kaltxi, can you also provide the code that you are using to run the optimization?  If it's a complete example that we can compile and run, it's way easier to debug.  Thanks!

Kaltxi:
@rcurtin Hello! Here is a minimal example that minimizes a trivial function with a slow down in its `Evaluate()`. Optimization is launched via `std::async` and while it is running the callback is called.
[ensmallen-early-return.zip](https://github.com/mlpack/ensmallen/files/12506154/ensmallen-early-return.zip)

zoq:
It looks like you never pass the callback to the Optimize function:

`const double minimum = optimizer.Optimize(f, coefs, ens::Report{});`

should be something like:

```c++
OptimizationEarlyStop early_stop;
const double minimum = optimizer.Optimize(f, coefs, ens::Report{}, early_stop);
```

Kaltxi:
@zoq Yeah, thats a typo, fixing it yields the same behavior. In any case the `bool` version of `OptimizationEarlyStop::Evaluate` is never called, when it's supposed to be called on every function evaluation.

zoq:
let's keep this open

[...]

PR #383, opened 2023-11-10
Fix callback return values and clarify docs

rcurtin:
This is a sibling PR to #382 and part of fixing #377.

I went through and audited the code to find places where the return values of callback functions were being ignored.  I modified optimizers as necessary to handle those return values, attempting to terminate optimization before any more steps were taken.

I also modified the documentation to clarify that `bool`s should be returned.  However, for `BeginOptimization()` and `EndOptimization()` returning a `bool` doesn't make sense---so I made those return `void`.

In a couple places there were various other cleanups that I did, but nothing that affected functionality in any real way unrelated to callbacks.

Once this is done/merged, I also plan to update the technical report's section on callbacks and re-upload to arXiv.

### DIFF
--- a/doc/callbacks.md
+++ b/doc/callbacks.md
@@ -424,18 +424,18 @@ std::cout << "The optimized model found by AdaDelta has the "
 Callbacks are called at several states during the optimization process:
 
 * At the beginning and end of the optimization process.
-* After any call to `Evaluate()` and `EvaluateConstraint`.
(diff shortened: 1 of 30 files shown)


zoq (approved):

rcurtin:
I submitted an update to the arXiv technical report for ensmallen that clarifies the new callback structure.

PR #384, opened 2023-11-15
Fix some callbacks that ignored return values

rcurtin:
I was debugging #377 further when I noticed that several of the callback functions (`Evaluate()`, `Gradient()`, `EvaluateConstraint()`, and others) actually don't use the return value of the callback.

Any overload of those callbacks would call:

```
    return (const_cast<CallbackType&>(callback).Evaluate(
        optimizer, function, coordinates, objective), false);
```

which always returns `false`.  The fix for this was to adjust the traits for each callback type to provide `::hasBool` and `::hasVoid` members, and then use the return value for the `::hasBool` version, just like `StepTaken()` does (and a couple other callbacks).

I also modified all our existing callbacks to return `bool` instead of `void`.

Once this PR is merged, the example in #377 works, and we should do a release.

### DIFF
--- a/include/ensmallen_bits/callbacks/traits.hpp
+++ b/include/ensmallen_bits/callbacks/traits.hpp
@@ -239,23 +239,11 @@ template<typename CallbackType,
          typename MatType>
 struct HasBeginOptimizationSignature
 {
-  const static bool hasBool =
+  const static bool value =
       HasBeginOptimization<CallbackType, TypedForms<OptimizerType,
-          FunctionType, MatType>::template BeginOptimizationBoolForm>::value &&
-      !HasBeginOptimization<CallbackType, TypedForms<OptimizerType,
+          FunctionType, MatType>::template BeginOptimizationBoolForm>::value ||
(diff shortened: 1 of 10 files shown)


include/ensmallen_bits/callbacks/traits.hpp
```diff
@@ -239,23 +239,11 @@ template<typename CallbackType,
          typename MatType>
 struct HasBeginOptimizationSignature
 {
-  const static bool hasBool =
+  const static bool value =
```
conradsnicta, line 242: Would it make sense to use `static constexpr bool` here, instead of plain `const` ?

Apart from the semantic info (clarity of intent), `constexpr` explicitly tells the compiler that this can be evaluated at compile time, instead of the compiler working this out for itself during optimisation.  This extra info should reduce compilation times somewhat.

rcurtin: Ah, good point---fixed in e21a3c57. :+1:

conradsnicta (approved):

PR #385, opened 2023-11-22
Add a newline after printing coordinates to `Report` callback

rcurtin:
Pretty simple change... before:

```
Optimization Report
--------------------------------------------------------------------------------

Initial coordinates: 
 0.0000  0.0000  0.0000 ...
 0.0000 
Final coordinates: 
-2.3592 -0.0324 -0.0261 ...
-0.1400 iter          loss          loss change   |gradient|    step size     total time    
0             0.883         0.000         4.532         0.010         0.007         
10            0.762         0.121         0.000         0.010         0.078         
20            1.195         -0.433        867.404       0.010         0.136         
30            0.627         0.567         0.104         0.010         0.200         
40            0.660         -0.033        55.468        0.010         0.261         
50            0.733         -0.074        37.082        0.010         0.327         
60            0.561         0.172         4.335         0.010         0.389         
70            0.558         0.003         723.906       0.010         0.453         
80            0.696         -0.138        463.292       0.010         0.516         
90            0.654         0.043         451.552       0.010         0.581         

--------------------------------------------------------------------------------
```

After:

```
Optimization Report
--------------------------------------------------------------------------------

Initial coordinates: 
 0.0000  0.0000  0.0000 ...
 0.0000 
Final coordinates: 
-2.3592 -0.0324 -0.0261 ...
-0.1400 
iter          loss          loss change   |gradient|    step size     total time    
0             0.883         0.000         4.532         0.010         0.023         
10            0.762         0.121         0.000         0.010         0.097         
20            1.195         -0.433        867.404       0.010         0.154         
30            0.627         0.567         0.104         0.010         0.215         
40            0.660         -0.033        55.468        0.010         0.279         
50            0.733         -0.074        37.082        0.010         0.346         
60            0.561         0.172         4.335         0.010         0.411         
70            0.558         0.003         723.906       0.010         0.475         
80            0.696         -0.138        463.292       0.010         0.536         
90            0.654         0.043         451.552       0.010         0.601         

--------------------------------------------------------------------------------
```

PR #386, opened 2023-11-26
Release version 2.20.1: "Stripped Bolt Head"

rcurtin:
This automatically-generated pull request adds the commits necessary to make the 2.20.1 release.

Once the PR is merged, mlpack-bot will tag the release as HEAD~1 (so that it doesn't include the new HISTORY block) and publish it.

Or, well, hopefully that will happen someday.

When you merge this PR, be sure to merge it using a *rebase*.

### Changelog

 * Clarify return values for different callback types ([#383](https://github.com/mlpack/ensmallen/pull/383)).

 * Fix return types of callbacks ([#382](https://github.com/mlpack/ensmallen/pull/382)).

 * Minor cleanup for printing optimization reports via `Report()` ([#385](https://github.com/mlpack/ensmallen/pull/385)).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,8 +108,8 @@ $ cd ensmallen
 
 # - or -
 
-$ wget http://ensmallen.org/files/ensmallen-2.20.0.tar.gz
-$ tar -xvzpf ensmallen-2.20.0.tar.gz
+$ wget http://ensmallen.org/files/ensmallen-2.20.1.tar.gz
+$ tar -xvzpf ensmallen-2.20.1.tar.gz
 $ cd ensmallen-latest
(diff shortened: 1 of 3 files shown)


conradsnicta:
@rcurtin Instead of a mere patch bump, suggest to change the version to 2.21.0 or even 2.30.0.

A patch bump usually indicates small bug fixes.  To my eyes at least, the callback mechanism has significant changes that go beyond that.

There is also a risk that the changes can introduce regressions in user code (user defined callbacks that we don't see), so it would be better to have a clearer delineation of the "old" and "new" callback mechanism within the ensmallen version number.

rcurtin:
Yeah, good point---let me do that instead.

PR #391, opened 2024-01-27
Coot 2: Change arma::function to function

shrit:
This PR handles most of Armadillo functions in ensmallen with a focus on functions that accept ADL with no issues.

I have also noticed that we are heavily using `arma::mat` and `arma::vec` in the library with is a bit annoying as we have MatType is most places. :unamused:


zoq:
https://github.com/mlpack/ensmallen/pull/352 does something similar but also makes sure, the tests work as expected.

shrit:
I will close this one, coot one already exist will continue the development there

PR #393, opened 2024-02-13
Release version 2.21.1: "Bent Antenna"

rcurtin:
This automatically-generated pull request adds the commits necessary to make the 2.21.1 release.

Once the PR is merged, mlpack-bot will tag the release as HEAD~1 (so that it doesn't include the new HISTORY block) and publish it.

Or, well, hopefully that will happen someday.

When you merge this PR, be sure to merge it using a *rebase*.

### Changelog

 * Fix numerical precision issues for small-gradient L-BFGS scaling factor computations ([#392](https://github.com/mlpack/ensmallen/pull/392)).

PR #394, opened 2024-02-14
ensure the tests are built with optimisation enabled

conradsnicta:
Ensure the tests are built with optimisation enabled.   Otherwise the tests take forever to run.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,9 @@
 # just installs the headers to the install location, and optionally builds the
 # test program.
 cmake_minimum_required(VERSION 3.3.2)
-set(CMAKE_BUILD_TYPE "Release")  # ensure the tests are built with optimisation
+if (NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE "Release")  # ensure the tests are built with optimisation
+endif ()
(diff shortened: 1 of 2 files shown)


conradsnicta:
@rcurtin As mentioned in https://github.com/mlpack/ensmallen/pull/393#issuecomment-1942955296

CMakeLists.txt
```diff
@@ -2,6 +2,7 @@
 # just installs the headers to the install location, and optionally builds the
 # test program.
 cmake_minimum_required(VERSION 3.3.2)
+set(CMAKE_BUILD_TYPE "Release")  # ensure the tests are built with optimisation
```
rcurtin, line 5: This could make it really hard to debug, though, and specifying `-DCMAKE_BUILD_TYPE` in the usual way would be overridden.  Maybe we can do this instead?
```suggestion
if (NOT CMAKE_BUILD_TYPE)
  set(CMAKE_BUILD_TYPE "Release")  # ensure the tests are built with optimisation
endif ()
```

conradsnicta: Good idea.

zoq (approved):

coatless (approved):

PR #398, opened 2024-03-23
catch2 workaround for recent issue under macOS

conradsnicta:
adapted from upstream catch2 repo: https://github.com/catchorg/Catch2/pull/2840


zoq (approved):
Looks good to me.

PR #400, opened 2024-06-08
update to C++14 standard

conradsnicta:
Update C++ standard to C++14.

mlpack 4.4 already requires C++17, and Armadillo 14.0 will use C++14.

The only OS that required C++11 is RHEL 7 / CentOS 7, which is EOL as of 30 June 2024:
* https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux
* https://www.redhat.com/en/topics/linux/centos-linux-eol
* https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/

Using "two most recent releases of long term Linux distros" as the yardstick for compiler support.  The Linux distros are: RHEL (+ clones like Rocky Linux and Alma Linux), Ubuntu LTS, and Debian.

The current two supported releases of RHEL (8 and 9), Ubuntu LTS (22.04 and 24.04), Debian (11 and 12) all provide GCC versions that support at least C++14.

RHEL 8 does not support C++17 properly, so not bumping the C++ standard to C++17.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,8 +13,8 @@ option(USE_OPENMP "If available, use OpenMP for parallelization." ON)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/CMake")
 
-# Set required C++ standard to C++11.
-set(CMAKE_CXX_STANDARD 11)
+# Set required C++ standard to C++14.
+set(CMAKE_CXX_STANDARD 14)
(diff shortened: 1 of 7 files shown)


zoq (approved):

shrit (approved):

coatless (approved):

rcurtin:
Awesome, I actually did not know about `[[deprecated]]`, nice cleanup.  I'll do the same for mlpack and we can get rid of `mlpack_deprecated` :smile:

PR #401, opened 2024-06-14
CMA stop conditions

beny-bee:
Hi,

I am currently enhancing the implementation of the CMA-ES algorithm by introducing additional stopping conditions. These modifications are intended to improve numerical stability and are advocated in [A Restart CMA Evolution Strategy With Increasing Population Size](http://www.cmap.polytechnique.fr/~nikolaus.hansen/cec2005ipopcmaes.pdf) and [The CMA Evolution Strategy: A Tutorial](https://arxiv.org/pdf/1604.00772). The changes are also in line with implementations found in [Python libraries](https://github.com/CMA-ES/pycma) maintained by the same author.

These are basically:
- [Implemented] **Maximum Generations (`maxiter`)**: Stop if the number of iterations reaches a predefined limit.
- [Implemented] **Target Objective Value (`ftarget`)**: Stop if the best objective value is less than or equal to a target value.
- [Implemented] **Maximum Function Evaluations (`maxfevals`)**: Stop if the number of function evaluations reaches a predefined limit.
- [Implemented] **Tolerance on Function Value (`tolfun`)**: Stop if the range of function values of the best solutions falls below a threshold over a certain number of iterations.
- [Implemented] **Tolerance on Fitness Stagnation (`tolflatfitness`)**: Stop if the best objective value does not change significantly over a number of generations.
- [Implemented] **Condition Number of Covariance Matrix (`tolconditioncov`)**: Stop if the condition number of the covariance matrix exceeds a certain threshold.
- **No Effect Axis (`noeffectaxis`)**: Stop if adding 0.1 times the standard deviation in the search space in a principal axis direction does not change the solution.
- **No Effect Coordinate (`noeffectcoord`)**: Stop if adding 0.2 times the standard deviation in the search space along one coordinate does not change the solution.
- **Tolerance on Variables (`tolx`)**: Stop if the change in the variables between generations falls below a threshold.
- **Tolerance on Upscaled Sigma (`tolfacupx`)**: Stop if any scaled step sizes exceed the product of the initial step sizes and a specified tolerance factor.
- **Historical Fitness Tolerance (`tolfunhist`)**: Stop if the range of fitness values in the historical best fitness list remains below a specified threshold.
- **Sigma Ratio Increase (`tolupsigma`)**: Stop if the ratio of the current step size to the maximum eigenvalue of the covariance matrix exceeds a specified multiple of the initial step size.
- **Stagnation (`tolstagnation`)**: Stop if there is no significant improvement over a large number of generations.
[...]

PR #373, opened 2023-07-10
Implementation of IPOP CMAES

SuvarshaChennareddy:
This pull request is an initial implementation of IPOP CMAES, which is part of my Google Summer of Code (GSoC) project. Tests have been added, but not all of them are passing.

PR #403, opened 2024-07-07
IPOP and BIPOP restart mechanisms

beny-bee:
Hi everyone!

I've taken a break from the termination criteria and have been implementing restart mechanisms for CMA-ES. The plan was to do IPOP, but adding BIPOP was quite simple so I suggest adding it too.

[**IPOP**](http://www.cmap.polytechnique.fr/~nikolaus.hansen/cec2005ipopcmaes.pdf): is an extension of the standard CMA-ES that features an increasing population size strategy, where at each termination of the optimizer the population is increased by a constant factor. This is based on the initial work found in [PR #373](https://github.com/mlpack/ensmallen/pull/373).

[**BIPOP**](https://dl.acm.org/doi/pdf/10.1145/1570256.1570333): is pretty straightforward after setting up IPOP. BIPOP uses two interlaced strategies: one with an increase in population size with a constant factor (identical to IPOP) and another with a smaller, varying population size. I've seen some places that this strategy works better on certain functions.

Additionally, there are two things to consider:
- Termination Criteria and API Adjustments: Depending on the [termination criteria](https://github.com/mlpack/ensmallen/pull/401), we might need to introduce additional parameters for finer control. Thus, I know there are more parameters we can add on IPOP and BIPOP, but i would suggest waiting until the termination criteria are done to add all.
- Testing: I have copy pasted some of the tests from the other CMA-ES, but feel free to suggest more.


include/ensmallen_bits/cmaes/bipop_cmaes.hpp
```diff
@@ -0,0 +1,147 @@
...
+   * are processed (i.e., one iteration equals one point; one iteration does not
+   * equal one pass over the dataset).
+   *
+   * @param CMAES The CMAES object.
+   * @param populationFactor The factor by which population increases
+   *      after each restart.
+   * @param maxRestarts Maximum number of restarts.
+   */
+  BIPOPCMAES(const CMAESType& CMAES = CMAESType(),
+            const size_t maxRestarts = 9);
```
zoq, line 69:
```suggestion
  BIPOPCMAES(const CMAESType& CMAES = CMAESType(),
             const size_t maxRestarts = 9);
```

include/ensmallen_bits/cmaes/bipop_cmaes_impl.hpp
```diff
@@ -0,0 +1,113 @@
...
+            iterate = iterateIn;
+            cmaes.Optimize(function, iterate, sbc, callbacks...);
+            
+            size_t evaluations = cmaes.FunctionEvaluations();
+            totalFunctionEvaluations += evaluations;
+            smallPopulationBudget += evaluations;
+        }
+
+        // Check if the total number of evaluations has exceeded the limit
+        if (totalFunctionEvaluations >= 1e9) {
```
zoq, line 100: Should this be configurable?

beny-dl: Yeah, most likely. Plus there should be more parameters to control the termination criteria of the underlying CMA (from the other PR) which I will need to add. But since I still need to review the correctness of the termination criteria I don't want to add/delete stuff now.

rcurtin:
Don't forget to add documentation to `doc/`. Without documentation nobody will ever know about these features.

beny-dl:
Completely forgot about the documentation, added it now.

include/ensmallen_bits/cmaes/bipop_cmaes_impl.hpp
zoq: Okay, makes sense.

rcurtin (commented):
Thanks for adding documentation, just a couple comments about design and style.  I hope to get the automatic style checker job online sometime, but until then if you don't mind going through and making sure everything matches I would appreciate it (and that would make it easier to fix that CI job whenever it does come online)

include/ensmallen_bits/cmaes/ipop_cmaes.hpp
```diff
@@ -0,0 +1,159 @@
...
+   * maximum number of iterations refers to the maximum number of points that
+   * are processed (i.e., one iteration equals one point; one iteration does not
+   * equal one pass over the dataset).
+   *
+   * @param CMAES The CMAES object.
+   * @param populationFactor The factor by which population increases
+   *      after each restart.
+   * @param maxRestarts Maximum number of restarts.
+   */
+  IPOPCMAES(const CMAESType& CMAES = CMAESType(),
```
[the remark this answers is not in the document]
beny-dl: Both IPOP and BIPOP are restart mechanisms for the underlying CMAES. They are originally defined as extensions for the original "vanilla" CMAES, however, the restarts could be done for any variant. My intent was to have a CMAES object so the user could do the restart mechanism on whatever CMAES they wanted ("vanilla" CMAES or `ActiveCMAES`), which is why the user would input a CMAES object.

Nevertheless, I could just initialize whichever CMAES the user wants inside with the constructor you suggest. Or just ignore the different CMAES variants and just use the "vanilla" CMAES always.

What do you suggest?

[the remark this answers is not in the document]
rcurtin: The way you described it there, does it make more sense to simply modify CMAES to add another template parameter that represents the restart mechanism?  That could simplify the code significantly.

From a user perspective, regardless of what choice you make internally, I do think it's important to make the constructor "feel" the same as the other CMAES variants, whether that is done by having IPOP-CMAES and BIPOP-CMAES be typedefs of CMAES with certain template parameters, or by holding a CMAES object internally.  So, e.g., a constructor taking each of the same arguments as the CMAES constructor (or however many of those arguments are applicable).

Hopefully that makes sense, let me know if I can clarify anything I wrote. :+1:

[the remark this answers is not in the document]
beny-dl: > The way you described it there, does it make more sense to simply modify CMAES to add another template parameter that represents the restart mechanism? That could simplify the code significantly.

In this case, you mean adding something like `template  typename SelectionPolicyType = FullSelection,  typename TransformationPolicyType = EmptyTransformation<>,  typename RestartPolicyType = NoRestart>` instead of having what I did right now? 

Then I would need to modify the CMAES_impl.hpp `Optimize` to have something along the lines of `restartPolicy.shouldRestart()` and `restartPolicy.restart()`? Where the restart policy could either be BIPOP or IPOP or new policies.

If that's the case, I see what you mean, and it might indeed be cleaner. 

But currently, I'm using the `Optimize` from the CMAES, and when it terminates, that elicits a restart. I would have to change quite a bit of code so that instead of returning when there is a stop condition, it loops until the restarts are done.

I'm fine doing it either way, but I might need to think about how to do it.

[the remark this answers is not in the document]
zoq: This could be handled by the callback feature, that said, I don't want to make the code too complex. Instead should we just create an independent IPOP implementation, that does not reuse CMAES? BIPOP could be a flag to the IPOP implementation. What do you think?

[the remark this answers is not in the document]
rcurtin: Right, I had sort of been thinking about adding a `RestartPolicyType` template parameter to CMAES.  But an independent IPOP implementation would maybe be the better thing.  Honestly I don't have a huge opinion either way (and I don't deeply understand the algorithm), my perspective is entirely user-facing.  So as long as the class "feels the same" as other optimizers to use, I'm happy. :smile:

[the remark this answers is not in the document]
beny-dl: > BIPOP could be a flag to the IPOP implementation. 

I can definitely reduce both into one, with a flag to choose. 

As for the code, I'm also fine doing either way, I feel like creating an instance of CMAES inside the IPOP and then use `Optimize` might be the best though. This way the user can control all the parameters in the same fashion, while also not duplicating code, and any improvement we do to the CMAES will be reflected in the IPOP (and BIPOP).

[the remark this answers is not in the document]
beny-dl: @zoq @rcurtin I refactored it into one code only (POP-CMA-ES) where with a flag it launches BIPOP or IPOP. I also tried to make it as similar to the other CMAES optimizers. However, now instead of passing it an CMAES object, I just initializes the CMAES class inside and use it.

Let me know if you prefer this version, and if you do, i'll write out the documentation for this and should be pretty much ready.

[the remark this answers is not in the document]
rcurtin: I like it, thanks for taking the effort to do that. Two quick notes:

 * Feel free to use underscores in the class names, like `POP_CMAES`; this can make the class name a little more decipherable. :)
 * If you want to keep the BIPOP and IPOP versions as separate optimizers, you could make the `useBIPOP` parameter into a template parameter and use a `using` statement to define the `IPOPCMAES` and `BIPOPCMAES` classes. Up to you whether you want to do that---it really hinges on whether you want to provide to the user a single class `POPCMAES` or two classes. (I guess two classes would be more in line with how ensmallen is structured, with one optimizer per class, but either way is fine with me.)

Thanks again for the effort; at least from my side I have no other major comments. :+1:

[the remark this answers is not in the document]
zoq: All the comments make sense to me.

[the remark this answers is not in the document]
beny-dl: Done the suggestions @rcurtin!

Going to mark this no longer as draft

include/ensmallen_bits/cmaes/pop_cmaes.hpp
```diff
@@ -0,0 +1,175 @@
...
+           const TransformationPolicyType& transformationPolicy = 
+              TransformationPolicyType(),
+           const size_t batchSize = 32,
+           const size_t maxIterations = 1000,
+           const double tolerance = 1e-5,
+           const SelectionPolicyType& selectionPolicy = SelectionPolicyType(),
+           double stepSize = 0,
+           const double populationFactor = 2,
+           const size_t maxRestarts = 9,
+           const size_t maxFunctionEvaluations = 1e9);
```
rcurtin, line 119:
```suggestion
            const TransformationPolicyType& transformationPolicy = 
                 TransformationPolicyType(),
            const size_t batchSize = 32,
            const size_t maxIterations = 1000,
            const double tolerance = 1e-5,
            const SelectionPolicyType& selectionPolicy = SelectionPolicyType(),
            double stepSize = 0,
            const double populationFactor = 2,
            const size_t maxRestarts = 9,
            const size_t maxFunctionEvaluations = 1e9);
```
Super trivial style comment that I saw when looking through the new design.  It looks great to me :+1:

doc/optimizers.md
```diff
@@ -778,6 +780,72 @@ optimizer2.Optimize(f, coordinates);
  * [SGD in Wikipedia](https://en.wikipedia.org/wiki/Stochastic_gradient_descent)
  * [SGD](#standard-sgd)
 
+## BIPOP_CMAES
```
rcurtin, line 783: One more documentation bit I realized; can you add BIPOP_CMAES and IPOP_CMAES to the list of optimizers for separable functions in `function_types.md`?  Thanks :+1:

beny-dl: Just did that, also added ActiveCMAES which wasn't added.

Also, had to increase a bit the margin of the tests as in 2 environments it was failing

zoq: Looks reasonable.

zoq:

rcurtin:
@beny-bee @zoq does anything else need to be done here?  It looks to me like this one is pretty much ready to go, but I don't know the state of it.  It would be great to get it merged. :+1:

beny-bee:
@rcurtin should be ready to be merged!

conradsnicta:
@rcurtin @zoq This might be ready to go in.  Given this and the other changes that have been accumulating since February, it might be time to release the next version of ensmallen.

doc/optimizers.md
```diff
@@ -1575,6 +1646,73 @@ optimizer.Optimize(f, coordinates);
...
+| `size_t` | **`maxIterations`** | Maximum number of iterations. | `1000` |
+| `double` | **`tolerance`** | Maximum absolute tolerance to terminate the algorithm. | `1e-5` |
+| `SelectionPolicyType` | **`selectionPolicy`** | Instantiated selection policy used to calculate the objective. | `SelectionPolicyType()` |
+| `double` | **`stepSize`** | Initial step size. | `0` |
+| `size_t` | **`maxRestarts`** | Maximum number of restarts. | `9` |
+| `double` | **`populationFactor`** | The factor by which the population increases after each restart. | `2.0` |
+| `size_t` | **`maxFunctionEvaluations`** | Maximum number of function evaluations. | `1e9` |
+
+The _`SelectionPolicyType`_ and _`TransformationPolicyType`_ template parameters are used to control how the selection of points and the transformation of the decision variables are handled during the optimization process. For more information, check the [CMAES](#cmaes) class. The `lambda`, `transformationPolicy`, `batchSize`, `maxIterations`, `tolerance`, `selectionPolicy`, and `stepSize` are all parameters that affect the underlying CMAES process.
```
rcurtin, line 1687:
```suggestion
The _`SelectionPolicyType`_ and _`TransformationPolicyType`_ template parameters
are used to control how the selection of points and the transformation of the
decision variables are handled during the optimization process. For more
information, check the [`CMAES`](#cma-es) class. The `lambda`,
`transformationPolicy`, `batchSize`, `maxIterations`, `tolerance`,
`selectionPolicy`, and `stepSize` are all parameters that affect the underlying
CMA-ES process.

For convenience, the following type can be used:

 * **`IPOP_CMAES<>`** (equivalent to `IPOP_CMAES<FullSelection, EmptyTransformation<>>`): uses all separable functions to compute objective
```

I also added the default typedef here.

doc/optimizers.md
```diff
@@ -1575,6 +1646,73 @@ optimizer.Optimize(f, coordinates);
...
+
+<details open>
+<summary>Click to collapse/expand example code.
+</summary>
+
+```c++
+RosenbrockFunction f;
+arma::mat coordinates = f.GetInitialPoint();
+
+// IPOP Active CMA-ES
```
rcurtin, line 1699:
```suggestion
// IPOP CMA-ES
```
Seems like this was a typo?

doc/optimizers.md
```diff
@@ -1575,6 +1646,73 @@ optimizer.Optimize(f, coordinates);
...
+Optimizer.Optimize(f, coordinates);
+```
+
+</details>
+
+#### See also:
+
+ * [CMAES](#cmaes)
+ * [ActiveCMAES](#activecmaes)
+ * [BIPOP_CMAES](#bipopcmaes)
```
rcurtin, line 1711:
```suggestion
 * [CMA-ES](#cmaes)
 * [Active CMA-ES](#active-cma-es)
 * [BIPOP CMA-ES](#bipop-cma-es)
```

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -172,6 +172,15 @@ class CMAES
...
 
+  //! Get the total number of function evaluations.
+  size_t FunctionEvaluations() const 
+  { return functionEvaluations; }
+
+  //! Type of Selection Policy.
+  typedef SelectionPolicyType selectionPolicyType;
+  //! Type of Transformation Policy.
+  typedef TransformationPolicyType transformationPolicyType;
+
```
rcurtin, line 183:
```suggestion
```

I don't see these used anywhere---I think we can remove them.

include/ensmallen_bits/cmaes/cmaes.hpp
```diff
@@ -172,6 +172,15 @@ class CMAES
   double& StepSize()
   { return stepSize; }
 
+  //! Get the total number of function evaluations.
+  size_t FunctionEvaluations() const 
+  { return functionEvaluations; }
```
rcurtin, line 177:
```suggestion
  size_t FunctionEvaluations() const  { return functionEvaluations; }
```
This will fit all on one line.

tests/pop_cmaes_test.cpp
```diff
@@ -0,0 +1,133 @@
+#include <ensmallen.hpp>
```
rcurtin, line 1:
```suggestion
/**
 * @file pop_cmaes_test.cpp
 * @author Benjami Parellada
 *
 * Tests for the POP_CMAES class, including IPOP CMA-ES and BIPOP CMA-ES.
 *
 * ensmallen is free software; you may redistribute it and/or modify it under
 * the terms of the 3-clause BSD license.  You should have received a copy of
 * the 3-clause BSD license along with ensmallen.  If not, see
 * http://www.opensource.org/licenses/BSD-3-Clause for more information.
 */
#include <ensmallen.hpp>
```
We should include a license header here too.  (I put your name here just as a guess, but you could remove it if for some reason you want to :))

include/ensmallen_bits/cmaes/pop_cmaes_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_CMAES_POP_CMAES_IMPL_HPP
+#define ENSMALLEN_CMAES_POP_CMAES_IMPL_HPP
+
+#include "pop_cmaes.hpp"
+#include <ensmallen_bits/function.hpp>
+#include <random>
```
rcurtin, line 22:
```suggestion
```

We can actually use Armadillo's RNGs directly in the class, and avoid the inclusion of random.  See the comments below.

include/ensmallen_bits/cmaes/pop_cmaes_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+  StoreBestCoordinates<MatType> sbc;
+  StoreBestCoordinates<MatType> overallSBC;
+  size_t totalFunctionEvaluations = 0;
+  size_t largePopulationBudget = 0;
+  size_t smallPopulationBudget = 0;
+
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_real_distribution<> dis(0.0, 1.0);
+
```
rcurtin, line 70:
```suggestion
```

(Another comment lower down will adapt the usage of `dis` to using Armadillo's `randu()` function.)

include/ensmallen_bits/cmaes/pop_cmaes_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+      evaluations = this->FunctionEvaluations();
+      largePopulationBudget += evaluations;
+    }
+    else if (UseBIPOPFlag)
+    {
+      // Small population regime (BIPOP only)
+      double u = dis(gen);
+      size_t smallLambda = static_cast<size_t>(defaultLambda * std::pow(0.5 * 
+          currentLargeLambda / defaultLambda, u * u));
+      double stepSizeSmall = 2 * std::pow(10, -2*dis(gen));
```
rcurtin, line 120:
```suggestion
      double u = arma::randu<double>();
      size_t smallLambda = static_cast<size_t>(defaultLambda * std::pow(0.5 * 
          currentLargeLambda / defaultLambda, u * u));
      double stepSizeSmall = 2 * std::pow(10, -2 * arma::randu<double>());
```

include/ensmallen_bits/cmaes/pop_cmaes_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+      
+      this->PopulationSize() = smallLambda;
+      this->StepSize() = stepSizeSmall;
+
+      Info << "BIPOP-CMA-ES: restart " << restart << ", small population" <<
+          " size (lambda): " << this->PopulationSize() << std::endl;
```
rcurtin, line 126:
```suggestion
          " size (lambda): " << this->PopulationSize() << "." << std::endl;
```

include/ensmallen_bits/cmaes/pop_cmaes_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+      evaluations = this->FunctionEvaluations();
+      smallPopulationBudget += evaluations;
+    }
+
+    if (objective < overallObjective)
+    {
+      overallObjective = objective;
+      overallSBC = sbc;
+      Info << "POP-CMA-ES: New best objective: " << overallObjective << 
+          std::endl;
```
rcurtin, line 144:
```suggestion
      Info << "POP-CMA-ES: New best objective: " << overallObjective
          << "." << std::endl;
```

include/ensmallen_bits/cmaes/pop_cmaes_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+    if (!UseBIPOPFlag || largePopulationBudget <= smallPopulationBudget || 
+        restart == 0 || restart == maxRestarts - 1)
+    {
+      // Large population regime (IPOP or BIPOP)
+      currentLargeLambda *= populationFactor;
+      this->PopulationSize() = currentLargeLambda;
+      this->StepSize() = stepSizeDefault;
+
+      Info << "POP-CMA-ES: restart " << restart << ", large population size" <<
+          " (lambda): " << this->PopulationSize() << std::endl;
```
rcurtin, line 102:
```suggestion
          " (lambda): " << this->PopulationSize() << "." << std::endl;
```

include/ensmallen_bits/cmaes/pop_cmaes_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+  ElemType objective;
+  size_t evaluations;
+
+  size_t defaultLambda = this->PopulationSize();
+  size_t currentLargeLambda = defaultLambda;
+
+  double stepSizeDefault = this->StepSize();
+
+  // Print out the default population size
+  Info << "Default population size: " << defaultLambda << std::endl;
```
rcurtin, line 87:
```suggestion
  Info << "Default population size: " << defaultLambda << "." << std::endl;
```
I know this is super pedantic.  A couple similar comments below.

include/ensmallen_bits/cmaes/pop_cmaes_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    totalFunctionEvaluations += evaluations;
+    // Check if the total number of evaluations has exceeded the limit
+    if (totalFunctionEvaluations >= maxFunctionEvaluations) {
+      Warn << "POP-CMA-ES: Maximum function overall evaluations reached. "
+           << "terminating optimization." << std::endl;
+
+      Callback::EndOptimization(*this, function, iterate, callbacks...);
+      iterateIn = overallSBC.BestCoordinates();
```
rcurtin, line 154:
```suggestion
      iterateIn = std::move(overallSBC.BestCoordinates());
```
We can probably avoid the copy of the data here.

include/ensmallen_bits/cmaes/pop_cmaes_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+      return overallSBC.BestObjective();
+    }
+
+    ++restart;
+  }
+
+  Callback::EndOptimization(*this, function, iterate, callbacks...);
+  iterateIn = overallSBC.BestCoordinates();
```
rcurtin, line 162:
```suggestion
  iterateIn = std::move(overallSBC.BestCoordinates());
```
Another copy we can avoid.

rcurtin (commented):
Hey @beny-bee, sorry it took so long to get back to this.  I believe the implementation is correct and I think everything is good to go.  However I have a bunch of small comments.  I've tried to make them as simple to address as possible---I think all of them are suggestions.  If you disagree on any of the suggestions, feel free to discuss, mostly my goal is just to make things consistent throughout.  And if you prefer, I can just accept all the suggestions as part of the merge.  The only functionality change is the RNG changes, but I did test those locally with 1000 different random seeds and saw no test failures, so it seems fine.

Last thing would be to update HISTORY.md before merge, but like I said I can handle that too if you like. :)

coatless:
@conradsnicta let's wait until 2.22 launches

beny-bee:
Thank you for the suggestions, @rcurtin, and for handling the suggestions, @conradsnicta. I apologize for the delay in my response; I've been quite busy recently.

I'm fine with all the style changes, as well as the RNG one. Additionally, I tested them locally and everything works as before. 

I updated the HISTORY.md accordingly.

Thanks again for your support!

zoq (approved):

rcurtin:
I pushed one more changes for the "CMA-ES" name changes to the documentation.  Otherwise, I think this is ready for merge so I'll merge once there is a green build.

conradsnicta (approved):

PR #413, opened 2024-11-30
Bugfix/unused variables

coatless:
This PR tackles #412 by removing unused variables. 

I've also modified `CMakeLists.txt` to warn if it detects unused variables and private fields. (It's already warning on unused parameters!)

PR #414, opened 2024-12-02
Add an entry in the HISTORY for #413

rcurtin:
I noticed that there was not a line in HISTORY.md for #413, so I added one.  (This will mean that the changelog won't be empty when I open a PR to release 2.22.1 with that fix.)


zoq (approved):

eddelbuettel (approved):
:+1:

PR #419, opened 2025-02-25
fix include

conradsnicta:
Fix include statement to be consistent with other tests.
This is to resolve the following bug:

> include path
> The C++ test files must #include <ensmallen.hpp>
> rather than "../include/ensmallen.hpp"
> in order to allow the installed files, rather than the repo-local files, to be tested.
> 
> https://sources.debian.org/src/ensmallen/2.22.1-1/debian/patches/0001-include-path.patch/

@rcurtin @shrit @zoq @barak @eddelbuettel

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,8 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
+ * Fix include statement in `tests/de_test.cpp`
+   ([#419](https://github.com/mlpack/ensmallen/pull/419)).
+
  * Fix `exactObjective` output for SGD-like optimizers when the number of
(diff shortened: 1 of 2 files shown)


HISTORY.md
```diff
@@ -1,5 +1,8 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
```
eddelbuettel, line 2: Extra newline?

eddelbuettel (approved):
LGTM.  Not entirely sure why it is needed as during a Debian package build both location should contain the same files (as the system build is from the repo sources).  Maybe I am missing something....

Issue #418, opened 2025-02-12
Test failures on some architectures (32-bit ARM)


barak:
#### Issue description

Test failures on some architectures

#### Your environment

 * version of ensmallen: 2.22.1
 * operating system: Debian
 * compiler: GCC 14.2.0
 * version of Armadillo: 14.2.1
 * any other environment information you think is relevant: autobuilders/autotesters

#### Steps to reproduce

See:
* https://tracker.debian.org/pkg/ensmallen (section Testing Migrations, "Issues Preventing Migration")
* https://ci.debian.net/packages/e/ensmallen/testing/armel/57602218/ (arch: armel error log)
* https://ci.debian.net/packages/e/ensmallen/testing/armhf/57617012/ (arch: armhf error log)

The error is:
```

916s 0.023 s: DemonAdaMaxLogisticRegressionTest
916s 
916s ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
916s ensmallen-test-0wVsef is a Catch v2.13.10 host application.
916s Run with -? for options
916s 
916s -------------------------------------------------------------------------------
916s DemonSGDLogisticRegressionTest
916s -------------------------------------------------------------------------------
916s tests/demon_sgd_test.cpp:22
916s ...............................................................................
916s 
916s tests/test_function_tools.hpp:215: FAILED:
916s   REQUIRE( acc == Approx(100.0).epsilon(trainAccuracyTolerance) )
916s with expansion:
916s   99.4 == Approx( 100.0 )
```

#### Expected behavior

Pass all tests

#### Actual behavior

Fails one test

conradsnicta:
@barak Proposed fix is in #420. Can you see if this passes on 32 bit armel and armhf?

barak:
Right, I get that. So why was the test failing on the 32-bit systems and passing on the 64-bit ones? It's a mystery!

I am not sure lowering the tolerance is the right thing to do. As a short-term fix, fine sure. But somebody should track down what's going on.

Here's a wild-ass guess. Perhaps there is some sort of conversion issue where a number is being converted from a float (or double) to an int, and back, or something like that. Something that would cause a systematic *numeric* difference on 32-bit-pointer vs 64-bit-pointer (and corresponding difference in length of long int) machines. This might even happen implicitly if a double is being passed as an argument to a procedure that expects an int, or something like that.

[...]

PR #420, opened 2025-02-25
increase tolerance

conradsnicta:
Increase tolerance for tests failing on 32 bit armel and armhf, aiming to address bug report #418.

PS. The form `Approx(x).epsilon(y)` as used by ensmallen tests is deprecated by Catch2, as it has several problems. The corresponding docs suggest to use `WithinAbs()` and `WithinRel`, as `Approx()` cannot be fixed without breaking backwards compatibility.  This would require rewriting all of the tests in ensmallen, which is beyond the scope of this PR.

https://github.com/catchorg/Catch2/blob/devel/docs/comparing-floating-point-numbers.md#top

@rcurtin @shrit @zoq @barak @eddelbuettel

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,6 +4,9 @@
    iterations is an even number of epochs
    ([#417](https://github.com/mlpack/ensmallen/pull/417)).
 
+ * Increase tolerance in `demon_sgd_test.cpp`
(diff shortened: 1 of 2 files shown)


eddelbuettel (approved):
LGTM as a narrow and focused change addressing the one pertinent issue.

PR #425, opened 2025-05-08
Bandicoot intergration

zoq:
Integrating [Bandicoot](https://coot.sourceforge.io/) into ensmallen, so we can build against [Armadillo](https://arma.sourceforge.net/) and Bandicoot.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,16 @@ if(USE_OPENMP)
   target_link_libraries(ensmallen INTERFACE OpenMP::OpenMP_CXX)
 endif()
 
+# Find Bandicoot and link it.
+find_package(Bandicoot 0.100.0)
+if(BANDICOOT_FOUND AND NOT TARGET Bandicoot:Bandicoot)
+  target_link_libraries(ensmallen INTERFACE Bandicoot::Bandicoot)
+  target_include_directories(ensmallen INTERFACE ${BANDICOOT_INCLUDE_DIR})
+
+  set(USE_COOT "1")
+  add_definitions(-DUSE_COOT)
+endif()
+
 # Find Armadillo and link it.
 find_package(Armadillo 10.8.2 REQUIRED)
 target_link_libraries(ensmallen INTERFACE Armadillo::Armadillo)
--- a/include/ensmallen.hpp
+++ b/include/ensmallen.hpp
@@ -34,6 +34,10 @@
 
 #include <armadillo>
 
+#ifdef USE_COOT
+  #include <bandicoot>
+#endif
+
 #if ((ARMA_VERSION_MAJOR < 10) || ((ARMA_VERSION_MAJOR == 10) && (ARMA_VERSION_MINOR < 8)))
   #error "need Armadillo version 10.8 or newer"
 #endif
--- a/CMake/COOT_FindCLBLAS.cmake
+++ b/CMake/COOT_FindCLBLAS.cmake
@@ -1,0 +1,44 @@
+# - Find clBLAS (includes and library)
+# This module defines
+#  CLBLAS_INCLUDE_DIR
+#  CLBLAS_LIBRARIES
+#  CLBLAS_FOUND
+# also defined, but not for general use are
+#  CLBLAS_LIBRARY, where to find the library.
+
+find_path(CLBLAS_INCLUDE_DIR clBLAS.h
+/usr/include/
+/usr/local/include/
+)
+
+set(CLBLAS_NAMES ${CLBLAS_NAMES} clBLAS)
+find_library(CLBLAS_LIBRARY
+  NAMES ${CLBLAS_NAMES}
+  PATHS /usr/lib64/ /usr/local/lib64/ /usr/lib /usr/local/lib
+  )
+
+if (CLBLAS_LIBRARY AND CLBLAS_INCLUDE_DIR)
+    set(CLBLAS_LIBRARIES ${CLBLAS_LIBRARY})
+    set(CLBLAS_FOUND "YES")
+else ()
+  set(CLBLAS_FOUND "NO")
+endif ()
+
+if (CLBLAS_FOUND)
+   if (NOT CLBLAS_FIND_QUIETLY)
+      message(STATUS "Found a clBLAS library: ${CLBLAS_LIBRARIES}")
+   endif ()
+else ()
+    if (CLBLAS_FIND_REQUIRED)
+      message(FATAL_ERROR "Could not find a clBLAS library")
+   endif ()
+endif ()
+
+# Deprecated declarations.
+set (NATIVE_CLBLAS_INCLUDE_PATH ${CLBLAS_INCLUDE_DIR} )
+get_filename_component (NATIVE_CLBLAS_LIB_PATH ${CLBLAS_LIBRARY} PATH)
+
+mark_as_advanced(
+  CLBLAS_LIBRARY
+  CLBLAS_INCLUDE_DIR
+  )
--- a/CMake/COOT_FindCLBlast.cmake
+++ b/CMake/COOT_FindCLBlast.cmake
@@ -1,0 +1,44 @@
+# - Find clBlast (includes and library)
+# This module defines
+#  CLBLAST_INCLUDE_DIR
+#  CLBLAST_LIBRARIES
+#  CLBLAST_FOUND
+# also defined, but not for general use are
+#  CLBLAST_LIBRARY, where to find the library.
+
+find_path(CLBLAST_INCLUDE_DIR clblast.h
+/usr/include/
+/usr/local/include/
+)
+
+set(CLBLAST_NAMES ${CLBLAST_NAMES} clblast)
+find_library(CLBLAST_LIBRARY
+  NAMES ${CLBLAST_NAMES}
+  PATHS /usr/lib64/ /usr/local/lib64/ /usr/lib /usr/local/lib
+  )
+
+if (CLBLAST_LIBRARY AND CLBLAST_INCLUDE_DIR)
+    set(CLBLAST_LIBRARIES ${CLBLAST_LIBRARY})
+    set(CLBLAST_FOUND "YES")
+else ()
+  set(CLBLAST_FOUND "NO")
+endif ()
+
+if (CLBLAST_FOUND)
+   if (NOT CLBLAST_FIND_QUIETLY)
+      message(STATUS "Found a clBlast library: ${CLBLAST_LIBRARIES}")
+   endif ()
+else ()
+    if (CLBLAST_FIND_REQUIRED)
+      message(FATAL_ERROR "Could not find a clBlast library")
+   endif ()
+endif ()
+
+# Deprecated declarations.
+set (NATIVE_CLBLAST_INCLUDE_PATH ${CLBLAST_INCLUDE_DIR} )
+get_filename_component (NATIVE_CLBLAST_LIB_PATH ${CLBLAST_LIBRARY} PATH)
+
+mark_as_advanced(
+  CLBLAST_LIBRARY
+  CLBLAST_INCLUDE_DIR
+  )
--- a/CMake/COOT_FindNVRTC.cmake
+++ b/CMake/COOT_FindNVRTC.cmake
@@ -1,0 +1,35 @@
+# - Find clBlast (includes and library)
+# This module defines
+#  CLBLAST_INCLUDE_DIR
+#  CLBLAST_LIBRARIES
+#  CLBLAST_FOUND
+# also defined, but not for general use are
+#  CLBLAST_LIBRARY, where to find the library.
+
+set(NVRTC_NAMES ${NVRTC_NAMES} nvrtc)
+find_library(NVRTC_LIBRARY
+  NAMES ${NVRTC_NAMES}
+  PATHS /usr/lib64/ /usr/local/lib64/ /usr/lib /usr/local/lib /usr/lib/x86_64-linux-gnu/
+  )
+
+if (NVRTC_LIBRARY)
+  set(NVRTC_LIBRARIES ${NVRTC_LIBRARY})
+  set(NVRTC_FOUND "YES")
+else ()
+  set(NVRTC_FOUND "NO")
+endif ()
+
+if (NVRTC_FOUND)
+  if (NOT NVRTC_FIND_QUIETLY)
+    message(STATUS "Found NVRTC library: ${NVRTC_LIBRARIES}")
+   endif ()
+else ()
+  if (NVRTC_FIND_REQUIRED)
+    message(FATAL_ERROR "Could not find NVRTC library")
+   endif ()
+endif ()
+
+# Deprecated declarations.
+get_filename_component (NATIVE_NVRTC_LIB_PATH ${NVRTC_LIBRARY} PATH)
+
+mark_as_advanced(NVRTC_LIBRARY)
--- a/CMake/FindBandicoot.cmake
+++ b/CMake/FindBandicoot.cmake
@@ -1,0 +1,333 @@
+# - Find Bandicoot
+# Find Bandicoot: GPU accelerator add-on for the Armadillo C++ linear algebra
+# library
+#
+# Using Bandicoot:
+#  find_package(Bandicoot REQUIRED)
+#  include_directories(${BANDICOOT_INCLUDE_DIRS})
+#  add_executable(foo foo.cc)
+#  target_link_libraries(foo ${BANDICOOT_LIBRARIES})
+# This module sets the following variables:
+#  BANDICOOT_FOUND - set to true if the library is found
+#  BANDICOOT_INCLUDE_DIRS - list of required include directories
+#  BANDICOOT_LIBRARIES - list of libraries to be linked
+#  BANDICOOT_VERSION_MAJOR - major version number
+#  BANDICOOT_VERSION_MINOR - minor version number
+#  BANDICOOT_VERSION_PATCH - patch version number
+#  BANDICOOT_VERSION_STRING - version number as a string (ex: "1.0.4")
+#  BANDICOOT_VERSION_NOTE - name of the version (ex: "unstable development version")
+
+find_path(BANDICOOT_INCLUDE_DIR
+  NAMES bandicoot
+  PATHS "$ENV{ProgramFiles}/Bandicoot/include"
+  )
+
+if(BANDICOOT_INCLUDE_DIR)
+  # Extract version information.
+  file(READ "${BANDICOOT_INCLUDE_DIR}/bandicoot_bits/coot_version.hpp" _bandicoot_HEADER_CONTENTS)
+  string(REGEX REPLACE ".*#define COOT_VERSION_MAJOR ([0-9]+).*" "\\1" BANDICOOT_VERSION_MAJOR "${_bandicoot_HEADER_CONTENTS}")
+  string(REGEX REPLACE ".*#define COOT_VERSION_MINOR ([0-9]+).*" "\\1" BANDICOOT_VERSION_MINOR "${_bandicoot_HEADER_CONTENTS}")
+  string(REGEX REPLACE ".*#define COOT_VERSION_PATCH ([0-9]+).*" "\\1" BANDICOOT_VERSION_PATCH "${_bandicoot_HEADER_CONTENTS}")
+  string(REGEX REPLACE ".*#define COOT_VERSION_NOTE\ +\"([0-9a-zA-Z\ _-]+)\".*" "\\1" BANDICOOT_VERSION_NOTE "${_bandicoot_HEADER_CONTENTS}")
+
+  set(BANDICOOT_VERSION_STRING "${BANDICOOT_VERSION_MAJOR}.${BANDICOOT_VERSION_MINOR}.${BANDICOOT_VERSION_PATCH}")
+endif ()
+
+# Determine what support libraries are being used, and whether or not we need to
+# link against them.  We need to look in config.hpp.
+set(SUPPORT_INCLUDE_DIRS "")
+set(SUPPORT_LIBRARIES "")
+set(COOT_NEED_LIBRARY true) # Assume true.
+if(EXISTS "${BANDICOOT_INCLUDE_DIR}/bandicoot_bits/config.hpp")
+  file(READ "${BANDICOOT_INCLUDE_DIR}/bandicoot_bits/config.hpp" _bandicoot_CONFIG_CONTENTS)
+  # COOT_USE_WRAPPER
+  string(REGEX MATCH "\r?\n[\t ]*#define[ \t]+COOT_USE_WRAPPER[ \t]*\r?\n" COOT_USE_WRAPPER "${_bandicoot_CONFIG_CONTENTS}")
+
+  # COOT_USE_OPENCL
+  string(REGEX MATCH "\r?\n[\t ]*#if[\t ]+!defined[(]COOT_USE_OPENCL[)][\t ]*\r?\n[\t
+  ]*#define[ \t]+COOT_USE_OPENCL[ \t]*\r?\n" COOT_USE_OPENCL "${_bandicoot_CONFIG_CONTENTS}")
+
+  # COOT_USE_CUDA
+  string(REGEX MATCH "\r?\n[\t ]*#if[\t ]+!defined[(]COOT_USE_CUDA[)][\t ]*\r?\n[\t
+  ]*#define[ \t]+COOT_USE_CUDA[ \t]*\r?\n" COOT_USE_CUDA "${_bandicoot_CONFIG_CONTENTS}")
+
+  # COOT_USE_LAPACK
+  string(REGEX MATCH "\r?\n[\t ]*#if[\t ]+!defined[(]COOT_USE_LAPACK[)][\t ]*\r?\n[\t ]*#define[ \t]+COOT_USE_LAPACK[ \t]*\r?\n" COOT_USE_LAPACK "${_bandicoot_CONFIG_CONTENTS}")
+
+  # COOT_USE_BLAS
+  string(REGEX MATCH "\r?\n[\t ]*#if[\t ]+!defined[(]COOT_USE_BLAS[)][\t ]*\r?\n[\t ]*#define[ \t]+COOT_USE_BLAS[ \t]*\r?\n" COOT_USE_BLAS "${_bandicoot_CONFIG_CONTENTS}")
+
+  # If we aren't wrapping, things get a little more complex.
+  if("${COOT_USE_WRAPPER}" STREQUAL "")
+    set(COOT_NEED_LIBRARY false)
+    message(STATUS "COOT_USE_WRAPPER is not defined, so all dependencies of "
+                   "Bandicoot must be manually linked.")
+
+    set(HAVE_OPENCL false)
+    set(HAVE_CUDA   false)
+    set(HAVE_LAPACK false)
+    set(HAVE_BLAS   false)
+
+    # Search for OpenCL.
+    if (NOT "${COOT_USE_OPENCL}" STREQUAL "" AND NOT HAVE_OPENCL)
+      set(OpenCL_FIND_QUIETLY true)
+      include(FindOpenCL)
+
+      if (OpenCL_FOUND)
+        message(STATUS "OpenCL includes: ${OpenCL_INCLUDE_DIRS}")
+        message(STATUS "OpenCL libraries: ${OpenCL_LIBRARIES}")
+
+        set(SUPPORT_INCLUDE_DIRS "${SUPPORT_INCLUDE_DIRS}"
+            "${OpenCL_INCLUDE_DIRS}")
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${OpenCL_LIBRARIES}")
+        set(HAVE_OPENCL true)
+      endif ()
+
+      # Search for clBLAS.
+      set(CLBLAS_FIND_QUIETLY true)
+      include(COOT_FindCLBLAS)
+
+      if (CLBLAS_FOUND)
+        message(STATUS "clBLAS includes: ${CLBLAS_INCLUDE_DIR}")
+        message(STATUS "clBLAS libraries: ${CLBLAS_LIBRARIES}")
+
+        set(SUPPORT_INCLUDE_DIRS "${SUPPORT_INCLUDE_DIRS}"
+            "${CLBLAS_INCLUDE_DIR}")
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${CLBLAS_LIBRARIES}")
+        set(HAVE_CLBLAS true)
+      endif ()
+
+      # Search for clBlast.
+      set(CLBLAST_FIND_QUIETLY true)
+      include(COOT_FindCLBlast)
+
+      if (CLBLAST_FOUND)
+        message(STATUS "clBlast includes: ${CLBLAST_INCLUDE_DIR}")
+        message(STATUS "clBlast libraries: ${CLBLAST_LIBRARIES}")
+
+        set(SUPPORT_INCLUDE_DIRS "${SUPPORT_INCLUDE_DIRS}"
+            "${CLBLAST_INCLUDE_DIR}")
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${CLBLAST_LIBRARIES}")
+        set(HAVE_CLBLAST true)
+      endif ()
+    endif ()
+
+    # Search for CUDA.
+    if (NOT "${COOT_USE_CUDA}" STREQUAL "" AND NOT HAVE_CUDA)
+      # FindCUDA is deprecated since version 3.10 and replaced with
+      # FindCUDAToolkit wich was added in CMake 3.17.
+      message(STATUS "${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION}")
+      if ("${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION}" LESS "3.67")
+        set(CUDA_FIND_QUIETLY true)
+        find_package(CUDA)
+
+        if (CUDA_FOUND)
+	        message(STATUS "CUDA includes: ${CUDA_INCLUDE_DIRS}")
+	        message(STATUS "CUDA libraries: ${CUDA_LIBRARIES}")
+
+          # We also need NVRTC and also libcuda itself, which the old FindCUDA package do not find.
+          find_library(CUDA_cuda_LIBRARY cuda
+              HINTS ${CUDA_TOOLKIT_ROOT_DIR} ${CUDA_TOOLKIT_ROOT_DIR}/lib ${CUDA_TOOLKIT_ROOT_DIR}/lib64)
+          find_library(CUDA_nvrtc_LIBRARY nvrtc
+              HINTS ${CUDA_TOOLKIT_ROOT_DIR} ${CUDA_TOOLKIT_ROOT_DIR}/lib ${CUDA_TOOLKIT_ROOT_DIR}/lib64)
+
+          include(COOT_FindNVRTC)
+
+          if (NVRTC_FOUND)
+            message(STATUS "NVRTC libraries: ${NVRTC_LIBRARIES}")
+            set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${NVRTC_LIBRARIES}")
+          endif ()
+
+          set(SUPPORT_INCLUDE_DIRS "${SUPPORT_INCLUDE_DIRS}"
+              "${CUDA_INCLUDE_DIRS}")
+          set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}"
+              "${CUDA_LIBRARIES}"
+              "${CUDA_nvrtc_LIBRARY}"
+              "${CUDA_CUDA_LIBRARY}"
+              "${CUDA_CUBLAS_LIBRARIES}"
+              "${CUDA_curand_LIBRARY}"
+              "${CUDA_cusolver_LIBRARY}")
+          set(CUDA_INCLUDE_DIRS "")
+          set(HAVE_CUDA true)
+
+        endif ()
+      else ()
+        set(CUDA_TOOLKIT_FIND_QUIETLY true)
+        find_package(CUDAToolkit REQUIRED)
+
+        if (CUDAToolkit_FOUND)
+	        message(STATUS "CUDA includes: ${CUDAToolkit_INCLUDE_DIRS}")
+	        message(STATUS "CUDA libraries: ${CUDAToolkit_LIBRARY_DIR}")
+
+	  set(CUDA_LIBRARIES CUDA::cudart CUDA::cuda_driver)
+          set(CUDA_CUBLAS_LIBRARIES CUDA::cublas)
+          set(CUDA_curand_LIBRARY CUDA::curand)
+          set(CUDA_cusolver_LIBRARY CUDA::cusolver)
+          set(CUDA_nvrtc_LIBRARY CUDA::nvrtc)
+
+          set(SUPPORT_INCLUDE_DIRS "${SUPPORT_INCLUDE_DIRS}"
+              "${CUDAToolkit_INCLUDE_DIRS}")
+          set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}"
+              CUDA_LIBRARIES
+	      CUDA_CUBLAS_LIBRARIES
+	      CUDA_curand_LIBRARY
+	      CUDA_cusolver_LIBRARY
+	      CUDA_nvrtc_LIBRARY)
+          set(HAVE_CUDA true)
+        endif()
+      endif ()
+    endif ()
+
+    # Search for LAPACK/BLAS (or replacement).
+    if ((NOT "${COOT_USE_LAPACK}" STREQUAL "") AND
+      (NOT "${COOT_USE_BLAS}" STREQUAL ""))
+      # In order of preference: MKL, ACML, OpenBLAS, ATLAS
+      set(MKL_FIND_QUIETLY true)
+      include(ARMA_FindMKL)
+      set(ACMLMP_FIND_QUIETLY true)
+      include(ARMA_FindACMLMP)
+      set(ACML_FIND_QUIETLY true)
+      include(ARMA_FindACML)
+
+      if (MKL_FOUND)
+        message(STATUS "Using MKL for LAPACK/BLAS: ${MKL_LIBRARIES}")
+
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${MKL_LIBRARIES}")
+        set(HAVE_LAPACK true)
+        set(HAVE_BLAS   true)
+      elseif (ACMLMP_FOUND)
+        message(STATUS "Using multi-core ACML libraries for LAPACK/BLAS:
+            ${ACMLMP_LIBRARIES}")
+
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${ACMLMP_LIBRARIES}")
+        set(HAVE_LAPACK true)
+        set(HAVE_BLAS   true)
+      elseif (ACML_FOUND)
+        message(STATUS "Using ACML for LAPACK/BLAS: ${ACML_LIBRARIES}")
+
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${ACML_LIBRARIES}")
+        set(HAVE_LAPACK true)
+        set(HAVE_BLAS   true)
+      endif ()
+    endif ()
+
+    # If we haven't found BLAS, try.
+    if (NOT "${COOT_USE_BLAS}" STREQUAL "" AND NOT HAVE_BLAS)
+      # Search for BLAS.
+      set(OpenBLAS_FIND_QUIETLY false)
+      include(ARMA_FindOpenBLAS)
+      set(CBLAS_FIND_QUIETLY true)
+      include(ARMA_FindCBLAS)
+      set(BLAS_FIND_QUIETLY true)
+      include(ARMA_FindBLAS)
+
+      if (OpenBLAS_FOUND)
+        # Warn if ATLAS is found also.
+        if (CBLAS_FOUND)
+          message(STATUS "Warning: both OpenBLAS and ATLAS have been found; "
+              "ATLAS will not be used.")
+        endif ()
+        message(STATUS "Using OpenBLAS for BLAS: ${OpenBLAS_LIBRARIES}")
+
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${OpenBLAS_LIBRARIES}")
+        set(HAVE_BLAS true)
+      elseif (CBLAS_FOUND)
+        message(STATUS "Using ATLAS for BLAS: ${CBLAS_LIBRARIES}")
+
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${CBLAS_LIBRARIES}")
+        set(SUPPORT_INCLUDE_DIRS "${SUPPORT_INCLUDE_DIRS}"
+            "${CBLAS_INCLUDE_DIR}")
+        set(HAVE_BLAS true)
+      elseif (BLAS_FOUND)
+        message(STATUS "Using standard BLAS: ${BLAS_LIBRARIES}")
+
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${BLAS_LIBRARIES}")
+        set(HAVE_BLAS true)
+      endif ()
+    endif ()
+
+    # If we haven't found LAPACK, try.
+    if (NOT "${COOT_USE_LAPACK}" STREQUAL "" AND NOT HAVE_LAPACK)
+      # Search for LAPACK.
+      set(CLAPACK_FIND_QUIETLY true)
+      include(ARMA_FindCLAPACK)
+      set(LAPACK_FIND_QUIETLY true)
+      include(ARMA_FindLAPACK)
+
+      # Only use ATLAS if OpenBLAS isn't being used.
+      if (CLAPACK_FOUND AND NOT OpenBLAS_FOUND)
+        message(STATUS "Using ATLAS for LAPACK: ${CLAPACK_LIBRARIES}")
+
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${CLAPACK_LIBRARIES}")
+        set(SUPPORT_INCLUDE_DIRS "${SUPPORT_INCLUDE_DIRS}"
+            "${CLAPACK_INCLUDE_DIR}")
+        set(HAVE_LAPACK true)
+      elseif (LAPACK_FOUND)
+        message(STATUS "Using standard LAPACK: ${LAPACK_LIBRARIES}")
+
+        set(SUPPORT_LIBRARIES "${SUPPORT_LIBRARIES}" "${LAPACK_LIBRARIES}")
+        set(HAVE_LAPACK true)
+      endif ()
+    endif ()
+
+    if (NOT "${COOT_USE_LAPACK}" STREQUAL "" AND NOT HAVE_LAPACK)
+      message(FATAL_ERROR "Cannot find LAPACK library, but COOT_USE_LAPACK is "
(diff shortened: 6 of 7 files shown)


zoq:
Adding the GPU Jenkins job next, before pushing further changes, so I can make sure, that each commit builds.

include/ensmallen.hpp
```diff
@@ -34,6 +34,10 @@
 
 #include <armadillo>
 
+#ifdef USE_COOT
```
conradsnicta, line 37: @zoq Suggest changing `USE_COOT` to `ENS_USE_COOT` here and in CMakeLists.txt.

This is for consistency with all the other macros used by ensmallen, where the `ENS_` prefix is used.  The prefix helps avoid potential clashes.

zoq: Makes sense, will change it.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -238,28 +259,30 @@ Optimize(std::tuple<ArbitraryFunctionType...>& objectives,
       Mutate(candidate, 1.0 / static_cast<double>(numVariables),
           castedLowerBound, castedUpperBound);
 
-      arma::Col<ElemType> candidateFitness(numObjectives);
+          ColType candidateFitness(numObjectives);
```
rcurtin, line 262:
```suggestion
      ColType candidateFitness(numObjectives);
```

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -238,28 +259,30 @@ Optimize(std::tuple<ArbitraryFunctionType...>& objectives,
...
+      idealPoint = min(idealPoint, candidateFitness);
 
       // 2.5 Update of the population.
       size_t replaceCounter = 0;
       const size_t sampleSize = sampleNeighbor ? neighborSize : populationSize;
 
-      const arma::uvec idxShuffle = arma::shuffle(
-          arma::linspace<arma::uvec>(0, sampleSize - 1, sampleSize));
+      const UVecType idxShuffle = shuffle(
+          linspace<UVecType>(0, sampleSize - 1, sampleSize));
```
rcurtin, line 280: I think this might make more sense to always keep as an `arma::uvec` since we are always iterating over the individual elements on the CPU.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -283,8 +308,9 @@ Optimize(std::tuple<ArbitraryFunctionType...>& objectives,
     } // End of pass over all subproblems.
 
     //  The final population itself is the best front.
-    const std::vector<arma::uvec> frontIndices { arma::shuffle(
-        arma::linspace<arma::uvec>(0, populationSize - 1, populationSize)) };
+    // const std::vector<UVecType> frontIndices { shuffle(
+    //     linspace<UVecType>(0, populationSize - 1, populationSize)) };
+    const std::vector<UVecType> frontIndices;
```
rcurtin, line 313: Is `frontIndices` only used by the `GenerationalStepTaken()` callback?  If so, and if the intention is just that that is the set of the population that is in the front, maybe it makes more sense to keep that on the CPU too.  But I am not deeply familiar with MOEAD.

include/ensmallen_bits/nsga2/nsga2.hpp
```diff
@@ -51,7 +51,12 @@ namespace ens {
  * see the documentation on function types included with this distribution or
  * on the ensmallen website.
  */
-class NSGA2
+template<
+  typename MatType,
+  typename ColType = typename ForwardType<MatType>::bvec,
+  typename CubeType = typename ForwardType<MatType>::bcube
+>
+class NSGA2Type
```
rcurtin, line 59: We should also update the NSGA2 documentation to indicate what's necessary to use NSGA2 with non-`arma::mat` types.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -56,17 +60,19 @@ inline NSGA2::NSGA2(const size_t populationSize,
     mutationProb(mutationProb),
     mutationStrength(mutationStrength),
     epsilon(epsilon),
-    lowerBound(lowerBound * arma::ones(1, 1)),
-    upperBound(upperBound * arma::ones(1, 1))
+    lowerBound(lowerBound * ColType(1, GetFillType<MatType>::ones)),
+    upperBound(upperBound * ColType(1, GetFillType<MatType>::ones))
```
rcurtin, line 64: Couldn't you also do `lowerBound({ lowerBound })`?  I think that should work for both Armadillo and Bandicoot.

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -87,13 +93,23 @@ typename MatType::elem_type NSGA2::Optimize(
...
+        iterate.n_rows, iterate.n_cols, GetFillType<MatType>::ones);
+  }
+  else
+  {
+    castedLowerBound = conv_to<BaseMatType>::from(lowerBound);
+    castedUpperBound = conv_to<BaseMatType>::from(upperBound);
+  }
 
   // Check the dimensions of lowerBound and upperBound.
   assert(lowerBound.n_rows == iterate.n_rows && "The dimensions of "
```
rcurtin, line 115: Do these checks now need to concern `castedLowerBound` since we are not updating `lowerBound` anymore?  (Should we still update `lowerBound` and `upperBound` as before?)

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -322,24 +355,29 @@ inline void NSGA2::Crossover(MatType& childA,
...
 //! Perform mutation of the candidates weights with some noise.
-template<typename MatType>
-inline void NSGA2::Mutate(MatType& child,
-                          const MatType& lowerBound,
-                          const MatType& upperBound)
+template<typename MatType, typename ColType, typename CubeType>
+template<typename InputMatType>
+void NSGA2Type<MatType, ColType, CubeType>::Mutate(InputMatType& child,
+                                const InputMatType& lowerBound,
+                                const InputMatType& upperBound)
```
rcurtin, line 362:
```suggestion
void NSGA2Type<MatType, ColType, CubeType>::Mutate(
    InputMatType& child,
    const InputMatType& lowerBound,
    const InputMatType& upperBound)
```

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -409,47 +448,60 @@ inline bool NSGA2::Dominates(
...
         {
-          return (fValues(frontIdxA) < fValues(frontIdxB));
-        });
+          minIdx = j;
+        }
+      }
+      size_t temp = sortedIdx[i];
+      sortedIdx[i] = sortedIdx[minIdx];
+      sortedIdx[minIdx] = temp;
+    }
```
rcurtin, line 504: This sorting (the entire block above) is done on the CPU with `std::sort`; but couldn't we use `sort_index()` here instead to achieve the same result?

include/ensmallen_bits/nsga2/nsga2_impl.hpp
```diff
@@ -409,47 +448,60 @@ inline bool NSGA2::Dominates(
...
-      [&](const size_t& individual)
-        {
-          return calculatedObjectives[individual](m);
-        });
+    ColType fValues(fSize);
+    for (size_t i = 0; i < front.size(); ++i)
+    {
+      size_t individual = front[i];
+      fValues[i] = calculatedObjectives[individual](m);
+    }
```
rcurtin, line 488: If `calculatedObjectives` was represented as a cube (I am not sure if it can be or how easy that refactoring would be), then this entire operation here could just be a subcube view, which would be way more efficient for Bandicoot and I think also faster for Armadillo too.

include/ensmallen_bits/parallel_sgd/parallel_sgd_impl.hpp
```diff
@@ -92,7 +96,7 @@ typename MatType::elem_type>::type ParallelSGD<DecayPolicyType>::Optimize(
 
   // The order in which the functions will be visited.
   // TODO: maybe use function.Shuffle() instead?
-  arma::Col<size_t> visitationOrder = arma::linspace<arma::Col<size_t>>(0,
+  UVecType visitationOrder = linspace<UVecType>(0,
```
rcurtin, line 99: Since the visitation order is entirely used on the CPU (e.g. each thread gets the next element of the visitation order and then decides what to do based on that), I don't think there can be any advantage to moving it to the GPU; so we should probably stick with `arma::uvec` here.

include/ensmallen_bits/problems/aug_lagrangian_test_functions.hpp
```diff
@@ -56,40 +56,43 @@ class AugLagrangianTestFunction
...
+  template<typename InputMatType>
+  InputMatType GetInitialPoint() const
   {
-    return arma::conv_to<MatType>::from(initialPoint);
+    return conv_to<InputMatType>::from(initialPoint);
   }
 
  private:
-  arma::mat initialPoint;
+  MatType initialPoint;
```
rcurtin, line 91: Very technically, it could be okay to leave this and other functions as holding an `arma::mat`; since `conv_to` can convert from CPU to GPU matrices, it's totally okay in `GetInitialPoint()` to do the conversion there on the spot, only once.  That could simplify a lot of the changes you've made in `problems/`.

include/ensmallen_bits/problems/logistic_regression_function_impl.hpp
```diff
@@ -64,19 +66,22 @@ LogisticRegressionFunction<MatType>::LogisticRegressionFunction(
...
 
   arma::uvec ordering = arma::shuffle(arma::linspace<arma::uvec>(0,
       predictors.n_cols - 1, predictors.n_cols));
 
   newPredictors.set_size(predictors.n_rows, predictors.n_cols);
+  newResponses.set_size(responses.n_elem);
   for (size_t i = 0; i < predictors.n_cols; ++i)
+  {
     newPredictors.col(i) = predictors.col(ordering[i]);
-  newResponses = responses.cols(ordering);
```
rcurtin, line 79: If you make `ordering` into `LabelsType` (or some equivalent unsigned vector type), then I think `.cols()` should work on the GPU too?

include/ensmallen_bits/problems/logistic_regression_function_impl.hpp
```diff
@@ -98,28 +103,29 @@ typename MatType::elem_type LogisticRegressionFunction<MatType>::Evaluate(
...
   // doesn't actually affect the optimization result, so we'll just ignore those
   // terms for computational efficiency.  Note that the conversion causes some
   // copy and slowdown, but this is so negligible compared to the rest of the
   // calculation it is not worth optimizing for.
-  const ElemType result = arma::accu(arma::log(1.0 -
-      arma::conv_to<arma::Row<ElemType>>::from(responses) + sigmoid %
-      (2 * arma::conv_to<arma::Row<ElemType>>::from(responses) - 1.0)));
+  const ElemType result = accu(log(1.0 -
+      conv_to<BaseRowType>::from(responses) + sigmoid %
+      (2 * conv_to<BaseRowType>::from(responses) - 1.0)));
```
rcurtin, line 128: Just as a side note, unrelated to this PR, this function is way faster if we convert to `BaseRowType` in the constructor, just once, and then use the converted version in the `Evaluate()`/`Gradient()` implementations.

include/ensmallen_bits/problems/logistic_regression_function_impl.hpp
```diff
@@ -129,97 +135,100 @@ typename MatType::elem_type LogisticRegressionFunction<MatType>::Evaluate(
...
-void LogisticRegressionFunction<MatType>::PartialGradient(
+template <typename MatType, typename LabelsType>
+template<typename ElemType>
+void LogisticRegressionFunction<MatType, LabelsType>::PartialGradient(
     const MatType& parameters,
     const size_t j,
-    arma::sp_mat& gradient) const
+    arma::SpMat<ElemType>& gradient) const
 {
   const arma::Row<typename MatType::elem_type> diffs = responses -
```
rcurtin, line 233: I guess you can't use the implementation below because the result here has to be dense?

include/ensmallen_bits/problems/logistic_regression_function_impl.hpp
```diff
@@ -239,52 +248,80 @@ void LogisticRegressionFunction<MatType>::PartialGradient(
...
+  gradient.set_size(size(parameters));
+
+  if (j == 0)
+  {
+    gradient[j] = -accu(diffs);
+  }
+  else
+  {
+    gradient[j] = dot(-predictors.row(j - 1), diffs) + lambda *
+      parameters(0, j);
```
rcurtin, line 273:
```suggestion
        parameters(0, j);
```

include/ensmallen_bits/problems/problems.hpp
```diff
@@ -54,11 +54,5 @@
 #include "dtlz/dtlz5_function.hpp"
 #include "dtlz/dtlz6_function.hpp"
 #include "dtlz/dtlz7_function.hpp"
-#include "maf/maf1_function.hpp"
-#include "maf/maf2_function.hpp"
-#include "maf/maf3_function.hpp"
-#include "maf/maf4_function.hpp"
-#include "maf/maf5_function.hpp"
-#include "maf/maf6_function.hpp"
```
rcurtin, line 62: Any particular reason to remove these?

include/ensmallen_bits/problems/zdt/zdt3_function.hpp
```diff
@@ -78,13 +81,14 @@ namespace test {
...
+    ColType Evaluate(const MatType& coords)
     {
       typedef typename MatType::elem_type ElemType;
 
-      arma::Col<ElemType> objectives(numObjectives);
+      ColType objectives(numObjectives);
       objectives(0) = coords[0];
-      ElemType sum = arma::accu(coords(arma::span(1, numVariables - 1), 0));
+      // ElemType sum = accu(coords(span(1, numVariables - 1), 0));
+      ElemType sum = 0;
```
rcurtin, line 91: I think this needs to be reverted to its initial implementation.

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -60,12 +60,14 @@ namespace ens {
...
  *    double Evaluate(const arma::mat& x);
  *
+ * @tparam MatType Type of matrix to optimize.
  * @tparam VelocityUpdatePolicy Velocity update policy. By default LBest update
  *     policy (see ens::LBestUpdate) is used.
  * @tparam InitPolicy Particle initialization policy. By default DefaultInit
  *     policy (see ens::DefaultInit) is used.
  */
-template<typename VelocityUpdatePolicy = LBestUpdate,
+template<typename MatType = arma::mat,
```
rcurtin, line 69: If we need to add the matrix type as a template parameter, we should add it last for reverse compatibility, and update `doc/function_types.md` accordingly for PSO.

include/ensmallen_bits/pso/pso.hpp
```diff
@@ -145,8 +149,8 @@ class PSOType
               VelocityUpdatePolicy(),
           const InitPolicy& initPolicy = InitPolicy()) :
           numParticles(numParticles),
-          lowerBound(lowerBound * arma::ones(1, 1)),
-          upperBound(upperBound * arma::ones(1, 1)),
+          lowerBound(lowerBound * MatType(1, 1, GetFillType<MatType>::ones)),
+          upperBound(upperBound * MatType(1, 1, GetFillType<MatType>::ones)),
```
rcurtin, line 153: I saw it somewhere else too and commented there also, but I think `lowerBound({ lowerBound })` will work here.

include/ensmallen_bits/sa/sa_impl.hpp
```diff
@@ -76,8 +76,9 @@ typename MatType::elem_type SA<CoolingScheduleType>::Optimize(
   size_t idx = 0;
   size_t sweepCounter = 0;
 
-  BaseMatType accept(rows, cols, arma::fill::zeros);
-  BaseMatType moveSize(rows, cols, arma::fill::none);
+  BaseMatType accept(rows, cols);
+  accept.zeros();
```
rcurtin, line 80: Technically this line is not necessary since it's already the default.

include/ensmallen_bits/sa/sa_impl.hpp
```diff
@@ -76,8 +76,9 @@ typename MatType::elem_type SA<CoolingScheduleType>::Optimize(
...
+  BaseMatType moveSize(rows, cols);
```
rcurtin, line 81:
```suggestion
  BaseMatType moveSize(rows, cols, GetFillType<BaseMatType>::none);
```
I think this is correct?

include/ensmallen_bits/sa/sa_impl.hpp
```diff
@@ -219,12 +220,12 @@ inline void SA<CoolingScheduleType>::MoveControl(const size_t nMoves,
                                                  MatType& accept,
                                                  MatType& moveSize)
 {
-  MatType target;
-  target.copy_size(accept);
+  MatType target(accept.n_rows, accept.n_cols);
```
rcurtin, line 223:
```suggestion
  MatType target(accept.n_rows, accept.n_cols, GetFillType<MatType>::none);
```
A small optimization.

include/ensmallen_bits/sa/sa_impl.hpp
```diff
@@ -219,12 +220,12 @@ inline void SA<CoolingScheduleType>::MoveControl(const size_t nMoves,
...
   target.fill(0.44);
-  moveSize = arma::log(moveSize);
+
+  moveSize = log(moveSize);
   moveSize += gain * (accept / (double) nMoves - target);
-  moveSize = arma::exp(moveSize);
+  moveSize = exp(moveSize);
 
   // To avoid the use of element-wise arma::min(), which is only available in
   // Armadillo after v3.930, we use a for loop here instead.
```
rcurtin, line 231: Wow, this is ancient!  You could replace this part now too, and it'll be a lot faster on the GPU.

include/ensmallen_bits/smorms3/smorms3_update.hpp
```diff
@@ -94,26 +100,24 @@ class SMORMS3Update
       g2 = (1 - r) % g2;
       g2 += r % (gradient % gradient);
 
-      MatType x = (g % g) / (g2 + parent.epsilon);
-
-      x.transform( [stepSize](typename MatType::elem_type &v)
-          { return std::min(v, (typename MatType::elem_type) stepSize); } );
-
-      iterate -= gradient % x / (arma::sqrt(g2) + parent.epsilon);
+      MatType x = min((g % g) / (g2 + parent.epsilon), lr);
```
rcurtin, line 103: I think that using `clamp()` would actually be easier here, I guess `x = clamp(..., 0, lr)`.  (I think `x` is always nonnegative... but I might be wrong on that, I'm not sure.  Double check me on that one...)

include/ensmallen_bits/spalera_sgd/spalera_sgd_impl.hpp
```diff
@@ -212,13 +213,10 @@ SPALeRASGD<DecayPolicyType>::Optimize(
...
-      if (i != actualMaxIterations)
-      {
-        lastObjective = overallObjective;
-        overallObjective = 0;
-        currentFunction = 0;
-      }
+      // Reset the counter variables.
+      lastObjective = overallObjective;
+      overallObjective = 0;
+      currentFunction = 0;
```
rcurtin, line 219: Same comment, not sure the `if` should be removed.

include/ensmallen_bits/spalera_sgd/spalera_stepsize.hpp
```diff
@@ -172,7 +172,9 @@ class SPALeRAStepsize
         // Faster.
         learningRates /= 2;
 
-        if (arma::any(arma::vectorise(learningRates) <= 1e-15))
+        //if (any(vectorise(learningRates) <= 1e-15))
+        /* if (min(vectorise(learningRates)) <= 1e-15) */
+        if (learningRates.min() <= 1e-15)
```
rcurtin, line 177:
```suggestion
        if (learningRates.min() <= 1e-15)
```
I think this is an improvement overall anyway, nice.

tests/ada_belief_test.cpp
```diff
@@ -8,64 +8,30 @@
...
-#include "catch.hpp"
-#include "test_function_tools.hpp"
+ #include <ensmallen.hpp>
+ #include "catch.hpp"
+ #include "test_function_tools.hpp"
 
-using namespace ens;
-using namespace ens::test;
+ using namespace ens;
+ using namespace ens::test;
```
rcurtin, line 16: No need to add the spaces here?

tests/ada_bound_test.cpp
```diff
@@ -8,103 +8,92 @@
...
-#include "catch.hpp"
-#include "test_function_tools.hpp"
+ #include <ensmallen.hpp>
+ #include "catch.hpp"
+ #include "test_function_tools.hpp"
 
-using namespace ens;
-using namespace ens::test;
+ using namespace ens;
+ using namespace ens::test;
```
rcurtin, line 16: Same here, I don't think the spaces are needed.

tests/adam_test.cpp
```diff
@@ -13,319 +13,279 @@
...
+   Adam optimizer(1e-1, 1, 0.7, 0.999, 1e-8, 50000, 1e-9, true);
+   FunctionTest<BoothFunction, TestType>(optimizer);
+ }
+
+ TEMPLATE_TEST_CASE("AMSGrad_SphereFunction", "[Adam]", arma::fmat)
+ {
+   AMSGrad optimizer(1e-3, 1, 0.9, 0.999, 1e-8, 50000, 1e-11, true);
+   FunctionTest<SphereFunctionType<TestType, arma::Row<size_t>>, TestType>(
+       optimizer, 0.5, 0.1);
+ }
```
rcurtin, line 66: Should this and some of the other optimizer tests be in this file, or in another file?  I think they were originally in other files.

tests/moead_test.cpp
```diff
@@ -776,36 +654,16 @@ TEST_CASE("MOEADDIRICHLETMAF4Test", "[MOEADTest]")
...
-  }
 
-  REQUIRE(success == true);
-}
\ No newline at end of file
+  const coot::Cube<ElemType>& finalPopulation = opt.ParetoSet();
+  REQUIRE(VariableBoundsCheck(finalPopulation));
+}
+
+#endif
```
rcurtin, line 669: Correct me if I'm wrong, but it looks like a few tests were removed for MOEAD---should we add them back?

tests/spsa_test.cpp
```diff
@@ -11,57 +11,70 @@
...
-/**
- * Test the SPSA optimizer on the Matyas function.
- */
-TEST_CASE("SPSAMatyasFunctionTest", "[SPSATest]")
-{
-  SPSA optimizer(0.1, 0.102, 0.16, 0.3, 100000, 0);
-  FunctionTest<MatyasFunction>(optimizer, 0.1, 0.01);
-}
+ #if ARMA_VERSION_MAJOR > 9 ||\
+     (ARMA_VERSION_MAJOR == 9 && ARMA_VERSION_MINOR >= 400)
```
rcurtin, line 45: This could be removed now.

rcurtin (commented):
Hopefully I won't break Github with all of these comments :grimacing: :grimacing: 

Anyway this is great.  I left a lot of comments but hopefully most of them are trivial.  I tried to identify places where it would be better to keep things on the CPU instead of potentially moving them to the GPU.  I'm not certain I identified every place but I did my best.  Many of the test files had some strange indentation so it made it hard to review that part---can you see maybe what happened there?

It looks like some of the Jenkins CI nodes (hinode or polar) are down so some of the CI builds don't run.  If you want to move those to GHA instead of Jenkins that works too, up to you.

include/ensmallen_bits/de/de_impl.hpp
```diff
@@ -100,13 +102,13 @@ typename MatType::elem_type DE::Optimize(FunctionType& function,
       size_t l = 0, m = 0;
       do
       {
-        l = arma::randi<arma::uword>(arma::distr_param(0, populationSize - 1));
+        l = rand() % populationSize;
```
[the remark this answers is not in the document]
zoq: I can replace this with a`rma::randi` as well, and just generate a scalar, I think.

include/ensmallen_bits/utility/proxies.hpp
```diff
@@ -0,0 +1,241 @@
...
+}
+
+/**
+ * A utility class that based on the data type forwards to `coot::conv_to` or
+ * `arma::conv_to`.
+ *
+ * @tparam OutputType The data type to convert to.
+ */
+template<typename OutputType>
+class conv_to
```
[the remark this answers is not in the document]
zoq: Yes, removed.

CMakeLists.txt
```diff
@@ -59,6 +60,17 @@ if(USE_OPENMP)
   target_link_libraries(ensmallen INTERFACE OpenMP::OpenMP_CXX)
 endif()
 
+if(USE_BANDICOOT)
+  # Find Bandicoot and link it.
+  find_package(Bandicoot 0.100.0)
```
rcurtin, line 65:
```suggestion
  find_package(Bandicoot 2.1.0)
```

I think this is the actual minimum version we require?

include/ensmallen_bits/gradient_descent/gradient_descent.hpp
```diff
@@ -120,27 +120,29 @@ class GradientDescent
...
            typename GradType,
+           typename CategoryType,
            typename... CallbackTypes>
   typename std::enable_if<IsArmaType<GradType>::value,
       typename MatType::elem_type>::type
   Optimize(FunctionType& function,
            MatType& iterate,
            const std::vector<bool>& categoricalDimensions,
-           const arma::Row<size_t>& numCategories,
+           const CategoryType& numCategories,
```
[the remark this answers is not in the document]
zoq: Will change it back.

tests/cmaes_test.cpp
```diff
@@ -18,60 +18,87 @@ using namespace ens;
...
- * Run CMA-ES with the random selection and empty transformation policies
- * on logistic regression and make sure the results are acceptable.
- * Use arma::fmat.
- */
-TEST_CASE("ApproxCMAESEmptyTransformationLogisticRegressionFMatTest",
-  "[CMAESTest]")
+* Run CMA-ES with the random selection and empty transformation policies
+* on logistic regression and make sure the results are acceptable.
+* Use arma::fmat.
+*/
```
rcurtin, line 74:
```suggestion
 * Run CMA-ES with the random selection and empty transformation policies
 * on logistic regression and make sure the results are acceptable.
 * Use arma::fmat.
 */
```

tests/cmaes_test.cpp
```diff
@@ -18,60 +18,87 @@ using namespace ens;
...
 }
 
 /**
- * Run CMA-ES with the random selection policy on logistic regression and
- * make sure the results are acceptable.  Use arma::fmat.
- */
-TEST_CASE("ApproxCMAESLogisticRegressionFMatTest", "[CMAESTest]")
+* Run CMA-ES with the random selection policy on logistic regression and
+* make sure the results are acceptable.  Use arma::fmat.
+*/
```
rcurtin, line 61:
```suggestion
 * Run CMA-ES with the random selection policy on logistic regression and
 * make sure the results are acceptable.  Use arma::fmat.
 */
```

tests/cmaes_test.cpp
```diff
@@ -18,60 +18,87 @@ using namespace ens;
...
 }
 
 /**
- * Run CMA-ES with the full selection policy on logistic regression and
- * make sure the results are acceptable.  Use arma::fmat.
- */
-TEST_CASE("CMAESLogisticRegressionFMatTest", "[CMAESTest]")
+* Run CMA-ES with the full selection policy on logistic regression and
+* make sure the results are acceptable.  Use arma::fmat.
+*/
```
rcurtin, line 49:
```suggestion
 * Run CMA-ES with the full selection policy on logistic regression and
 * make sure the results are acceptable.  Use arma::fmat.
 */
```

tests/cmaes_test.cpp
```diff
@@ -18,60 +18,87 @@ using namespace ens;
...
 }
 
 /**
- * Run CMA-ES with the random selection policy on logistic regression and
- * make sure the results are acceptable.
- */
-TEST_CASE("ApproxCMAESLogisticRegressionTest", "[CMAESTest]")
+* Run CMA-ES with the random selection policy on logistic regression and
+* make sure the results are acceptable.
+*/
```
rcurtin, line 36:
```suggestion
 * Run CMA-ES with the random selection policy on logistic regression and
 * make sure the results are acceptable.
 */
```

tests/cmaes_test.cpp
```diff
@@ -18,60 +18,87 @@ using namespace ens;
 using namespace ens::test;
 
 /**
- * Run CMA-ES with the full selection policy on logistic regression and
- * make sure the results are acceptable.
- */
-TEST_CASE("CMAESLogisticRegressionTest", "[CMAESTest]")
+* Run CMA-ES with the full selection policy on logistic regression and
+* make sure the results are acceptable.
+*/
```
rcurtin, line 23:
```suggestion
 * Run CMA-ES with the full selection policy on logistic regression and
 * make sure the results are acceptable.
 */
```

tests/aug_lagrangian_test.cpp
```diff
@@ -20,16 +20,19 @@ using namespace ens;
 using namespace ens::test;
 
 /**
- * Tests the Augmented Lagrangian optimizer using the
- * AugmentedLagrangianTestFunction class.
- */
-TEST_CASE("AugLagrangianTestFunctionTest", "[AugLagrangianTest]")
+* Tests the Augmented Lagrangian optimizer using the
+* AugmentedLagrangianTestFunction class.
+*/
```
rcurtin, line 25:
```suggestion
 * Tests the Augmented Lagrangian optimizer using the
 * AugmentedLagrangianTestFunction class.
 */
```

tests/aug_lagrangian_test.cpp
```diff
@@ -42,69 +45,79 @@ TEST_CASE("AugLagrangianTestFunctionTest", "[AugLagrangianTest]")
 }
 
 /**
- * Tests the Augmented Lagrangian optimizer using the Gockenbach function.
- */
-TEST_CASE("GockenbachFunctionTest", "[AugLagrangianTest]")
+* Tests the Augmented Lagrangian optimizer using the Gockenbach function.
+*/
```
rcurtin, line 49:
```suggestion
 * Tests the Augmented Lagrangian optimizer using the Gockenbach function.
 */
```

tests/aug_lagrangian_test.cpp
```diff
@@ -42,69 +45,79 @@ TEST_CASE("AugLagrangianTestFunctionTest", "[AugLagrangianTest]")
...
 
 /**
- * Tests the Augmented Lagrangian optimizer using the Gockenbach function.  Uses
- * arma::fmat.
- */
-TEST_CASE("GockenbachFunctionFMatTest", "[AugLagrangianTest]")
+* Tests the Augmented Lagrangian optimizer using the Gockenbach function.  Uses
+* arma::sp_mat.
+*/
+TEMPLATE_TEST_CASE("AugLagrangian_GockenbachFunction", "[AugLagrangian]",
```
rcurtin, line 76:
```suggestion
 * Tests the Augmented Lagrangian optimizer using the Gockenbach function.  Uses
 * arma::sp_mat.
 */
TEMPLATE_TEST_CASE("AugLagrangian_GockenbachFunction", "[AugLagrangian]",
```

tests/ada_grad_test.cpp
```diff
@@ -10,27 +10,29 @@
  * http://www.opensource.org/licenses/BSD-3-Clause for more information.
  */
 
-#include <ensmallen.hpp>
-#include "catch.hpp"
-#include "test_function_tools.hpp"
+ #include <ensmallen.hpp>
+ #include "catch.hpp"
+ #include "test_function_tools.hpp"
```
rcurtin, line 15:
```suggestion
#include <ensmallen.hpp>
#include "catch.hpp"
#include "test_function_tools.hpp"
```

tests/ada_grad_test.cpp
```diff
@@ -10,27 +10,29 @@
...
 
-using namespace ens;
-using namespace ens::test;
+ using namespace ens;
+ using namespace ens::test;
```
rcurtin, line 18:
```suggestion
using namespace ens;
using namespace ens::test;
```

tests/ada_grad_test.cpp
```diff
@@ -10,27 +10,29 @@
...
-  AdaGrad adagrad(0.99, 32, 1e-8, 5000000, 1e-9, true);
-  LogisticRegressionFunctionTest(adagrad, 0.003, 0.006);
-}
+ TEMPLATE_TEST_CASE("AdaGrad_LogisticRegressionFunction", "[AdaGrad]",
+     arma::mat, arma::fmat)
+ {
+   AdaGrad adagrad(0.99, 32, 1e-8, 5000000, 1e-9, true);
+   LogisticRegressionFunctionTest<TestType, arma::Row<size_t>>(
+       adagrad, 0.003, 0.006);
+ }
```
rcurtin, line 26:
```suggestion
TEMPLATE_TEST_CASE("AdaGrad_LogisticRegressionFunction", "[AdaGrad]",
    arma::mat, arma::fmat)
{
  AdaGrad adagrad(0.99, 32, 1e-8, 5000000, 1e-9, true);
  LogisticRegressionFunctionTest<TestType, arma::Row<size_t>>(
      adagrad, 0.003, 0.006);
}
```

rcurtin (commented):
Thanks for handling a bunch of the comments I left.  I think any new comments are generally just style fix suggestions so hopefully it doesn't add to the workload in any significant way and instead actually makes things easier :)

include/ensmallen_bits/ada_belief/ada_belief.hpp
```diff
@@ -97,8 +97,8 @@ class AdaBelief
            typename MatType,
            typename GradType,
            typename... CallbackTypes>
-  typename std::enable_if<IsArmaType<GradType>::value,
-      typename MatType::elem_type>::type
+  typename std::enable_if<IsArmaType<GradType>::value ||
+      IsCootType<GradType>::value, typename MatType::elem_type>::type
```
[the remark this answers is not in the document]
rcurtin: I'm happy to do this in a follow-up PR if you want, just let me know.

include/ensmallen_bits/eve/eve_impl.hpp
```diff
@@ -186,13 +187,10 @@ Eve::Optimize(SeparableFunctionType& function,
...
-      if (i != actualMaxIterations)
-      {
-        lastOverallObjective = overallObjective;
-        overallObjective = 0;
-        currentFunction = 0;
-      }
+      // Reset the counter variables.
+      lastOverallObjective = overallObjective;
+      overallObjective = 0;
+      currentFunction = 0;
```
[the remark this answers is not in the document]
rcurtin: I checked in other places, and I see this also in `SGD` and `SPALeRASGD`.  I'm not 100% sure that's all, but I think it is.

include/ensmallen_bits/gradient_descent/gradient_descent.hpp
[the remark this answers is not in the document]
rcurtin: Sounds good, also this would apply to the changes to `GridSearch` too.

include/ensmallen_bits/parallel_sgd/parallel_sgd_impl.hpp
rcurtin: I thought about this one further: I actually don't think it makes sense to use `ParallelSGD` with Bandicoot or any non-CPU matrix type.  The premise of ParallelSGD is that you use multiple threads that each update parts of the (sparse) gradient, and the idea is that because each function corresponds (mostly) to different parts of the sparse gradient, that there will not be much contention.  This implementation uses OpenMP atomic locks for update---but that doesn't make much sense for a GPU update, it will be terrible.  If we wanted ParallelSGD on the GPU, likely a fully different implementation (not using OpenMP) would be needed.

So personally I think we should revert the changes to this class.

include/ensmallen_bits/problems/aug_lagrangian_test_functions.hpp
rcurtin: I dug through this a bit further and I think it makes more sense to just leave the functions as-is and not templatize them, and only add a `conv_to` to `GetInitialPoint()` where it is used.  It seems like these classes are the ones affected:

 - GeneralizedRosenbrockFunction (also visitationOrder should stay an arma::Row<size_t> here and in others)
 - LogisticRegressionFunction but don't worry about changing this one back because I had to refactor it differently for fp16
 - RastriginFunction
 - RosenbrockWoodFunction
 - SGDTestFunction
 - SchafferFunctionN1 (we can just return a MatType with one column, no need for a column at all; I dunno, up to you, this one could be a little different)
 - SphereFunction
 - StyblinskiTangFunction
 - ZDT3, although for this one, would anyone ever set ColType and CubeType differently?

tests/ada_delta_test.cpp
```diff
@@ -18,21 +18,22 @@
...
 
-/**
- * Run AdaDelta on logistic regression and make sure the results are acceptable
- * with arma::fmat as the type.
- */
-TEST_CASE("AdaDeltaLogisticRegressionTestFMat", "[AdaDeltaTest]")
+#ifdef USE_COOT
+
+TEMPLATE_TEST_CASE("AdaDeltaLogisticRegressionTest", "[AdaDelta]",
+    coot::mat, coot::fmat)
```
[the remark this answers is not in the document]
rcurtin: In the FP16 work I am doing, I'll handle this issue, and then in a follow-up PR we can merge the list of Armadillo types with the Bandicoot types that are available.  So, no need to do anything with the comment above.

include/ensmallen_bits/utility/proxies.hpp
```diff
@@ -0,0 +1,294 @@
...
+ *
+ * This allows for writing generic code that can work with both `coot` and
+ * `arma` types without needing to know which library is being used at compile
+ * time.
+ */
+template<typename MatType, typename ElemType = typename MatType::elem_type>
+struct ForwardType
+{
+    // If MatType is an Bandicoot type, then we use Bandicoot types.
+    #ifdef USE_COOT
```
rcurtin, line 35: I think it might be simpler here to use SFINAE to specialize the struct for all Bandicoot types at once, instead of using `std::conditional`, but I don't think it's a big deal either way---let me know what you think.  Happy to take a stab at it if you want.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
```diff
@@ -105,8 +107,7 @@ class LBestUpdate
        chi = 2.0 / std::abs(2.0 - phi - std::sqrt((phi - 4.0) * phi));
 
        // Initialize local best indices to self indices of particles.
-       localBestIndices = arma::linspace<
-           arma::Col<typename MatType::elem_type> >(0, n-1, n);
+       localBestIndices = linspace<ColType>(0, n - 1, n);
```
rcurtin, line 110: These all seem to be individually accessed, so I think it is best to keep this on the CPU.  Also, I'm not sure why it was `arma::Col<typename MatType::elem_type>` to begin with---it seems like it would work just fine as `arma::uvec` (in fact that would make more sense to me).

include/ensmallen_bits/ada_belief/ada_belief.hpp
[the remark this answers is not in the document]
zoq: Great idea, I can do that one as part of the PR, should be straightforward.

include/ensmallen_bits/cmaes/cmaes_impl.hpp
```diff
@@ -196,22 +198,24 @@ typename MatType::elem_type CMAES<SelectionPolicyType,
     // Perform Cholesky decomposition. If the matrix is not positive definite,
     // add a small value and try again.
     BaseMatType covLower;
-    while (!arma::chol(covLower, C[idx0], "lower"))
+    // while (!chol(covLower, C[idx0], "lower"))
+    while (!chol(covLower, C[idx0]))
```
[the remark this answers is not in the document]
zoq: Yes, `upper` works here, I'll add a TODO, and once it's implemented on the Bandicoot side, we can switch it back.

include/ensmallen_bits/parallel_sgd/parallel_sgd_impl.hpp
zoq: Makes sense, will revert `ParallelSGD`.

include/ensmallen_bits/utility/proxies.hpp
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// Armadillo proxy for `arma::regspace` if `OutputType` is an Armadillo type.
+template<typename OutputType>
+inline static typename std::enable_if<
+    arma::is_arma_type<OutputType>::value, OutputType>::type
+regspace(const double start, const double end, const size_t num = 100u)
+{
+  return arma::regspace<OutputType>(start, end, num);
+}
```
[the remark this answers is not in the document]
zoq: I don't have the error message, but I tried it and it failed. I'll remove it again and post the error message here.

include/ensmallen_bits/pso/update_policies/lbest_update.hpp
zoq: Okay, will leave it on the CPU.

include/ensmallen_bits/utility/proxies.hpp
zoq: If you can take a look at this one I would appreciated it.

rcurtin: Done in 44b215fa. :+1:

include/ensmallen_bits/parallel_sgd/parallel_sgd.hpp
```diff
@@ -146,7 +147,7 @@ class ParallelSGD
 
   //! Controls whether or not the individual functions are shuffled when
   //! iterating.
-  bool shuffle;
+  bool shuffleFlag;
```
rcurtin, line 150: I noticed that there are specific specializations in mlpack for `ParallelSGD` that use the `shuffle` member; so if we change the name, downstream mlpack will fail.  I don't mind if we want to do an `#ifdef` check on the ensmallen version downstream, but it would be better if we don't need to change this variable's name at all.  Do you think that's possible or is there a reason it needs to become `shuffleFlag`?

rcurtin: Just answered my own question... it conflicts with `shuffle()`.  I know the downstream change to make; will open a PR there.

include/ensmallen_bits/ada_belief/ada_belief.hpp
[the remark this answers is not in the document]
zoq: @rcurtin can you handle this one?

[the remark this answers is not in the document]
rcurtin: Done in c3af24bf. :+1:

include/ensmallen_bits/utility/using.hpp
```diff
@@ -0,0 +1,169 @@
...
+using arma::pow;
+using arma::randn;
+using arma::randi;
+using arma::randu;
+using arma::regspace;
+using arma::repmat;
+using arma::shuffle;
+using arma::sign;
+using arma::size;
+using arma::sort;
```
rcurtin, line 95:
```suggestion
using arma::sort;
using arma::sort_index;
```
I think we need this one too so that ADL doesn't get used.

include/ensmallen_bits/utility/using.hpp
```diff
@@ -0,0 +1,169 @@
...
+using coot::ones;
+using coot::pow;
+using coot::randn;
+using coot::randi;
+using coot::randu;
+using coot::regspace;
+using coot::repmat;
+using coot::shuffle;
+using coot::sign;
+using coot::size;
```
rcurtin, line 52:
```suggestion
using coot::size;
using coot::sort;
using coot::sort_index;
```

Looks like these were missing, good to add them too.

include/ensmallen_bits/bigbatch_sgd/bigbatch_sgd_impl.hpp
```diff
@@ -244,13 +244,10 @@ BigBatchSGD<UpdatePolicyType>::Optimize(
...
-      if (i != actualMaxIterations)
-      {
-        lastObjective = overallObjective;
-        overallObjective = 0;
-        currentFunction = 0;
-      }
+      // Reset the counter variables.
+      lastObjective = overallObjective;
+      overallObjective = 0;
+      currentFunction = 0;
```
rcurtin, line 250: Here is one other place where we should revert to the old implementation (see https://github.com/mlpack/ensmallen/pull/425#discussion_r2089631394).

```suggestion
      // Reset the counter variables if we will continue.
      if (i != actualMaxIterations)
      {
        lastObjective = overallObjective;
        overallObjective = 0;
        currentFunction = 0;
      }
```

include/ensmallen_bits/sgdr/snapshot_sgdr.hpp
```diff
@@ -170,12 +170,14 @@ class SnapshotSGDR
   bool& ExactObjective() { return optimizer.ExactObjective(); }
 
   //! Get the snapshots.
-  std::vector<arma::mat> Snapshots() const
+  template<typename MatType>
```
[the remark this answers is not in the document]
rcurtin: Actually will this even work correctly if `MatType` is different than whatever the decay policy is holding?

include/ensmallen_bits/sgd/update_policies/nesterov_momentum_update.hpp
```diff
@@ -69,10 +69,9 @@ class NesterovMomentumUpdate
     Policy(const NesterovMomentumUpdate& parent,
            const size_t rows,
            const size_t cols) :
-        parent(parent),
-        velocity(arma::zeros<MatType>(rows, cols))
+        parent(parent)
     {
-      // Nothing to do.
+      velocity.zeros(rows, cols);
```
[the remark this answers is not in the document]
zoq: Fixed in [503ecd1](https://github.com/mlpack/ensmallen/pull/425/commits/503ecd16adb5ada64c7d165a58a834be159b7e61)

include/ensmallen_bits/sgd/update_policies/momentum_update.hpp
```diff
@@ -93,10 +93,9 @@ class MomentumUpdate
      * @param cols Number of columns in the gradient matrix.
      */
     Policy(const MomentumUpdate& parent, const size_t rows, const size_t cols) :
-        parent(parent),
-        velocity(arma::zeros<MatType>(rows, cols))
+        parent(parent)
     {
-      // Nothing to do.
+      velocity.zeros(rows, cols);
```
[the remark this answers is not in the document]
zoq: Fixed in [503ecd1](https://github.com/mlpack/ensmallen/pull/425/commits/503ecd16adb5ada64c7d165a58a834be159b7e61)

tests/sdp_primal_dual_test.cpp
```diff
@@ -286,7 +285,7 @@ TEST_CASE("SmallMaxCutSdp","[SdpPrimalDualTest]")
 }
 
 // This test is deprecated and can be removed in ensmallen 2.10.0.
```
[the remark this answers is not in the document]
zoq: Resolved with [b3775d5](https://github.com/mlpack/ensmallen/pull/425/commits/b3775d535e71daea8c89dd366885baac2766e9b8)

include/ensmallen_bits/problems/aug_lagrangian_test_functions.hpp
rcurtin: It took a bit of work but I got this all handled in e2b67b22.

include/ensmallen_bits/sgdr/snapshot_sgdr.hpp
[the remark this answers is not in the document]
rcurtin: I don't know how that would be possible though.  The optimizers themselves don't carry information about the types to be optimized, because they are generally generic: that is, an `SGD` object can just as easily optimize an `arma::mat` as an `arma::fmat`, and so on and so forth.

But, then you can also have decay policies and step size update strategies that hold some information during optimization that depends on the matrix type---so for instance with momentum, you hold simple momentum parameters of type `double`.  That's easy enough, and you still don't need to have the type used during optimization.  But then you also need to hold the previous gradient/step, which has to have type equivalent to the matrix type used during optimization.  Hence the instantiated `Policy<MatType, GradType>` classes got created, which are created when an optimization starts and we know what the `MatType` and `GradType` are.

Those policy objects are accessible via members like `InstDecayPolicy()` and `InstUpdatePolicy()`, but because the enclosing class does not know the matrix type or gradient type, it's not possible to automatically return "the right thing"; the only thing we can return is the `Any&` object used internally.  Instead as far as I can tell there is no way around forcing the user to specify the appropriate `MatType` and `GradType`.  So I don't see any better thing we can do than tell users to use `Snapshots()` like this:

```
SnapshotSGDR sgdr(...);
FunctionType function = ...;
arma::mat coordinates = ...;
sgdr.Optimize(function, coordinates);

// if you want the snapshots, you have to specify the matrix type;
// if the matrix type is specified wrong, then this will throw an exception
const std::vector<arma::mat>& snapshots = sgdr.Snapshots<arma::mat>();
```

Do you have any idea for a better solution?  I don't want to make the matrix or gradient type a template parameter of the overall optimizer---that's confusing for users, especially since the optimization process itself is generally independent of the type of matrix being used.  (e.g. I want to keep the `MatType` template parameter a part of the member `Optimize()` function, not a part of the overall class)

include/ensmallen_bits/moead/moead.hpp
```diff
@@ -48,9 +48,13 @@ namespace ens {
  *   year={2008},
  * @endcode
  */
-template<typename InitPolicyType = Uniform,
-         typename DecompPolicyType = Tchebycheff>
-class MOEAD {
+template<
+    typename MatType = arma::mat,
+    typename ColType = typename ForwardType<MatType>::bcol,
+    typename CubeType = typename ForwardType<MatType>::bcube,
```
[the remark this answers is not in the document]
rcurtin: Thanks for refactoring these.  I updated documentation accordingly in 13c228c4.

include/ensmallen_bits/sa/sa_impl.hpp
rcurtin: Handled in dcbdcdc4.

include/ensmallen_bits/sa/sa_impl.hpp
rcurtin: Handled in dcbdcdc4.

include/ensmallen_bits/pso/pso.hpp
rcurtin, line 69: Ah, I think this one got overlooked---we should remove this template parameter from `PSO` too, and just store the bounds as `arma::mat`s and convert them to the desired type when `Optimize()` is called.

include/ensmallen_bits/sgdr/snapshot_sgdr.hpp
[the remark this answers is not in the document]
rcurtin: I handled this change in 1d3102c6.

include/ensmallen_bits/moead/moead_impl.hpp
zoq: Fixed in [a9ce40c](https://github.com/mlpack/ensmallen/pull/425/commits/a9ce40c322030b65a9efffa9841ae8db53fd2cb7)

rcurtin:
Nice, everything is handled here and I think this is ready to go once we get it to pass the builds.  I noticed that the `IQN_LogisticRegressionFunction - arma::fmat` test is failing in CI---that one actually fails quite a lot.  In #426 I adapted that test to allow five trials (the `LogisticRegressionFunctionTest()` function takes that as the last parameter).  You can either import that change here, or we can ignore the test failure and it will get fixed when #426 is merged; either way is fine with me.

rcurtin:
We'll need to get the FP16 PR merged to fix some of the tests before a release, but, really great to finally get this in!

PR #426, opened 2025-07-26
Test FP16 support when available with Armadillo

rcurtin:
This PR requires #425 to be merged, so, until then, the diff will be pretty mangled.  We also can't merge this PR first, and I will need to do some rebasing as that PR gets updated and completed.

Whenever Armadillo provides the `arma::hmat` type (FP16 matrices), this adds `arma::hmat` to the set of types that are tested.  Each test type (`arma::mat`, `arma::fmat`, `arma::hmat`) has its own set of tolerances, as FP16 convergence cannot be expected to be as good as FP32 or FP64 convergence.

The new file `tests/test_types.hpp` defines the set of supported test types and macros that are used in tandem with `TEMPLATE_TEST_CASE()` to provide single test cases that test all supported element types.

Throughout, to prevent `-Wnarrowing` compiler warnings, I had to make various modifications to explicitly cast `double`s to the element type of the optimization.  These do not change the functionality or behavior or ensmallen (at least tangibly), and when the element type is `double` this is identical code to before.

I tuned each of the tests so that they run 1000 times without any failures.

After this PR, it is possible without warnings to do FP16 optimizations with ensmallen.


rcurtin:
CC @conradsnicta

rcurtin:

conradsnicta:
@rcurtin I'll take a look next week (4 Aug onwards).  High level question though: is running optimisers with low precision elements (like fp16) practical?  The reduced precision and poor range can result in convergence issues, and/or missing/skipping optimum points.

From a high level point of view, if fp16 elements are to be used in machine learning, my initial thoughts are that model parameters are better found via optimisers running with fp32 or fp64 elements, followed by downgrading the parameters to fp16 precision as the final step.

As an example, the accuracy of simple operations like summation and mean calculation are quite bad with fp16.  The armadillo experimental branch has a workaround for the fp16 version of mean(), where elements are accumulated in fp32 mode, with the final result cast to fp16.

Sophisticated optimisers are far more complicated than the above example, meaning that there is a risk of internal amplification of errors due to precision loss.

(The obvious counter-example is the use of bfloat16 for training neural networks, but that's a very specific use case that also uses an element type with a far wider range than fp16).

rcurtin:
> is running optimisers with low precision elements (like fp16) practical? The reduced precision and poor range can result in convergence issues, and/or missing/skipping optimum points.

Mostly agree: I can imagine some niche cases where someone wants to fine-tune an mlpack network in reduced precision, which in turn means the use of FP16 with ensmallen.  So I don't think anyone is going to say "I have a completely new problem, let's optimize it in FP16"---you would end up there through various constraints, like, e.g., "I have a spacecraft with a power budget and I want to fine-tune on images I only have in space but due to the limited power resources it will be better to do coarse fine-tuning in FP16" or... something along those lines, I think you get the idea.

However more importantly, mlpack's neural networks (the primary target for FP16) support training, and so it is possible that a user could, e.g., make an FP16 network and throw some SGD at it.  So, fixing all the `-Wnarrowing` inside of ensmallen is necessary otherwise if a user tries to do this, they will get all kinds of `-Wnarrowing` warnings that seem scary and make FP16 seem unsupported.

It's also a nice and useful test for the correctness of FP16 to ensure that on various test problems, optimizers can converge correctly.  (Albeit with wider tolerances for exactly the reasons you mentioned.)

include/ensmallen_bits/lbfgs/lbfgs_impl.hpp
```diff
@@ -417,34 +440,43 @@ L_BFGS::Optimize(FunctionType& function,
     // least one descent step.
     // TODO: to speed this up, investigate use of arma::norm2est() in Armadillo
     // 12.4
-    if (arma::norm(gradient, 2) < minGradientNorm)
+    //
+    // If we are using FP16, norm() is not currently implemented, so we have to
+    // use norm2est() instead.
+    #if defined(ARMA_HAVE_FP16)
+    const ElemType gradNorm = std::is_same<ElemType, arma::fp16>::value ?
+        norm2est(gradient) : norm(gradient, 2);
```
conradsnicta, line 448: @rcurtin This sets off a red flag for me.  The norm() and norm2est() functions are closely related, but are not the same thing.  They can end up with potentially different results, meaning we have an inconsistency that has nothing to do with the reduced precision of the fp16 element type.

I'd suggest to use norm2est() regardless of the element type, though that would entail bumping the minimum  version of Armadillo.

Alternatively, we can add a workaround directly in Armadillo, where norm(x,2) is implemented for fp16 via internally converting the given matrix to fp32 format.  This in turn allows use of LAPACK functions.  Any modern hardware that does fp16 can also do fp32.

include/ensmallen_bits/moead/decomposition_policies/tchebycheff_decomposition.hpp
```diff
@@ -57,7 +57,7 @@ class Tchebycheff
                                     const VecType& idealPoint,
                                     const VecType& candidateFitness)
   {
-    return arma::max(weight % arma::abs(candidateFitness - idealPoint));
+    return max(weight % abs(candidateFitness - idealPoint));
```
conradsnicta, line 60: @rcurtin Suggest to be careful with removing the arma:: prefix.  I encountered issues where C++ compilers were getting confused between std::max() and arma::max(), due to poor implementations of the corresponding standard C++ library and the byzantine mess known as ADL: https://en.cppreference.com/w/cpp/language/adl.html

This change would essentially increase the risk of an ADL blowup happening all over again.

include/ensmallen_bits/moead/moead_impl.hpp
```diff
@@ -140,66 +171,72 @@ Optimize(std::tuple<ArbitraryFunctionType...>& objectives,
...
   for (BaseMatType& individual : population)
   {
-    individual = arma::randu<BaseMatType>(
-        iterate.n_rows, iterate.n_cols) - 0.5 + iterate;
+    individual = randu<BaseMatType>(
+        iterate.n_rows, iterate.n_cols) - ElemType(0.5) + iterate;
 
     // Constrain all genes to be within bounds.
-    individual = arma::min(arma::max(individual, castedLowerBound), castedUpperBound);
+    individual = min(max(individual, castedLowerBound), castedUpperBound);
```
conradsnicta, line 204: @rcurtin Same issue as per https://github.com/mlpack/ensmallen/pull/426/files#r2252873983

include/ensmallen_bits/lbfgs/lbfgs_impl.hpp
rcurtin: Since norm(x, 2) is now implemented in Armadillo, I reverted to `norm()` in 1d963d6c.

zoq (commented):
Once the build issue is fixed, this seems ready to be merged from my side.

include/ensmallen_bits/fw/atoms.hpp
```diff
@@ -96,8 +96,7 @@ class Atoms
       // Find possible atom to be deleted.
       arma::vec gap = sqTerm -
           currentCoeffs % trans(gradient.t() * currentAtoms);
-      arma::uword ind;
-      gap.min(ind);
+      arma::uword ind = gap.min();
```
conradsnicta, line 99: @rcurtin shouldn't this be arma::uword ind = gap.index_min() ?

conradsnicta (changes requested):

conradsnicta (approved):

PR #427, opened 2025-07-30
Add proximal gradient optimizers: `FBS`, `FISTA`, and `FASTA`

rcurtin:
I implemented these a very long time ago but only just now got the okay to contribute them back.

This PR implements three new optimizers for ensmallen:

 * `FBS`: forward-backward splitting
 * `FISTA`: fast iterative shrinkage-thresholding algorithm
 * `FASTA`: fast adaptive shrinkage/thresholding algorithm

These are differentiable optimizers, sort of, but they are specifically made for a class of functions that consist of both a differentiable part and a non-differentiable part:

```
h(x) = f(x) + g(x)
 - f(x) is differentiable
 - g(x) is an arbitrary non-differentiable function!
```

This situation arises all over the place.  For me, the interesting application is L1-constrained objective functions like this one:

```
argminᵦ ∑loss(xᵢ, yᵢ, β) + |β|₁
```

Here, the L1 penalty is not differentiable and is instead often solved with "proximal gradient" algorithms, that take a 'forward' step to improve the differentiable `f(x)` term, then a 'backward' step to ensure that the objective of the non-differentiable `g(x)` parameter did not increase too much.  That backward step can also be called 'shrinkage'.

Inside of ensmallen, I implemented this type of constrained objective function by having the `FBS`, `FISTA`, and `FASTA` functions take both `f(x)` and `g(x)` separately.  The classes `L1Penalty` and `L1Constraint` are implemented, and the documentation includes examples of how to do L1-regularized optimization with ensmallen.  It looks like this:

```
  // Construct the objective function f(x), and the penalty function g(x) with a
  // lambda value of 0.1.
  LinearRegressionEWGFunction lrf(data, responses);
  ens::L1Penalty g(0.1);

  // Create the FBS optimizer with default parameters, and optimize the function
  // f(x) + g(x) (i.e. L1-penalized linear regression).
  // The ens::FBS class can be replaced with any ensmallen proximal gradient
  // optimizer.
  ens::FBS fbs(g);
  arma::mat lrfParams(startingPoint);
  fbs.Optimize(lrf, lrfParams);
```

Originally, my aim was to implement newGLMNET, another proximal gradient optimizer (it is what is inside LIBLINEAR), and then compare whether ensmallen's newGLMNET was better or worse than these three optimizers for learning L1-constrained logistic regression models.  However, that project was never completed and funding situation changes mean it's on indefinite hold (honestly speaking I am unlikely to ever return to that effort).  At the same time, these three optimizers are tested and ready to go.


conradsnicta (approved):

coatless (changes requested):
@rcurtin minor nit: missing a `HISTORY.md` entry.

rcurtin:
@coatless thanks for pointing that out; fixed now in bdb029a.

zoq (approved):

rcurtin:
A couple unexpected fixes were necessary to make this build... once I get things to pass I'll go ahead and merge.

Issue #428, opened 2025-08-12
Account for batch size


andrewfurey21:
Currently mlpack layers return gradients that are the summed across a batch, when typically they should be averaged otherwise gradients will change drastically if batch size changes.

rcurtin:
I agree with this change---basically we'd need to go through all differentiable optimizers and divide the step size by the batch size.  This would be a "breaking" change so I would want to do a major version bump.  We could also get the Bandicoot, FP16, and maybe FASTA/FISTA/FBS PRs merged before that too.  I'll see if I can find the time to open a PR for it soon.

PR #431, opened 2025-08-27
Normalize step size to the batch size used

rcurtin:
This handles #428.  I went through all separable differentiable optimizers and found that all of them that use `SGD` internally do not normalize by the batch size when taking a step.  This problem also exists with `Eve`.  But other separable optimizers do not have this issue because they are techniques that specifically account for the batch size (like `BigBatchSGD`, or `SVRG`, or others).

This is a breaking change, so it will necessitate a major release.  I also added the `ENS_OLD_SEPARABLE_STEP_BEHAVIOR` macro to allow users to continue to use the old behavior so that it doesn't break current workflows.  I could be persuaded to remove that, but I thought it was a nice thing to do.

### DIFF
--- a/include/ensmallen_bits/lookahead/lookahead_impl.hpp
+++ b/include/ensmallen_bits/lookahead/lookahead_impl.hpp
@@ -188,7 +188,7 @@ Lookahead<BaseOptimizerType, DecayPolicyType>::Optimize(
     size_t batchSize = 1;
     // Check if the optimizer implements the BatchSize() method and use the
     // parameter for the objective calculation.
-    if (traits::HasBatchSizeSignature<BaseOptimizerType>::value)
+    if constexpr (traits::HasBatchSizeSignature<BaseOptimizerType>::value)
       batchSize = baseOptimizer.BatchSize();
 
     overallObjective = 0;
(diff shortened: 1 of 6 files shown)


include/ensmallen_bits/lookahead/lookahead_impl.hpp
```diff
@@ -188,7 +188,7 @@ Lookahead<BaseOptimizerType, DecayPolicyType>::Optimize(
     size_t batchSize = 1;
     // Check if the optimizer implements the BatchSize() method and use the
     // parameter for the objective calculation.
-    if (traits::HasBatchSizeSignature<BaseOptimizerType>::value)
+    if constexpr (traits::HasBatchSizeSignature<BaseOptimizerType>::value)
```
conradsnicta, line 191: `if constexpr` is C++17 onwards.  ensmallen requires only C++14.  I'd prefer to keep the standard at C++14 unless there is really a compelling reason to bump it.

In this case, `if constexpr` doesn't actually help to produce any more efficient machine compared to standard `if` expression.  Any optimising compiler (all the way back to C++98 days) will figure out that `traits::HasBatchSizeSignature<...>::value` is a compile time constant (even when declared via old-school `static const`) and optimise accordingly.

conradsnicta (changes requested):

conradsnicta (approved):

rcurtin:
The mlpack integration test failure is expected---once mlpack/mlpack#4002 is merged, that will be fixed.

PR #432, opened 2025-08-27
Fix compilation warnings

rcurtin:
This fixes some warnings I see when I compile ensmallen locally.  I also fixed the style in the MAF functions (or at least took a quick pass at it).


rcurtin:
Probably?  These are just warnings I saw compiling locally; not sure when they snuck in.  Either way they are easy enough to fix and this probably pre-empts CRAN-related frustrations.

conradsnicta (approved):

PR #435, opened 2025-09-02
Remove deprecated functionality for ensmallen 3.x release.

rcurtin:
This removes the `ParetoFront()` and `ParetoSet()` functions from the multi-objective optimizers, since we are about to do a major version bump of ensmallen.  There are also a couple other deprecated things I found and removed.


conradsnicta:
@rcurtin Recommend making a note in HISTORY.md specifying what was removed and what functionality should be used instead.  Many people tend to compile stuff while disregarding the deprecation warning messages. Only when something breaks they get interested and start looking at what happened. Having an explanation in HISTORY.md would help them fix their code.

conradsnicta (approved):

PR #438, opened 2025-09-09
Merge Bandicoot tests with other tests

rcurtin:
Many of the tests already are of the form

```
TEMPLATE_TEST_CASE("TestName", "[Optimizer]", ENS_ALL_TEST_TYPES)
```

so they work over all ensmallen test types.  We can do the same thing with the Bandicoot tests, so I modified the definition of `ENS_ALL_TEST_TYPES` and merged all of the Bandicoot-specific test types accordingly.

No actual tests should really be changing here, it is just a cleanup.


zoq (approved):

rcurtin:
67m for OpenCL and 48m for CUDA, looks good to me.  I changed around a few more tests but all the same optimizers are tested with GPUs.

Issue #437, opened 2025-09-04
Feature Request: Jacob's `Delta-Bar-Delta` Update Rule For `GradientDescent` Optimizer


ranjodhsingh1729:
# Description
I was trying to implement t-SNE for mlpack and required the delta bar delta update rule given bellow.

<img width="774" height="455" alt="Image" src="https://github.com/user-attachments/assets/afaa7318-a419-495c-977c-ddd8a51a81db" />

I am currently using a callback to achieve the same result. It would be better if this was provided by ensmallen itself.

I can make a PR if there's interest for such a addition.

# References
[[1](https://www2.bcs.rochester.edu/sites/robbie/jacobs.nn88.pdf)] R.A. Jacobs. Increased rates of convergence through learning rate adaptation. Neural
Networks, 1:295–307, 1988.

ranjodhsingh1729:
I wanted to refactor the gradient descent class to accept the update rule as a template parameter and add support for the delta-bar-delta update, which users could select instead of the default one.

I'm not entirely sure how well this works with SGD - I think it does, but I'll confirm. If it does, it would be great if both gradient descent and SGD could share the same update rule.

Initially, I used callbacks to modify the gradient before the optimizer applied the update - a hacky approach, but I was eager to see the algorithm working.

I ended up modifying the momentum update to implement this rule. It was really just a small change (4-5 lines), and I'm now using it with SGD (batch size = number of points).

t-SNE traditionally, and for good reason, uses the full batch gradient descent, so this change makes a lot of sense to me.

Callbacks are still handy for things like early exaggeration with reduced momentum, but I'm leaning toward handling that with two seperate optimizer instances with different paramters running one after the other since the hyperparameter adjustment only happens once.

I hope this clears things up, and apologies for not being clearer earlier.

rcurtin:
No worries, there are a lot of design questions so there are tons of different ways we could go here, it all depended on what the needs were.  My recommendation would be that we add `DeltaBarDelta` as a new optimizer to ensmallen, and the implementation strategy could be like this:

 * Change `GradientDescent` to `GradientDescentType` and add `UpdateRuleType` as a template parameter.
 * Implement `VanillaUpdate` with the regular step type.
 * Implement `DeltaBarDeltaUpdate` with the new one.
 * Typedef `GradientDescent = GradientDescentType<VanillaUpdate>` and `DeltaBarDelta = GradientDescentType<DeltaBarDeltaUpdate>`
[...]

PR #440, opened 2025-09-20
Implement `DeltaBarDelta` using refactored `GradientDescent`.

ranjodhsingh1729:
Closes: #437

This PR:
  - Implements `DeltaBarDeltaUpdate`
  - Refactors `GradientDescent` to `GradientDescentType<UpdatePolicyType, DecayPolicyType>`
  - Defines `DeltaBarDelta` as a typedef to `GradientDescentType<DeltaBarDeltaUpdate, NoDecay>`

`GradientDescent` is now a typedef to `GradientDescentType<VanillaUpdate, NoDecay>`.


ranjodhsingh1729:
Hi,
would it make sense to include a change here which allows `PrintLoss` callback to work with `GradientDescentType` (since `ProgressBar()` would require more invasive changes). It doesn't currently as `PrintLoss` prints when `EndEpoch` is called and `GradientDescentType` never calls `EndEpoch`.

ranjodhsingh1729:
@conradsnicta could you please approove the CI build i can't test this with bandicoot locally.

zoq:
Had to restart the GPU runner, restarted the jobs.

ranjodhsingh1729:
Thanks @zoq 
The Tests ran successfully.

ranjodhsingh1729:
Hi

> Fantastic, can you also update the HISTORY.md file.

Done.

> Hi, would it make sense to include a change here which allows `PrintLoss` callback to work with `GradientDescentType` (since `ProgressBar()` would require more invasive changes). It doesn't currently as `PrintLoss` prints when `EndEpoch` is called and `GradientDescentType` never calls `EndEpoch`.

What do you think about this.

Additionally I found this small inconsistency.
If `maxIterations` is non-zero it will perform one less iteration.

https://github.com/mlpack/ensmallen/blob/9aebc1d2e43122213f25bdb4a560a938108c2ccd/include/ensmallen_bits/gradient_descent/gradient_descent_impl.hpp#L70

Should I go ahead and fix these here? OR seperately?

ranjodhsingh1729:
> In general, fixes are orthogonal to new features, and should be kept separated.

Thanks @zoq for the great advice.
It made sense the moment i read it. :)

I think this is ready for review then.
This isn't urgent or anything whenever you(or anyone else) can find the time will be good.

rcurtin:
It is definitely not okay to merge an optimizer that has not been thoroughly reviewed.  I will get to it when I have a chance and there is no schedule on when that will be.

doc/optimizers.md
```diff
@@ -823,8 +823,6 @@ parameters.
 If `lambda` and `sigma` are not specified, then 0 is used as the initial value
 for all Lagrange multipliers and 10 is used as the initial penalty parameter.
 
-</details>
-
```
rcurtin, line 827: Thanks, nice catch!

doc/optimizers.md
```diff
@@ -1261,6 +1259,63 @@ optimizer.Optimize(f, coordinates);
...
+
+DeltaBarDelta optimizer(0.001, 0, 1e-15, DeltaBarDeltaUpdate(0.2, 0.8, 0.5, 0.01));
+optimizer.Optimize(f, coordinates);
+```
+
+</details>
+
+#### See also:
+
+ * [Increased rates of convergence through learning rate adaptation](https://www.academia.edu/download/32005051/Jacobs.NN88.pdf)
```
rcurtin:
```suggestion
 * [Increased rates of convergence through learning rate adaptation (pdf)](https://www.academia.edu/download/32005051/Jacobs.NN88.pdf)
```
Most of the other links indicate to people when a PDF is going to be opened/downloaded.

Also it's probably worth linking to `GradientDescent` here too.

doc/optimizers.md
```diff
@@ -1261,6 +1259,63 @@ optimizer.Optimize(f, coordinates);
...
+#### Constructors
+
+ * `DeltaBarDelta()`
+ * `DeltaBarDelta(`_`stepSize`_`)`
+ * `DeltaBarDelta(`_`stepSize, maxIterations, tolerance`_`)`
+ * `DeltaBarDelta(`_`stepSize, maxIterations, tolerance, updatePolicy, decayPolicy, resetPolicy`_`)`
+
+Note that `DeltaBarDelta` is based on the templated type
+`GradientDescentType<`_`UpdatePolicyType, DecayPolicyType`_`>` with _`UpdatePolicyType`_` =
+`DeltaBarDeltaUpdate` and _`DecayPolicyType`_` = NoDecay`.
```
rcurtin: Because `DeltaBarDelta` always uses the `DeltaBarDeltaUpdate` class, it doesn't really make sense to the user to provide a constructor that takes `updatePolicy`, `decayPolicy`, and `resetPolicy`.  Of course an advanced user can do this if they take a look at the internals of the `GradientDescent` class.  But for the typical user, they just want to set the settings of `DeltaBarDelta` and move on.  So I would suggest adding a wrapper class `DeltaBarDelta` that just calls `GradientDescentType<DeltaBarDeltaUpdate, NoDecay>` internally (there are many of these for SGD variants), and also provides a constructor that forwards the parameters of `DeltaBarDeltaUpdate`:

```
DeltaBarDelta(stepSize, maxIterations, tolerance, kappa, phi, momentum, minGain, resetPolicy)
```

Then the table of options below will be much simplified too.

include/ensmallen_bits/gradient_descent/gradient_descent.hpp
```diff
@@ -160,6 +185,16 @@ class GradientDescent
...
 
+  //! Get the update policy.
+  const UpdatePolicyType& UpdatePolicy() const { return updatePolicy; }
+  //! Modify the update policy.
+  UpdatePolicyType& UpdatePolicy() { return updatePolicy; }
+
+  //! Get the step size decay policy.
+  const DecayPolicyType& DecayPolicy() const { return decayPolicy; }
+  //! Modify the step size decay policy.
+  DecayPolicyType& DecayPolicy() { return decayPolicy; }
```
rcurtin, line 208: Shouldn't there be one for `ResetPolicy()` too?

include/ensmallen_bits/gradient_descent/gradient_descent_impl.hpp
```diff
@@ -20,25 +20,39 @@
...
-    stepSize(stepSize),
-    maxIterations(maxIterations),
-    tolerance(tolerance)
+    const double tolerance,
+    const UpdatePolicyType& updatePolicy,
+    const DecayPolicyType& decayPolicy,
+    const bool resetPolicy)
+    : stepSize(stepSize), maxIterations(maxIterations), tolerance(tolerance),
+      updatePolicy(updatePolicy), decayPolicy(decayPolicy),
+      resetPolicy(resetPolicy), isInitialized(false)
```
rcurtin:
```suggestion
    const bool resetPolicy) : 
    stepSize(stepSize),
    maxIterations(maxIterations),
    tolerance(tolerance),
    updatePolicy(updatePolicy),
    decayPolicy(decayPolicy),
    resetPolicy(resetPolicy),
    isInitialized(false)
```
Just a style fix.

include/ensmallen_bits/gradient_descent/gradient_descent_impl.hpp
```diff
@@ -74,6 +113,18 @@ GradientDescent::Optimize(FunctionType& function,
...
+    instUpdatePolicy.As<InstUpdatePolicyType>().Update(iterate,
+                                                       stepSize,
+                                                       gradient);
+
+    terminate |= Callback::StepTaken(*this, f, iterate, callbacks...);
+
+    // Now update the learning rate if requested by the user.
+    instDecayPolicy.As<InstDecayPolicyType>().Update(iterate,
+                                                     stepSize,
+                                                     gradient);
```
rcurtin: I wouldn't move the step up here; if you do that, it means that the user never sees what the objective is at the first iteration (e.g. before optimization started), which can be confusing.

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * Delta-Bar-Delta update for Gradient Descent.
+ *
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_GRAIDENT_DESCENT_DELTA_BAR_DELTA_UPDATE_HPP
+#define ENSMALLEN_GRAIDENT_DESCENT_DELTA_BAR_DELTA_UPDATE_HPP
```
rcurtin:
```suggestion
#ifndef ENSMALLEN_GRADIENT_DESCENT_DELTA_BAR_DELTA_UPDATE_HPP
#define ENSMALLEN_GRADIENT_DESCENT_DELTA_BAR_DELTA_UPDATE_HPP
```

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+namespace ens {
+
+/**
+ * DeltaBarDelta Update Policy for Gradient Descent
+ *
+ * A heuristic designed to accelerate convergence by
+ * adapting the learning rate of each paramenter individually.
```
rcurtin:
```suggestion
 * adapting the learning rate of each parameter individually.
```

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * - If the current gradient and the exponential moving average of
+ *   previous gradients corresponding to a parameter have the same
+ *   sign, then the learning rate for that parameter is incremented by
+ *   \f$\kappa\f$. Otherwise, it is multiplied by \f$\phi\f$ (additive
+ *   increase, multiplicative decrease).
+ *
+ * @note This implementation uses a MinimumGain parameter to set a lower
+ *     bound for the learning rate. This prevents the learing rate from
+ *     droping to zero, which can occur due to floating-point underflow.
```
rcurtin:
```suggestion
 *     bound for the learning rate. This prevents the learning rate from
 *     dropping to zero, which can occur due to floating-point underflow.
```

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+ *     For tasks which require extreme fine-tuning, you may need to lower
+ *     this parameter than its default value(0.01) in order to allow for 
+ *     smaller learing rates.
```
rcurtin:
```suggestion
 *     this parameter below its default value(0.01) in order to allow for 
 *     smaller learning rates.
```

Also, is 0.01 really a good value for this?  That is very, very far away from floating-point underflow.  If the objective is just to prevent overflow I might consider more like 1e-8 or even 1000 * numeric_limits<ElemType>::epsilon() or similar.

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * @code
+ * @article{jacobs1988increased,
+ *   title   = {Increased Rates of Convergence Through Learning Rate
+ * Adaptation}, author  = {Jacobs, Robert A.}, journal = {Neural Networks},
```
rcurtin:
```suggestion
 *   title   = {Increased Rates of Convergence Through Learning Rate
 *              Adaptation},
 *   author  = {Jacobs, Robert A.},
 *   journal = {Neural Networks},
```

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+  //! The momentum hyperparameter.
+  double momentum;
+
+  //! The minGain hyperparameter
+  double minGain;
+};
+
+} // namespace ens
+
+#endif // ENSMALLEN_GRAIDENT_DESCENT_DELTA_BAR_DELTA_UPDATE_HPP
```
rcurtin:
```suggestion
#endif // ENSMALLEN_GRADIENT_DESCENT_DELTA_BAR_DELTA_UPDATE_HPP
```

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+   private:
+    //! The instantiated parent class.
+    const DeltaBarDeltaUpdate& parent;
+
+    //! The gains matrix.
+    MatType gains;
+
+    //! The velocity matrix.
+    MatType velocity;
```
rcurtin: Take a look at some of the other Policy classes for the SGD update policies: so that they compile correctly when the element type is FP16 (e.g. `std::float16_t`), I modified the Policy child classes to hold `ElemType` versions of the parameters in the parent class.  This allows users to specify their parameters as doubles because it can be inconvenient to specify them as FP16 (e.g. `0.0` will not work right and will throw a warning), but then also avoids the warnings when those are used.

In the actual update step, you would use the child-class versions instead of, e.g., `parent.Kappa()` and `parent.Phi()`.

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+     */
+    void Update(MatType& iterate,
+                const double stepSize,
+                const GradType& gradient)
+    {
+      const MatType mask = conv_to<MatType>::from((gradient % velocity) < 0.0);
+      
+      gains += mask * parent.Kappa();
+      gains %= mask + (1 - mask) *  parent.Phi();
+      gains.clamp(parent.MinimumGain(), arma::datum::inf);
```
rcurtin:
```suggestion
      gains.clamp(parent.MinimumGain(), arma::Datum<typename MatType::elem_type>::inf);
```
`MatType` may not always be `arma::mat`, so we need to make sure we use the right type here.

tests/delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+#endif
+#include <ensmallen.hpp>
+#include "catch.hpp"
+#include "test_function_tools.hpp"
+
+using namespace ens;
+using namespace ens::test;
+
+TEMPLATE_TEST_CASE("DeltaBarDelta_GDTestFunction", "[DeltaBarDelta]",
+   arma::mat, arma::fmat)
```
rcurtin: Can you use `ENS_ALL_TEST_TYPES` here?  That will also handle the Bandicoot types too, so you could remove them from below.

tests/delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+  FunctionTest<GDTestFunction, TestType>(s, 0.1, 0.01);
+}
+
+TEMPLATE_TEST_CASE("DeltaBarDelta_RosenbrockFunction", "[DeltaBarDelta]",
+    arma::mat, arma::fmat)
+{
+  DeltaBarDelta s(0.001, 0, 1e-15);
+  FunctionTest<RosenbrockFunction, TestType>(s, 0.01, 0.001);
+}
+
```
rcurtin, line 40: Can you add one more test for the logistic regression test function?  That is a really simple test but it is a good sanity check for machine learning applications.  It should be really easy; here's the Adam logistic regression function test:

```c++
TEMPLATE_TEST_CASE("Adam_LogisticRegressionFunction", "[Adam]",
    ENS_ALL_TEST_TYPES)
{
  Adam adam(0.032);
  LogisticRegressionFunctionTest<TestType>(adam);
}
```

tests/delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,51 @@
+/**
+ * @file delta_bar_detla_test.cpp
```
rcurtin:
```suggestion
 * @file delta_bar_delta_test.cpp
```

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+     * @param gradient The gradient matrix.
+     */
+    void Update(MatType& iterate,
+                const double stepSize,
+                const GradType& gradient)
+    {
+      const MatType mask = conv_to<MatType>::from((gradient % velocity) < 0.0);
+      
+      gains += mask * parent.Kappa();
+      gains %= mask + (1 - mask) *  parent.Phi();
```
rcurtin:
```suggestion
      gains %= mask + (1 - mask) * parent.Phi();
```

rcurtin: Based on my read of Eq. (4) in the paper I'm not sure this is correct.  The value of `gains[i]` should not change if `(gradient % velocity)[i]`  is 0; but given how `mask` is set, I can't quickly see that this is correct.

I would suggest maybe using `sign()` instead to create a matrix indicating whether the element is positive, negative, or zero, and then use that matrix to apply the update rule in (4).

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+      gains.clamp(parent.MinimumGain(), arma::datum::inf);
+
+      velocity *= parent.momentum;
+      velocity -= (stepSize * gains) % gradient;
+      iterate += velocity;
```
rcurtin: But the paper seems to indicate we update with `ϵ(t + 1)` in Eq. (3), not `ẟ'(t)`?

I would suggest using the same terminology as the paper throughout even if it is not intuitive what `ẟ'`, `ϵ`, and `θ` actually mean---you can clarify that part in the comments.

include/ensmallen_bits/gradient_descent/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,166 @@
...
+                const GradType& gradient)
+    {
+      const MatType mask = conv_to<MatType>::from((gradient % velocity) < 0.0);
+      
+      gains += mask * parent.Kappa();
+      gains %= mask + (1 - mask) *  parent.Phi();
+      gains.clamp(parent.MinimumGain(), arma::datum::inf);
+
+      velocity *= parent.momentum;
+      velocity -= (stepSize * gains) % gradient;
```
rcurtin: How does this implement the exponential average `ẟ'(t) = (1 - θ) ẟ(t)  + θ ẟ'(t - 1)`?  This doesn't appear to be that.

ranjodhsingh1729:
@rcurtin, Thanks a lot for the thorough review - turns out i've made quite a mistake here.

Since both t-SNE papers (the original and the tree-based approximation) state using Jacobs's scheme and cite this [paper](https://www2.bcs.rochester.edu/sites/robbie/jacobs.nn88.pdf), I assumed it was used as-is and didn't verifying that assumption.

Now i know that what is used in t-SNE was a modified version of the DBD rule that incorporates momentum, while the original DBD only includes momentum for 'comparison' and doesn't apply it in updates.

Glad this was caught during review - definitely a reminder not to rush things like this in the future. 😅

Since there's already an open issue, PR, and review, I've gone ahead and implemented the actual delta-bar-delta rule and also addressed the review suggestions.

So I guess this puts me back at square one - I'll need to implement the t-SNE update rule in mlpack and hook it up with the refactored gradient descent class. Since it's not really used elsewhere, adding it to ensmallen probably doesn't make sense. Does that sound right, or is there a cleaner way to handle it?

And sorry for the confusion here.

rcurtin:
I think it's just fine to add Delta-Bar-Delta with momentum here.  It would add to the list of optimizers that ensmallen supports, no problem at all with that even if the target use (for now) is just t-SNE.  You could either:

 * add options to the existing `DeltaBarDelta` optimizer for momentum
 * add a new `MomentumDeltaBarDelta` optimizer rule that is `DeltaBarDelta` with momentum added

I'm not picky about which, though I might lean very slightly towards the second option, since we have separated `SGD` and `MomentumSGD`.

ranjodhsingh1729:
I should have clarified this in my last response: tsne adds momentum yes but it also makes some further modifications:

Here is a little snippet showing both:
```python
"""Assumes epsilon is initialized with step_size"""
def dbd():
    sameSign = delta_bar * delta > 0.0
    diffSign = delta_bar * delta < 0.0

    epsilon[sameSign] += kappa
    epsilon[diffSign] -= phi * epsilon[diffSign]
    np.clip(epsilon, min_gain, np.inf, out=epsilon)

    delta_bar = theta * delta_bar + (1 - theta) * delta
    Y -= epsilon * delta

"""
Assumes delta_epsilon is initialized with zeros
Effective step_size = step_size * delta_epsilon

diffSign => increment epsilon, sameSign => decrement epsilon
This is opposite of what the dbd does.

decrement is by (1 - phi) * delta_epsilon (written here as *= phi)
whereas it was by phi * epsilon in original dbd

no delta_bar is present here
velocity is used for comparison
"""
def tsne():
    sameSign = velocity * delta > 0.0
    diffSign = velocity * delta < 0.0

    delta_epsilon[diffSign] += kappa
    delta_epsilon[sameSign] *= phi
    np.clip(delta_epsilon, min_gain, np.inf, out=delta_epsilon)

    velocity = theta * velocity - step_size * delta_epsilon * delta
    Y += update
```

These changes are not equivalent to simply adding momentum to the final update like this:
```python
"""Assumes epsilon is initialized with step_size"""
def momentum_dbd():
    sameSign = delta_bar * delta > 0.0
    diffSign = delta_bar * delta < 0.0

    epsilon[sameSign] += kappa
    epsilon[diffSign] -= phi * epsilon[diffSign]
    np.clip(epsilon, min_gain, np.inf, out=epsilon)

    delta_bar = theta * delta_bar + (1 - theta) * delta
    velocity = theta * velocity - epsilon * delta
    Y += velocity
```

So essentially tsne just uses the idea behind dbd with momentum.
Now given this i am not sure if implementing it under `MomentumDeltaBarDelta` would be okay. Let me know what you think!

rcurtin:
I see what you mean, thanks for the clarification.  I haven't spent enough time with the t-SNE paper yet to understand exactly why they chose to do things differently but I will get to that when I have a chance.  In the mean time it sounds like the route to a solution is to use two different update rule classes for each optimizer.  I guess `MomentumDeltaBarDelta` works for the t-SNE version but if you have a better idea for a name I'm all ears.  (The thought enters my mind of whether the paper has a typo in the algorithm description---that definitely does happen more than one would hope for peer-reviewed papers.  I'll think about it more as I have a chance to jump in.)

ranjodhsingh1729:
> The thought enters my mind of whether the paper has a typo in the algorithm description

Both the papers (the original and the tree-based approximation) stated using "Jacob's scheme". I just assumed (mistakenly) it was dbd since dbd was the only thing that looked similar in the cited jacob's paper.

Papers didn't mention any modifications those were only given in the implementation given on lvd's page

rcurtin:
That's totally fine, like I said, we can implement both as separate rules.  That's probably the best thing to do here since you already took the time to get the original DBD rule ready.  Does that work for you?

ranjodhsingh1729:
Sure i'll implement the other rule under `MomentumDeltaBarDelta` as you suggested (I don't see any other way either).

ranjodhsingh1729:
Sorry for the delay in getting to this. I've noted a few things i am not entirely sure about:

In `DeltaBarDeltaUpdate`
The class is passed with `initialStepSize` in order to fill the `epsilon` matrix. Another solution could be:
```diff
diff --git a/compare.1 b/compare.2
index 673c847..fef8c17 100644
--- a/compare.1
+++ b/compare.2
@@ -3,11 +3,11 @@
       const MatType diffSignMask = conv_to<MatType>::from(signMatrix == -1);
 
       epsilon += sameSignMask * kappa;
-      epsilon -= diffSignMask * phi % epsilon;
-      epsilon.clamp(minStepSize,
+      epsilon -= diffSignMask * phi % (stepSize + epsilon);
+      epsilon.clamp(minStepSize - stepSize,
           arma::Datum<typename MatType::elem_type>::inf);
 
       delta_bar *= theta;
       delta_bar += (1 - theta) * delta;
 
-      iterate -= epsilon % delta;
+      iterate -= (stepSize + epsilon) % delta;
```

In `MomentumDeltaBarDeltaUpdate`
Since actual `stepSize` used for update is `stepSize * gains`, gains had to be clamped this way. Another way could be to take `minGain` as input as done previously.
```cpp
      // stepSize * gains >= minStepSize
      // gains >= minStepSize / stepSize
      gains.clamp((double) minStepSize / stepSize,
          arma::Datum<typename MatType::elem_type>::inf);
```

tests/delta_bar_delta_test.cpp
ranjodhsingh1729: Done

rcurtin:
For the first thing, I'm not sure what the advantage of the second approach is; for `DeltaBarDelta`, the user specifies a step size in the constructor.  This is the initial step size of the entire optimization, which is then tuned and updated by the rule that you've implemented.  So I don't see what the purpose of the extra changes are.  I think it makes more sense with the `a/compare.1` implementation.

For `MomentumDeltaBarDelta`, I'm not sure I understand the full question: I agree with what you wrote, but is this really a question about the user interface and what options are available?  The original MATLAB implementation gives parameters that are the initial step size (e.g. initial epsilon value) and the minimum gain, so I might lean towards exposing those parameters for `MomentumDeltaBarDelta` instead of a different approach.

I'm not 100% sure I understood the questions right, so please clarify and point me in the right direction if I didn't. :)

ranjodhsingh1729:
@rcurtin
Thanks for the detailed response. You are correct on both counts. My questions were just for confirmation and/or input. I've changed `minStepSize` to `minGain` in `MomentDeltaBarDelta`. Another review at your convenience would be appreciated (no urgency).

HISTORY.md
```diff
@@ -1,5 +1,9 @@
 ### ensmallen ?.??.?: "???"
 ###### ????-??-??
+ * Refactor `GradientDescent` into
+   `GradientDescentType<UpdatePolicyType, DecayPolicyType>` and
+   add the `DeltaBarDelta` and `MomentumDeltaBarDelta` optimizers
+   ([#440](https://github.com/mlpack/ensmallen/pull/440)).
```
rcurtin, line 6:
```suggestion
   ([#440](https://github.com/mlpack/ensmallen/pull/440)).

```

doc/optimizers.md
```diff
@@ -1261,6 +1259,58 @@ optimizer.Optimize(f, coordinates);
...
+| **type** | **name** | **description** | **default** |
+|----------|----------|-----------------|-------------|
+| `double` | **`stepSize`** | Initial step size. | `1.0` |
+| `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+| `double` | **`tolerance`**  | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+| `double` | **`kappa`** | Additive increase constant for step size when gradient signs persist. | `0.2` |
+| `double` | **`phi`** | Multiplicative decrease factor for step size when gradient signs flip. | `0.2` |
+| `double` | **`theta`** | Decay rate for computing the exponential average of past gradients. | `0.5` |
+| `double` | **`minStepSize`** | Minimum allowed step size for any parameter. | `1e-8` |
+| `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call. | `true` |
```
rcurtin:
```suggestion
| `bool` | **`resetPolicy`** | If true, parameters are reset before every `Optimize()` call. | `true` |
```

doc/optimizers.md
```diff
@@ -1261,6 +1259,58 @@ optimizer.Optimize(f, coordinates);
...
+
+Attributes of the optimizer may be accessed and modified via member functions of the same name.
```
rcurtin: Just for consistency, do you mind naming the member functions, like in the other optimizers' documentation?

doc/optimizers.md
```diff
@@ -1907,9 +1962,14 @@ negative of the gradient of the function at the current point.
 | `double` | **`stepSize`** | Step size for each iteration. | `0.01` |
 | `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
 | `size_t` | **`tolerance`**  | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+| `UpdatePolicyType` | **`updatePolicy`** | Instantiated update policy used to adjust the given parameters. | `UpdatePolicyType()` |
+| `DecayPolicyType` | **`decayPolicy`** | Instantiated decay policy used to adjust the step size. | `DecayPolicyType()` |
+| `bool` | **`resetPolicy`** | Flag that determines whether update policy parameters are reset before every Optimize call. | `true` |
```
rcurtin: I'm not sure we should document these parameters here. It adds some amount of confusion to `GradientDescent`, since the default update and decay policies have no parameters to configure, and if you are using those, these three new parameters make no sense. Given that this is one of the first optimizers users are likely to encounter, I thunk we should keep this documentation simple.  If a user wants a more complex alternative, they should use `DeltaBarDelta` or `MomentumDeltaBarDelta`.  So, I do think it is worthwhile to add a note directing users to those optimizers as alternatives.

doc/optimizers.md
```diff
@@ -2396,6 +2456,58 @@ for all Lagrange multipliers and 10 is used as the initial penalty parameter.
  * [Semidefinite programming on Wikipedia](https://en.wikipedia.org/wiki/Semidefinite_programming)
  * [Semidefinite programs](#semidefinite-programs) (includes example usage of `PrimalDualSolver`)
 
+## Momentum DeltaBarDelta
+
+*An optimizer for [differentiable functions](#differentiable-functions).*
+
+A DeltaBarDelta variant that incorporates momentum and other modifications.
```
rcurtin: "other modifications" is a little vague, do you think you could be a little more clear about what the differences are?

doc/optimizers.md
```diff
@@ -2396,6 +2456,58 @@ for all Lagrange multipliers and 10 is used as the initial penalty parameter.
...
+| `double` | **`stepSize`** | Initial step size. | `1.0` |
+| `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+| `double` | **`tolerance`**  | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+| `double` | **`kappa`** | Additive increase constant for step size. | `0.2` |
+| `double` | **`phi`** | Multiplicative decrease factor for step size. | `0.8` |
+| `double` | **`momentum`** | The momentum decay hyperparameter. | `0.5` |
+| `double` | **`minGain`** | Minimum allowed gain (scaling factor) for any parameter. | `1e-8` |
+| `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call. | `true` |
+
+Attributes of the optimizer may be accessed and modified via member functions of the same name.
```
rcurtin: Same here, can we explicitly name the functions like in the other optimizers' documentation?

include/ensmallen_bits/delta_bar_delta/delta_bar_delta.hpp
```diff
@@ -0,0 +1,178 @@
...
+ * ensmallen is free software; you may redistribute it and/or modify it under
+ * the terms of the 3-clause BSD license.  You should have received a copy of
+ * the 3-clause BSD license along with ensmallen.  If not, see
+ * http://www.opensource.org/licenses/BSD-3-Clause for more information.
+ */
+#ifndef ENSMALLEN_DELTA_BAR_DELTA_HPP
+#define ENSMALLEN_DELTA_BAR_DELTA_HPP
+
+#include <ensmallen_bits/gradient_descent/gradient_descent.hpp>
+#include "./update_policies/delta_bar_delta_update.hpp"
```
rcurtin:
```suggestion
#include "update_policies/delta_bar_delta_update.hpp"
``
Just a simplification (it makes no difference really).

File: include/ensmallen_bits/delta_bar_delta/delta_bar_delta.hpp [hunk 28]
```diff
@@ -0,0 +1,178 @@
...
+ *   sign, then the step size for that parameter is incremented by
+ *   \f$\kappa\f$. Otherwise, it is decreased by a proportion \f$\phi\f$
+ *   of its current value (additive increase, multiplicative decrease).
+ *
+ * @note This implementation uses a minStepSize parameter to set a lower
+ *     bound for the learning rate. This prevents the learning rate from
+ *     dropping to zero, which can occur due to floating-point underflow.
+ *     For tasks which require extreme fine-tuning, you may need to lower
+ *     this parameter below its default value (1e-8) in order to allow for
+ *     smaller learning rates.
```
rcurtin, line 39: Worth adding this to the documentation page? Users are unlikely to see it here.

include/ensmallen_bits/delta_bar_delta/delta_bar_delta.hpp
```diff
@@ -0,0 +1,178 @@
...
+ *
+ * @code
+ * @article{jacobs1988increased,
+ *   title     = {Increased Rates of Convergence Through Learning Rate
+ *                Adaptation},
+ *   author    = {Jacobs, Robert A.}, journal = {Neural Networks},
```
rcurtin:
```suggestion
 *   author    = {Jacobs, Robert A.},
 *   journal   = {Neural Networks},
```

include/ensmallen_bits/delta_bar_delta/momentum_delta_bar_delta.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+namespace ens {
+
+/**
+ * MomentumDeltaBarDelta Optimizer.
+ *
+ * A DeltaBarDelta variant that incorporates momentum and other modifications.
+ * Note: This is the variant used for optimizing the t-SNE cost function.
+ *
+ * @code
```
rcurtin, line 47: Maybe better to link to the t-SNE paper instead?

include/ensmallen_bits/delta_bar_delta/momentum_delta_bar_delta.hpp
```diff
@@ -0,0 +1,161 @@
...
+   *     call; otherwise, their values are retained.
+   */
+  MomentumDeltaBarDelta(const double stepSize = 1.0,
+                const size_t maxIterations = 100000,
+                const double tolerance = 1e-5,
+                const double kappa = 0.2,
+                const double phi = 0.8,
+                const double momentum = 0.5,
+                const double minGain = 1e-8,
+                const bool resetPolicy = true);
```
rcurtin: Can you fix the indentation here?

include/ensmallen_bits/gradient_descent/gradient_descent.hpp
rcurtin: Looks like this comment still hasn't been handled, we should definitely do that before merge.

tests/delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <ensmallen.hpp>
+#include "catch.hpp"
+#include "test_function_tools.hpp"
+#include "test_types.hpp"
+
+using namespace ens;
+using namespace ens::test;
+
+TEMPLATE_TEST_CASE("DeltaBarDelta_GDTestFunction", "[DeltaBarDelta]",
+   ENS_ALL_TEST_TYPES)
```
rcurtin:
```suggestion
    ENS_ALL_TEST_TYPES)
```

tests/delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+      0.0001, 0.2, 0.5);
+  FunctionTest<RosenbrockFunction, TestType>(s,
+      10 * Tolerances<TestType>::LargeObj,
+      10 * Tolerances<TestType>::LargeCoord);
+}
+
+TEMPLATE_TEST_CASE("DeltaBarDelta_LogisticRegressionFunction",
+    "[DeltaBarDelta]", ENS_ALL_TEST_TYPES)
+{
+  DeltaBarDelta s(0.00032, 32, Tolerances<TestType>::Obj,
```
rcurtin: The parameters don't make much sense to me---why set the number of iterations to 32? This looks like the settings were taken from another test for a separable optimizer, but we should use parameters specific to DeltaBarDelta here.

tests/momentum_delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+    0.2, 0.8, 0.5);
+  FunctionTest<RosenbrockFunction, TestType>(s,
+      10 * Tolerances<TestType>::LargeObj,
+      10 * Tolerances<TestType>::LargeCoord);
+}
+
+TEMPLATE_TEST_CASE("MomentumDeltaBarDelta_LogisticRegressionFunction",
+    "[MomentumDeltaBarDelta]", ENS_ALL_TEST_TYPES)
+{
+  MomentumDeltaBarDelta s(0.00032, 32, Tolerances<TestType>::Obj,
```
rcurtin: Same comment here about parameter settings.

include/ensmallen_bits/delta_bar_delta/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,201 @@
...
+    void Update(MatType& iterate,
+                const double stepSize,
+                const GradType& delta)
+    {
+      const MatType signMatrix = sign(delta % delta_bar);
+      const MatType incrementMask = conv_to<MatType>::from(signMatrix == +1);
+      const MatType decrementMask = conv_to<MatType>::from(signMatrix == -1);
+
+      epsilon += incrementMask * kappa;
+      epsilon -= decrementMask * phi % epsilon;
```
rcurtin: I would suggest inlining the definition of `incrementMask` and `decrementMask`; you won't need the `conv_to` and it should be faster because Armadillo should be able to avoid ever actually materializing the mask matrices.  If you further make these both on one line, instead of two, it will be slightly uglier but should also be faster.

include/ensmallen_bits/delta_bar_delta/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,201 @@
...
+      epsilon.clamp(minStepSize,
+          arma::Datum<typename MatType::elem_type>::inf);
+
+      delta_bar *= theta;
+      delta_bar += (1 - theta) * delta;
```
rcurtin: Here too I would suggest condensing these to be on one line instead of two.  Armadillo is able to optimize complex expressions, but not across line boundaries.

include/ensmallen_bits/delta_bar_delta/update_policies/momentum_delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,172 @@
...
+     */
+    void Update(MatType& iterate,
+                const double stepSize,
+                const GradType& gradient)
+    {
+      const MatType mask = conv_to<MatType>::from(
+          sign(gradient) == sign(velocity));
+
+      gains += (1 - mask) * kappa;
+      gains -= mask * (1 - phi) % gains;
```
rcurtin: So when I am suggesting inlining things and compressing them to one expression, here is an example of what I mean:

```suggestion
      gains = (sign(gradient) != sign(velocity)) * kappa) -
          gains % (sign(gradient) == sign(velocity)) * (1 - phi);
```
I think I did the math right there still, but it's definitely worth double-checking.

include/ensmallen_bits/delta_bar_delta/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+      iterate -= epsilon % delta;
+    }
+
+   private:
+    //! The instantiated parent class.
+    const DeltaBarDeltaUpdate& parent;
+
+    //! The exponential average of past gradients.
+    MatType delta_bar;
```
rcurtin:
```suggestion
    MatType deltaBar;
```
Just to match the style guide, we use camel-case vs. snake-case.

include/ensmallen_bits/delta_bar_delta/update_policies/momentum_delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,172 @@
...
+      gains.clamp(minGain, arma::Datum<typename MatType::elem_type>::inf);
+
+      velocity *= momentum;
```
rcurtin: The original MATLAB implementation switches to a "final momentum" after 250 iterations.  Do you think it is necessary to add that here?  I'm okay with it either way, but it may help downstream convergence for t-SNE in mlpack.

doc/function_types.md
```diff
@@ -135,6 +135,7 @@ The following optimizers can be used with differentiable functions:
  * [Fast Adaptive Shrinkage/Thresholding Algorithm (FASTA)](#fast-adaptive-shrinkage-thresholding-algorithm-fasta) (`ens::FASTA`)
  * [FrankWolfe](#frank-wolfe) (`ens::FrankWolfe`)
  * [GradientDescent](#gradient-descent) (`ens::GradientDescent`)
+ * [DeltaBarDelta](#delta-bar-delta) (`ens::DeltaBarDelta`)
```
rcurtin: Also `MomentumDeltaBarDelta`?

doc/optimizers.md
```diff
@@ -2396,6 +2456,58 @@ for all Lagrange multipliers and 10 is used as the initial penalty parameter.
...
+#### Attributes
+
+| **type** | **name** | **description** | **default** |
+|----------|----------|-----------------|-------------|
+| `double` | **`stepSize`** | Initial step size. | `1.0` |
+| `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+| `double` | **`tolerance`**  | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+| `double` | **`kappa`** | Additive increase constant for step size. | `0.2` |
+| `double` | **`phi`** | Multiplicative decrease factor for step size. | `0.8` |
+| `double` | **`momentum`** | The momentum decay hyperparameter. | `0.5` |
```
rcurtin:
```suggestion
| `double` | **`momentum`** | The momentum hyperparameter. | `0.5` |
```
I don't know that `decay` is correct there?  Correct me if I'm wrong.

doc/optimizers.md
```diff
@@ -2396,6 +2456,58 @@ for all Lagrange multipliers and 10 is used as the initial penalty parameter.
...
+| `double` | **`minGain`** | Minimum allowed gain (scaling factor) for any parameter. | `1e-8` |
+| `bool` | **`resetPolicy`** | If true, parameters are reset before every Optimize call. | `true` |
```
rcurtin:
```suggestion
| `bool` | **`resetPolicy`** | If true, parameters are reset before every `Optimize()` call. | `true` |
```

include/ensmallen_bits/delta_bar_delta/update_policies/momentum_delta_bar_delta_update.hpp
ranjodhsingh1729: I think that part should be done inside t-SNE, since we also need to scale the the input joint probabilities by a early exaggeration factor for the first 250 iterations. Currently, I'm running the optimizer in two parts first the 250 iterations and then the rest, instead of adding a callback to switch at 250.

include/ensmallen_bits/delta_bar_delta/update_policies/delta_bar_delta_update.hpp
ranjodhsingh1729: Nice! I didn't know that Armadillo could do that. (I guess i'll have to read up on the how)

I've made the changes and it looks much better now :)

tests/delta_bar_delta_test.cpp
ranjodhsingh1729: I've adjusted the parameters by reducing the number of iterations as much as possible while using the highest learning rate that still works.
EDIT: had to increase the iterations bcz the tests were failing in ci

I also wanted to ask; are the parameters used for the other tests selected according to any particular procedure or guideline?

include/ensmallen_bits/delta_bar_delta/momentum_delta_bar_delta.hpp
ranjodhsingh1729: Does linking to both work?

doc/optimizers.md
ranjodhsingh1729: Please let me know how the write-up looks to you

include/ensmallen_bits/gradient_descent/gradient_descent.hpp
ranjodhsingh1729: I added it above the `UpdatePolicy()`. Should that be rearranged?

ranjodhsingh1729: I added it above the UpdatePolicy(). Should that be rearranged?

ranjodhsingh1729:
> Hey @ranjodhsingh1729, sorry that this took me so long to get back to. I verified the implementations using the original Delta-Bar-Delta paper and the MATLAB t-SNE reference implementation. Everything looks good implementation-wise. I have a whole bunch of other comments but they should be pretty easy and quick to handle. If anything I wrote is wrong don't hesitate to call me out on it---you've spent more time in the details of these algorithms than I have. Anyway once those comments are done let's get this merged and we can finally get back to the original task of t-SNE. :)
> Thanks for the hard work on this!

Hi @rcurtin, no worries - i was also tied up in academics. Thanks for your time and patient with this. It ended up taking far more effort than i anticipated, and i made mistakes, but learned a lot. I've taken care of all the comments (i think). If i missed anything or something still needs changes, let me know.

include/ensmallen_bits/delta_bar_delta/momentum_delta_bar_delta.hpp
rcurtin: Yep, that's just fine!  Thanks.

doc/optimizers.md
```diff
@@ -2396,6 +2466,67 @@ for all Lagrange multipliers and 10 is used as the initial penalty parameter.
  * [Semidefinite programming on Wikipedia](https://en.wikipedia.org/wiki/Semidefinite_programming)
  * [Semidefinite programs](#semidefinite-programs) (includes example usage of `PrimalDualSolver`)
 
+## Momentum DeltaBarDelta
+
+*An optimizer for [differentiable functions](#differentiable-functions).*
+
+A DeltaBarDelta variant that incorporates the following modifications:
```
rcurtin:
```suggestion
A [DeltaBarDelta](#deltabardelta) variant that incorporates the following modifications:
```
Could be nice to add the link here.

doc/optimizers.md
```diff
@@ -2396,6 +2466,67 @@ for all Lagrange multipliers and 10 is used as the initial penalty parameter.
...
+ - In the original DeltaBarDelta, the momentum term (delta_bar) is used
+   solely for sign comparison with the current gradient and does not
+   participate in the parameter update. In this modified variant, the
+   momentum term (velocity) is directly used to update the parameters.
```
rcurtin:
```suggestion
 - In the original DeltaBarDelta, the momentum term (`delta_bar`) is used
   solely for sign comparison with the current gradient and does not
   participate in the parameter update. In this modified variant, the
   momentum term (`velocity`) is directly used to update the parameters.
```

Just some ideas for clarity, let me know what you think.

doc/optimizers.md
rcurtin: Thanks, I appreciate the effort on that!

doc/optimizers.md
```diff
@@ -2396,6 +2466,67 @@ for all Lagrange multipliers and 10 is used as the initial penalty parameter.
...
+| `size_t` | **`maxIterations`** | Maximum number of iterations allowed (0 means no limit). | `100000` |
+| `double` | **`tolerance`**  | Maximum absolute tolerance to terminate algorithm. | `1e-5` |
+| `double` | **`kappa`** | Additive increase constant for step size. | `0.2` |
+| `double` | **`phi`** | Multiplicative decrease factor for step size. | `0.8` |
+| `double` | **`momentum`** | The momentum hyperparameter. | `0.5` |
+| `double` | **`minGain`** | Minimum allowed gain (scaling factor) for any parameter. | `1e-8` |
+| `bool` | **`resetPolicy`** | If true, parameters are reset before every `Optimize()` call. | `true` |
+
+Attributes of the optimizer may also be modified via the member methods
+`StepSize()`, `MaxIterations()`, `Tolerance()`, `Kappa()`, `Phi()`, `Momentum()`, `MinGain`, and `ResetPolicy()`.
```
rcurtin:
```suggestion
`StepSize()`, `MaxIterations()`, `Tolerance()`, `Kappa()`, `Phi()`, `Momentum()`, `MinGain()`, and `ResetPolicy()`.
```

tests/delta_bar_delta_test.cpp
rcurtin: Not really, just something reasonable that solves the problem.  There are a few cases where the parameters look very specific, but this is because there was a behavior change for how batch sizes and step sizes were handled together; but, you will see in those cases that there are #ifdefs surrounding the legacy behavior/parameters.

include/ensmallen_bits/delta_bar_delta/update_policies/momentum_delta_bar_delta_update.hpp
rcurtin: Sounds good, agreed, I guess you can do that with callbacks without too much trouble.

include/ensmallen_bits/gradient_descent/gradient_descent.hpp
rcurtin: Oh, sorry, maybe I missed it!  Wherever is fine.

include/ensmallen_bits/delta_bar_delta/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Update step for Gradient Descent.
+     *
+     * @param iterate Parameters that minimize the function.
+     * @param stepSize Step size to be used for the given iteration.
+     * @param delta The gradient matrix.
+     */
+    void Update(MatType& iterate,
+                const double stepSize,
```
rcurtin:
```suggestion
                const double /* stepSize */,
```
This should fix a -Wunused warning.

include/ensmallen_bits/delta_bar_delta/update_policies/momentum_delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,177 @@
...
+     * @param stepSize Step size to be used for the given iteration.
+     * @param gradient The gradient matrix.
+     */
+    void Update(MatType& iterate,
+                const double stepSize,
+                const GradType& gradient)
+    {
+      gains += (sign(gradient) != sign(velocity)) * kappa -
+          (sign(gradient) == sign(velocity)) * (1 - phi) % gains;
+      gains.clamp(minGain, arma::Datum<typename MatType::elem_type>::inf);
```
rcurtin:
```suggestion
      gains.clamp(minGain, arma::Datum<ElemType>::inf);
```
Just a simplification.

include/ensmallen_bits/delta_bar_delta/update_policies/momentum_delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+      velocity = momentum * velocity - (stepSize * gains) % gradient;
```
rcurtin:
```suggestion
      velocity = momentum * velocity - (ElemType(stepSize) * gains) % gradient;
```

For this to work with FP16, we actually have to cast the `stepSize` from `double` to whatever it needs to be.  This is because `double` cannot be implicitly cast to `std::float16_t` (or whatever `arma::fp16` is typedef'ed as).

I found this issue while testing the PR with FP16 enabled and Bandicoot available too (I just wanted to make sure that worked because our CI for ensmallen doesn't currently do that right).

include/ensmallen_bits/delta_bar_delta/update_policies/momentum_delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,177 @@
...
+     * @param iterate Parameters that minimize the function.
+     * @param stepSize Step size to be used for the given iteration.
+     * @param gradient The gradient matrix.
+     */
+    void Update(MatType& iterate,
+                const double stepSize,
+                const GradType& gradient)
+    {
+      gains += (sign(gradient) != sign(velocity)) * kappa -
+          (sign(gradient) == sign(velocity)) * (1 - phi) % gains;
```
rcurtin:
```suggestion
      gains += conv_to<MatType>::from(
          (sign(gradient) != sign(velocity)) * kappa -
          (sign(gradient) == sign(velocity)) * (1 - phi) % gains);
```
I know this seems redundant for most types, but for FP16, `umat % hmat` will yield `fmat` (because the `u32`/`u64` of a `umat` can't be represented by an FP16), so we have to convert back down to an FP16 matrix in that case.  For any other case, the type of the inner expression will already be `MatType` so the `conv_to` will do nothing.

include/ensmallen_bits/delta_bar_delta/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,198 @@
...
+                const GradType& delta)
+    {
+      const MatType signMatrix = sign(delta % deltaBar);
+
+      epsilon += (signMatrix == +1) * kappa -
+          (signMatrix == -1) * phi % epsilon;
+      epsilon.clamp(minStepSize,
+          arma::Datum<typename MatType::elem_type>::inf);
```
rcurtin:
```suggestion
      epsilon += conv_to<MatType>((signMatrix == +1) * kappa -
          (signMatrix == -1) * phi % epsilon);
      epsilon.clamp(minStepSize, arma::Datum<ElemType>::inf);
```
See the discussion in the other comment I left about why the `conv_to` is necessary here for the optimizer to work with FP16.

include/ensmallen_bits/delta_bar_delta/update_policies/delta_bar_delta_update.hpp
```diff
@@ -0,0 +1,198 @@
...
+        const size_t cols) :
+        parent(parent),
+        kappa(ElemType(parent.kappa)),
+        phi(ElemType(parent.phi)),
+        theta(ElemType(parent.theta)),
+        minStepSize(ElemType(parent.minStepSize))
+    {
+      deltaBar.zeros(rows, cols);
+      epsilon.set_size(rows, cols);
+      epsilon.fill(parent.InitialStepSize());
```
rcurtin:
```suggestion
      epsilon.fill(ElemType(parent.InitialStepSize()));
```
This is necessary for use of this optimizer with FP16 since a `double` cannot be implicitly converted to `std::float16_t` (or whatever type `arma::fp16` is).

tests/delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+    ENS_ALL_TEST_TYPES)
+{
+  DeltaBarDelta s(0.9, 50, 1e-9, 0.001, 0.2, 0.5);
+  FunctionTest<GDTestFunction, TestType>(s,
+      Tolerances<TestType>::LargeObj,
+      Tolerances<TestType>::LargeCoord);
+}
+
+TEMPLATE_TEST_CASE("DeltaBarDelta_RosenbrockFunction", "[DeltaBarDelta]",
+    ENS_ALL_TEST_TYPES)
```
rcurtin:
```suggestion
    ENS_ALL_CPU_TEST_TYPES)
```
For the Rosenbrock function, which accesses elements of its coordinates individually, GPU usage with Bandicoot is [super slow](https://coot.sourceforge.io/docs.html#arma_comparison).  So, using this macro will disable these tests for Bandicoot types (when Bandicoot is available).

tests/delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+      0.0001, 0.2, 0.5);
+  FunctionTest<RosenbrockFunction, TestType>(s,
+      10 * Tolerances<TestType>::LargeObj,
+      10 * Tolerances<TestType>::LargeCoord);
+}
+
+TEMPLATE_TEST_CASE("DeltaBarDelta_LogisticRegressionFunction",
+    "[DeltaBarDelta]", ENS_ALL_TEST_TYPES)
+{
+  DeltaBarDelta s(0.0008, 20, Tolerances<TestType>::Obj,
```
rcurtin:
```suggestion
  DeltaBarDelta s(0.0001, 500, Tolerances<TestType>::Obj,
```
I had to tune these parameters to get convergence for FP16.

tests/delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+  DeltaBarDelta s(0.9, 50, 1e-9, 0.001, 0.2, 0.5);
+  FunctionTest<GDTestFunction, TestType>(s,
+      Tolerances<TestType>::LargeObj,
+      Tolerances<TestType>::LargeCoord);
+}
+
+TEMPLATE_TEST_CASE("DeltaBarDelta_RosenbrockFunction", "[DeltaBarDelta]",
+    ENS_ALL_TEST_TYPES)
+{
+  DeltaBarDelta s(0.001, 0, Tolerances<TestType>::Obj / 100,
```
rcurtin:
```suggestion
  DeltaBarDelta s(0.001, 100000, Tolerances<TestType>::Obj / 100,
```
I missed this the first time, but it's not a great idea to potentially enter an infinite loop in the tests, so setting a maximum number of iterations is really helpful.

tests/momentum_delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "test_function_tools.hpp"
+#include "test_types.hpp"
+
+using namespace ens;
+using namespace ens::test;
+
+TEMPLATE_TEST_CASE("MomentumDeltaBarDelta_GDTestFunction",
+    "[MomentumDeltaBarDelta]", ENS_ALL_TEST_TYPES)
+{
+  MomentumDeltaBarDelta s(0.9, 50, 1e-9, 0.2, 0.8, 0.5);
```
rcurtin:
```suggestion
  MomentumDeltaBarDelta s(0.1, 1000, 1e-9, 0.2, 0.8, 0.5);
```
These changes were necessary to get convergence with FP16.

tests/momentum_delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+  FunctionTest<GDTestFunction, TestType>(s,
+      Tolerances<TestType>::LargeObj,
+      Tolerances<TestType>::LargeCoord);
+}
+
+TEMPLATE_TEST_CASE("MomentumDeltaBarDelta_RosenbrockFunction",
+    "[MomentumDeltaBarDelta]", ENS_ALL_TEST_TYPES)
```
rcurtin:
```suggestion
    "[MomentumDeltaBarDelta]", ENS_ALL_CPU_TEST_TYPES)
```
(Same thing as for the other Rosenbrock function test.)

tests/momentum_delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+{
+  MomentumDeltaBarDelta s(0.001, 0, Tolerances<TestType>::Obj / 100,
+    0.2, 0.8, 0.5);
```
rcurtin:
```suggestion
  MomentumDeltaBarDelta s(0.001, 100000, Tolerances<TestType>::Obj / 100, 0.2,
      0.8, 0.5);
```
(Same thing as for the other Rosenbrock function test.)

tests/momentum_delta_bar_delta_test.cpp
```diff
@@ -0,0 +1,47 @@
...
+  FunctionTest<RosenbrockFunction, TestType>(s,
+      10 * Tolerances<TestType>::LargeObj,
+      10 * Tolerances<TestType>::LargeCoord);
+}
+
+TEMPLATE_TEST_CASE("MomentumDeltaBarDelta_LogisticRegressionFunction",
+    "[MomentumDeltaBarDelta]", ENS_ALL_TEST_TYPES)
+{
+  MomentumDeltaBarDelta s(0.0004, 20, Tolerances<TestType>::Obj,
```
rcurtin:
```suggestion
  MomentumDeltaBarDelta s(0.00005, 2000, Tolerances<TestType>::Obj,
```
These parameters seemed to be necessary to get everything to converge reliably with FP16 as the element type.

rcurtin (commented):
Hey @ranjodhsingh1729, I realized before I merged this that our CI setup had not adequately tested it with [FP16 types](https://arma.sourceforge.net/docs.html#fp16_type) or with Bandicoot.  You can do this by installing a version of Armadillo that has FP16 support locally, and also installing Bandicoot locally, and then configuring CMake with

```
-DARMADILLO_INCLUDE_DIR=/path/to/arma/tmp/include/ -DARMADILLO_LIBRARY=/path/to/arma/libarmadillo.so -DBANDICOOT_INCLUDE_DIR=/path/to/coot/build/tmp/include/ -DBANDICOOT_LIBRARY=/path/to/coot/build/libbandicoot.so -DCMAKE_CXX_FLAGS="-DARMA_FORCE_USE_FP16" -DCMAKE_CXX_STANDARD=23
```

which is a bit of a mouthful...

Anyway, I had it set up locally so I built it all and ran the tests 1000 times with different random seeds, and debugged a few compilation failures with FP16 elements and a couple convergence issues (FP16 has more trouble converging because of its much more limited precision).  I added everything here as suggestions---you can accept them, or let me know if something's wrong with them, otherwise I'll just wait a couple days then merge the suggestions, then merge the PR.  Overall I am really happy that there were only a couple of changes to be made to make the optimizer work either in reduced precision or on the GPU---it means we are doing a good job of implementing the algorithms in an abstract enough way that they can run wherever!  (Now, it is not true of all optimizers in ensmallen, but for DeltaBarDelta and gradient descent in general, it's actually pretty well-suited to the GPU so long as the size of `coordinates` is sufficiently large to get any speedup out of the GPU for the matrix multiplications and other operations... and also so long as the `FunctionType` being optimized runs well enough on the GPU.)

ranjodhsingh1729:
Hi @rcurtin, Thanks for the suggestions (and their detailed descriptions). :)
It kind of slipped my mind that i needed to test with fp16 and bandicoot too. :(

I tried to tests the changes locally, but either something is wrong with my setup or maybe something is missing in suggestions. I am getting compilation errors due to type mismatches in the update rule. So if you have working changes locally, could you please just push them directly to this branch? If you have the time that is. Otherwise, I'll try to dive deeper tommorrow and see if something is wrong on my side or something needs fixing.

Thanks again!

#### This is where i got so far:

coot was raising error while trying to perform element-wise multiplication or subtraction between unsigned long and float matrices. After some reading of bandicoot docs and not finding anyting, I decided to just slap two more `conv_to` leading to this not so good code:

```cpp
      const MatType signMatrix = sign(delta % deltaBar);

      const MatType increment = conv_to<MatType>::from(
          (signMatrix == +1));
      const MatType decrement = conv_to<MatType>::from(
          (signMatrix == -1));

      epsilon += conv_to<MatType>::from(increment * kappa 
          - decrement * phi % epsilon);
      epsilon.clamp(minStepSize, arma::Datum<ElemType>::inf);
```

After this, I got another in `momentum_delta_bar_delta_update.hpp`. Which i can't seem to fix yet ( i am kind of blaming bandicoot for this one :)

rcurtin:
> i am kind of blaming bandicoot for this one :)

Actually you are right to do that!  What I failed to mention in the comment I posted was that the DeltaBarDelta rule actually exposed a bug in Bandicoot that I fixed here: https://gitlab.com/bandicoot-lib/bandicoot-code/-/merge_requests/192  Sorry that I forgot about that part, but based on the output you gave it looks like you debugged it or worked around it all the way regardless.

Anyway, if you want to test with "working" Bandicoot, you either have to use that particular branch or wait a day or so until I release the next version of Bandicoot with that fix included.

It's my hope to make the CI for ensmallen a little bit better so these things are automatically caught by CI instead of making developers manually check out and test with FP16 and Bandicoot, etc., (it takes a long time and is tedious!), but, not there yet.

ranjodhsingh1729:
Thanks for letting me know.
I just had a feeling but didn't want to dive in bandicoot's codebase and confirm :).

rcurtin:
Thanks again @ranjodhsingh1729!  Happy to get this merged :rocket:   I'll get a new release in progress now.

PR #442, opened 2025-10-19
Add option to print loss every `printInterval` epochs in `PrintLoss` callback.

ranjodhsingh1729:
### Description

A small addition to the `PrintLoss` callback which allows to print loss value every `printInterval` epochs. The change is backwards compatible (`printInterval` has a default value of 1).

### DIFF
--- a/include/ensmallen_bits/callbacks/print_loss.hpp
+++ b/include/ensmallen_bits/callbacks/print_loss.hpp
@@ -12,6 +12,8 @@
 #ifndef ENSMALLEN_CALLBACKS_PRINT_LOSS_HPP
 #define ENSMALLEN_CALLBACKS_PRINT_LOSS_HPP
 
+#include <stdexcept>
+
 namespace ens {
 
 /**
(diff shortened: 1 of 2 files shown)


include/ensmallen_bits/callbacks/print_loss.hpp
```diff
@@ -41,16 +49,20 @@ class PrintLoss
   bool EndEpoch(OptimizerType& /* optimizer */,
                 FunctionType& /* function */,
                 const MatType& /* coordinates */,
-                const size_t /* epoch */,
+                const size_t epoch,
                 const double objective)
   {
-    output << objective << std::endl;
+    // epochs are 1-indexed.
```
rcurtin, line 55: Are they?  I thought they started at 0.  In either case, it doesn't really make a big difference: if they are 0-indexed they will print the first iteration too.

rcurtin (commented):
The implementation looks correct enough to me but I have to ask, is there really any need for this?  `PrintLoss` is meant to be a really primitive callback.  I haven't ever seen anyone ask for this particular feature and I'm not actually sure the situation where someone would want to use it.

ranjodhsingh1729:
I figured it might help, since with lots of iterations the terminal can get flooded and we lose all previous output. If we don't print anything, though, there's no sense of progress. I also wanted to include the epoch number in the output, but didn't want to change the current behavior.

Happy to leave it as is if you think keeping `PrintLoss` simple makes more sense.

PR #444, opened 2025-10-19
Add calls to `BeginEpoch` and `EndEpoch` in `GradientDescent`.

ranjodhsingh1729:
### Description

The current `GradientDescent` doesn't work with neither `ProgressBar` nor `PrintLoss`.
It doesn't work with `ProgressBar` as `ProgressBar` currently requires `OptimizerType` to have `NumFunctions` and `BatchSize`.
It doesn't work with `PrintLoss` as `PrintLoss` only prints loss when `EndEpoch` is called which `GradientDescent` doesn't call.

This PR adds calls to both `BeginEpoch` and `EndEpoch` in `GradientDescent` so that it can support `PrintLoss`.

### DIFF
--- a/include/ensmallen_bits/gradient_descent/gradient_descent_impl.hpp
+++ b/include/ensmallen_bits/gradient_descent/gradient_descent_impl.hpp
@@ -69,10 +69,21 @@ GradientDescent::Optimize(FunctionType& function,
   Callback::BeginOptimization(*this, f, iterate, callbacks...);
   for (size_t i = 1; i != maxIterations && !terminate; ++i)
   {
+    terminate |= Callback::BeginEpoch(*this, f, iterate, i,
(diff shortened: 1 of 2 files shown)


rcurtin:
The `BeginEpoch()` and `EndEpoch()` callback functions are specific to separable functions: https://www.ensmallen.org/docs.html#beginepoch

So really the fix here would be to have `PrintLoss` and `ProgressBar` detect whether the function is separable, and if it is not, then print the loss/progress step probably in `StepTaken()` or another callback function.

ranjodhsingh1729:
> The BeginEpoch() and EndEpoch() callback functions are specific to separable functions: https://www.ensmallen.org/docs.html#beginepoch

My Bad!
I'll close this and open one for `PrintLoss` seperately.

PR #446, opened 2025-11-12
fix missing return in aug_lagrangian

arminsch:
This fixes a missing `return` in one of the implementations of `Optimize` in `aug_lagrangian.hpp`.


rcurtin (approved):
Thank you for the fix!

eddelbuettel (approved):

zoq (approved):

arminsch:
Just a bit more than 1h for pulling in a PR (even if it is a tiny one). Impressive, you are really quick.

rcurtin:
It's easy to quickly review when the fix is so simple :smile:  we might be less fast if you, say, contribute an entire new optimizer or something...  Thanks again!

PR #448, opened 2025-12-13
Revert "Release version 3.11.0: "Sunny Day" (#447)"

rcurtin:
This reverts commit db5cef982396dcfea945ecab68f3b588b82be77c.  The reason is that when we merge releases (as listed in the auto-generated description) they need to be rebased to produce two commits: the first is the one that gets tagged for the release, the second is the new `HEAD` with the new empty `HISTORY.md` block.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,8 +109,8 @@ $ cd ensmallen
 
 # - or -
 
-$ wget http://ensmallen.org/files/ensmallen-3.11.0.tar.gz
-$ tar -xvzpf ensmallen-3.11.0.tar.gz
+$ wget http://ensmallen.org/files/ensmallen-3.10.0.tar.gz
+$ tar -xvzpf ensmallen-3.10.0.tar.gz
 $ cd ensmallen-latest
 ```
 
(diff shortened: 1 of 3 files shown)


conradsnicta:
@rcurtin Seems to be a rather convoluted release process :)  

(The ensmallen webpage and the Releases section also need to be updated.  I started a PR for this on the ensmallen.org repo, then realised that there was no new release .tar.gz yet)

conradsnicta (approved):

rcurtin:
> (The ensmallen webpage and the Releases section also need to be updated. I started a PR for this on the ensmallen.org repo, then realised that there was no new release .tar.gz yet)

Yes, because the entire purpose of the 'convoluted' release process was to automate it.  The scripts for this automation are all in `scripts/`, including the post-release website update script.  The intention is to run that inside of a Github action or bot equivalent so that releases really are just "run a script, it does everything else".  But, for this to work, it depends on the release merge being a rebase instead of a squash (or really anything other than a squash will do), otherwise you have no commit that you can tag to be the new release.  I'll try again once we merge this revert and then I will have a commit to use as the tag and all the scripts will work fine---and the webpage will be automatically updated via scripts.

Issue #452, opened 2026-03-13
[Discussion] More OpenMP support


Segfault1602:
Hi,

I am wondering if there has been discussion regarding better multithreading/openmp support in ensmallen. It seems to me that certain algorithms (PSO, CNE, and DE come to mind) could benefit greatly from parallel processing. Such a change would potentially break a lot of the callbacks shipped with ensmallen so I'm wondering if it's something people have already thought about and if that is something people are generally interested in.

Thanks,
-Alex

rcurtin:
I agree that it would be really nice for some algorithms---especially the ones that you specified which are in places embarrassingly parallel.  I'd be happy to help an effort to implement that support but I think we'd need to think about the following:

 * How could a user disable or control the multithreading so that, e.g., if their `Evaluate()` implementation is already using parallelism, they could disable the optimizer-level parallelism?

 * How could we ensure that callbacks (when they were called) were still thread-safe, or, e.g., only called from the main thread?

 * How can we ensure that the code will still compile and run just fine even if the user doesn't have OpenMP enabled?

 * Are there places where parallelism would actually affect the results (moreso than just changing floating-point operation ordering)?

Definitely those questions are not impossible to answer, I just think we'd need to keep them in mind as we worked out a solution.

Segfault1602:
For the build related questions:

I was hoping this part would be relatively easy given that ParallelSGD already uses OpenMP but it appears the build is broken when enabling `ENS_USE_OPENMP`. It's an easy fix but then the ParallelSGD tests are failing so that would probably need to be fixed first.

I do think that reusing `ENS_USE_OPENMP` could be a big breaking change so maybe we need to use a new `#define` flag? 

>How could we ensure that callbacks (when they were called) were still thread-safe, or, e.g., only called from the main thread?

I think that it would be reasonable to expect the user's code to be thread-safe when parallelization is enabled, I'm not sure I can imagine a scenario where I expect my `Evaluate()` calls to be called on multiple threads but then need my `Callback::Evaluate()` call to be sequential and or single threaded. Things like `ens::report()` can be made thread-safe relatively easily but for more complex callback that changes optimizer state, I don't know if there is a solution that will work for everyone. Taking `CNE` as an example, it looks like the optimizer state can't be changed in the middle of a generation so that is good. It also looks like it doesn't support early termination in the middle of a generation, so that also make things easier.

```cpp 
// Calculating fitness values of all candidates.
// cne_impl.hpp:122
[...]

PR #454, opened 2026-04-14
Fix ParallelSGD test step sizes and enable OpenMP for tests when found

rcurtin:
This is a follow-up for #452.  When the tests were refactored for Bandicoot over last summer, one thing that got overlooked was the ParallelSGD tests, since they require OpenMP to be enabled to run... and our CMake configuration did not actually do that even when it found OpenMP!  So, two fixes:

1. Actually enable OpenMP if we find it...

2. and update the step sizes for each of the ParallelSGD tests since they were way too large and caused divergence.

I ran the tests for 1000 iterations locally with no problems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,7 @@ endif()
 
 # Find OpenMP and link it.
 if(USE_OPENMP)
+  find_package(OpenMP)
   if(NOT TARGET OpenMP::OpenMP_CXX)
     find_package(Threads REQUIRED)
     add_library(OpenMP::OpenMP_CXX IMPORTED INTERFACE)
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -61,6 +62,8 @@ if(USE_OPENMP)
     set_property(TARGET OpenMP::OpenMP_CXX
         PROPERTY INTERFACE_LINK_LIBRARIES ${OpenMP_CXX_FLAGS} Threads::Threads)
   endif()
+
+  add_definitions(-DENS_USE_OPENMP)
```
Segfault1602: Should it be `target_compile_definition` instead?

zoq (approved):

CMakeLists.txt
rcurtin: Thanks---my CMake is often pretty primitive.  Fixed in fd4362d.

PR #456, opened 2026-05-01
Allow multiple trials for Lookahead logistic regression test

rcurtin:
I saw in another PR that this test randomly failed and was a little bit disappointed because of the time I spent last summer tuning all the test suites so they had no test failures in 1000 runs.  So, I ran the test locally another 1000 times and saw one failure... an easy fix is just to allow multiple trials for the logistic regression test, so I did that, and in ~4k trials I saw no failures.


zoq (approved):
Easy review.

eddelbuettel (approved):
Nice trick to fall victim to an unlucky (and rare) RNG draw

PR #458, opened 2026-07-10
Initialize the gradient constraint matrix instead of expecting the user to

### LINKED ISSUE #457: Augmented Lagrangian optimiser seems to fail?
Using 
ensmallen:                    3.11.0 (Sunny Day)
armadillo:                    15.2.4 (Medium Roast Deluxe)
on Linux Mint.

I am having a problem with the augmented Lagrangian optimiser failing in what seems like a very naive way. The attached MWE fails on first call to the user-defined GradientConstraint function. In particular, the gradient matrix passed in by reference is not dimensioned (i.e. 0 x0) resulting in a segfault when I try assign any element.

Looking at the stack trace, I can see that GradientConstraint is being called by AugLagrangianFunction, which passes a reference to the local variable constraintGradient defined on line 76 of aug_lagrangian_function_impl.hpp. But this is not dimensioned (presumably it's 0 x 0 by default.) Surely the user is not expected to define the dimensions?

When I manually alter the definition of constraintGradient to allocate the same dimensions as the gradient argument passed into AugLagrangianFunction then the thing works (albeit it gives the wrong answer, but this may be an error with my coding of the gradients. It is an MWE!)

Alternatively, because this seems so basic, am I misusing the library?

P.

[ensmallen-augLag-mwe.cpp](https://github.com/user-attachments/files/29897047/ensmallen-augLag-mwe.cpp)

rcurtin:
This fixes #457.  The bug report there is that `AugLagrangian` is not setting the size of the gradient matrix before calling the function's `GradientConstraint()` method.

It's a pretty simple fix, and it's now tested because I modified the AugLagrangian test functions to ensure that some of them expect the gradient size to already be set.

### DIFF
--- a/include/ensmallen_bits/aug_lagrangian/aug_lagrangian_function_impl.hpp
+++ b/include/ensmallen_bits/aug_lagrangian/aug_lagrangian_function_impl.hpp
@@ -73,7 +73,8 @@ void AugLagrangianFunction<LagrangianFunction, VecType>::Gradient(
   gradient.zeros();
   function.Gradient(coordinates, gradient);
 
-  GradType constraintGradient; // Temporary for constraint gradients.
+  // Temporary for constraint gradients.
+  GradType constraintGradient(gradient.n_rows, gradient.n_cols);
   for (size_t i = 0; i < function.NumConstraints(); i++)
   {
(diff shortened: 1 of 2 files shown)


eddelbuettel (approved):
LGTM

PR #461, opened 2026-08-04
Adapt mlpack CI workflow to ensmallen

rcurtin:
I used mlpack's `ci.yml` as a starting point to revamp ensmallen CI to only use Github Actions for CI builds.  Here I've set up two Linux and macOS builds (one for debug mode, one for release mode) and a Windows job.

I *think* this unfortunately will not build until it's merged to master, though, so I may need to merge this PR initially and then refactor/cleanup in a follow-up PR.  *(Edit: actually, I was wrong!  Seems like it's building fine.)*


.github/workflows/ci.yml
```diff
@@ -0,0 +1,255 @@
...
+    name: '${{ matrix.config.os }} (${{ matrix.config.name }})'
+    if: ${{ github.repository == 'mlpack/ensmallen' }}
+    runs-on: ${{ matrix.config.runner }}
+
+    steps:
+      - uses: actions/checkout@v6
+
+      # Set up ccache.
+      - name: Get ccache
+        uses: hendrikmuhs/ccache-action@v1.2
```
eddelbuettel: I recommend v1.2.23 to avoid the 'node 20 is going away' nag on each use.  Same below.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,255 @@
...
+      # Install base build dependencies via apt.
+      - name: Install apt build dependencies (Linux)
+        if: matrix.config.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes --allow-unauthenticated libopenblas-dev
+
+      # Install build dependencies.
+      - name: Install build dependencies (Linux)
+        if: matrix.config.os == 'Linux'
```
eddelbuettel: Merge with preceding block?  They both condition on leenucks.

.github/workflows/ci.yml
rcurtin: Thanks, done in 24b27145.

.github/workflows/ci.yml
rcurtin: Good point!  There was an extra block between these two for the mlpack CI that wasn't necessary here, and then I just overlooked the obvious simplification.  Done in c2c14e38.

.github/workflows/ci.yml
```diff
@@ -49,20 +49,14 @@ jobs:
             os: 'macOS',
             name: 'Debug',
             cmakeVars: '-DCMAKE_BUILD_TYPE=Debug',
-            # Up to 3 cores are supported but the amount of ram provided
-            # leads to severe swapping and massive slowdowns, using 1 is
-            # the best compromise between cores and ram given our C++ load
-            cores: 1
+            cores: 3
```
eddelbuettel, line 48: Yay!  Not the full mlpack workload so not the same need for reduction in force.  Nice catch!

rcurtin: Yeah, I was worried this might still swap, but looking at the build logs, it seems like it's fine.

rcurtin:
Ok, I think this one is ready to go.  I think I need to remove the AppVeyor hook separately from this PR, though.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,234 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+      - main
+  pull_request:
+    branches:
+      - master
+      - main
```
eddelbuettel: Really stupid nit:  If you use more compact `[square, bracket]`  notation on line 16. why not also on 9+10 and 13+14?

eddelbuettel (approved):
👍

.github/workflows/ci.yml
rcurtin: Because it's what was there before :smile: fixed in 38ca4f3.

rcurtin:
I deleted the webhook for AppVeyor, so that shouldn't trigger on other PRs.

